PR #3, opened 2019-06-13
fix: custom template verifies signature against entire style object

YSNIS:


src/services/banner/customTemplate.js
```diff
@@ -113,10 +128,23 @@ function fetcher(url) {
     });
 }
 
-function getCustomTemplate(sign, account, source) {
+// Removes sign, flattened, and possibly ratio if undefined from the style object.  Returns everything else.
+function trimStyles(obj) {
+    const styleObject = { ...obj };
+    delete styleObject.sign;
```
gillycheesesteak: I think it makes sense to move this outside the style object

src/services/banner/customTemplate.js
```diff
@@ -113,10 +128,23 @@ function fetcher(url) {
...
+    delete styleObject._flattened;
+    if (styleObject.ratio === undefined) {
```
gillycheesesteak: We can probably do this generically for any style field that is `undefined`

src/services/banner/customTemplate.js
```diff
@@ -113,10 +121,32 @@ function fetcher(url) {
     });
 }
 
-function getCustomTemplate(sign, account, source) {
+// Removes sign, flattened, and possibly ratio if undefined from the style object.  Returns everything else.
+function trimStyles(obj) {
+    const styleObject = objectEntries(obj).reduce((accum, kvp) => {
```
gillycheesesteak: Let's destructure kvp here and simplify the code below

PR #11, opened 2019-07-03
fix: custom validation refactor

YSNIS:


demo/index.html
```diff
@@ -4,20 +4,21 @@
...
-            //         }
-            //     })
-            //     .render('.messages');
+            paypal
+                .Messages({
+                    account: 'TMW5TEF2KZESG',
+                    amount: '500',
+                    style: {
+                        layout: 'custom',
+                        markup: 'https://localhost.paypal.com:8080/custom.html'
```
Seavenly: We should probably refrain from committing changes to the demo pages, especially with for non-general use-cases.

src/services/banner/customTemplate.js
```diff
@@ -121,33 +21,10 @@ function fetcher(url) {
...
-    return markupProm.then(markup =>
-        validateSign(sign, account, { ...styleObject, markup }).then(validated => {
-            if (!validated) {
-                logCustomValidationError(account, markup);
-            }
-
-            return validated ? markup : '';
-        })
-    );
+    return markupProm.then(markup => markup);
```
Seavenly: It looks like we are accepting any `style.markup` value. We should probably ensure that a URL is used from `paypalobjects` or give a blank message.

demo/custom.html
```diff
@@ -0,0 +1,280 @@
+<!--
+    {"foo":"bar","style":{"ratio":"2x1"}}
+-->
+<style>
```
Seavenly: Do we want an example of this in our repo?

src/models/Banner/index.js
```diff
@@ -56,16 +56,13 @@ const Banner = {
...
         function render(totalOptions) {
             const options = validateOptions(totalOptions);
-            const renderProm = getBannerMarkup(options) // Promise<Object(markup)>
+            const renderProm = getBannerMarkup(options) // Promise<Object({ markup, options })>
                 .then(
-                    pipe(
-                        partial(objectAssign, { options }), // Object(markup, options)
-                        insertMarkup // Promise<Object(meta)>
-                    )
+                    insertMarkup // Promise<Object(meta)>
```
Seavenly:
```suggestion
                    insertMarkup // Promise<Object(meta, options)>
```

src/models/Banner/index.js
```diff
@@ -56,16 +56,13 @@ const Banner = {
 
         function render(totalOptions) {
             const options = validateOptions(totalOptions);
-            const renderProm = getBannerMarkup(options) // Promise<Object(markup)>
+            const renderProm = getBannerMarkup(options) // Promise<Object({ markup, options })>
```
Seavenly:
```suggestion
            const renderProm = getBannerMarkup(options) // Promise<Object(markup, options)>
```

src/services/banner/index.js
```diff
@@ -68,30 +71,42 @@ function fetcher({ account, amount, countryCode }) {
...
-                // eslint-disable-next-line no-param-reassign
-                data.markup.template = template;
-            }
+    return ZalgoPromise.all([memoFetcher(options), getCustomTemplate(options.style)]).then(([data, template]) => {
+        if (typeof data.markup === 'object') {
+            data.markup.template = template;
 
-            return data;
+            const bannerOptions = getBannerOptions(template);
+            options = objectMerge(options, { ...bannerOptions });
```
Seavenly: Consider creating a new variable instead of reassigning the options reference.

Seavenly (changes requested):

src/services/banner/index.js
```diff
@@ -68,30 +71,42 @@ function fetcher({ account, amount, countryCode }) {
...
 
+/**
+ * Extract annotations from markup into a single options object
+ * @param {string} markup String of banner HTML markup
+ * @returns {Object} Banner annotations
+ */
+function getBannerOptions(markup) {
+    const annotationsString = markup.match(/^<!--([\s\S]+?)-->/);
+    if (annotationsString) {
+        return JSON.parse(annotationsString[1]);
```
gillycheesesteak: Would it be valuable to wrap this in a `try { } catch { }` in case the contents of the comment are not valid JSON?

src/services/banner/index.js
gillycheesesteak: Can `bannerOptions` be passed as-is as the second argument here?

src/services/banner/customTemplate.js
gillycheesesteak: This can just be `return markupProm`, or could just return directly without declaring `markupProm`

demo/index.html
YSNIS: Yeah I'm gonna revert this back.  For now though I wanted Ryan to have an easily accessible demo page to test the story.

src/services/banner/customTemplate.js
YSNIS: I changed the regex to match htt[s://www.paypalobjects.com

demo/custom.html
YSNIS: I think this can get reverted back along with the demo page.

src/models/Banner/index.js
YSNIS: Done.

src/models/Banner/index.js
YSNIS: Done.

src/services/banner/index.js
YSNIS: Done.

src/services/banner/index.js
YSNIS: Done.

src/services/banner/index.js
YSNIS: Done.

src/services/banner/customTemplate.js
YSNIS: Done.

src/services/banner/index.js
```diff
@@ -87,7 +87,12 @@ function fetcher(options) {
 function getBannerOptions(markup) {
     const annotationsString = markup.match(/^<!--([\s\S]+?)-->/);
     if (annotationsString) {
-        return JSON.parse(annotationsString[1]);
+        try {
+            const bannerOptions = JSON.parse(annotationsString[1]);
+            return bannerOptions;
```
Seavenly: Can't we just `return JSON.parse(annotationsString[1])`?

src/services/banner/customTemplate.js
```diff
@@ -22,9 +22,12 @@ function fetcher(url) {
 }
 
 function getCustomTemplate(styles) {
+    let markupRegex = /https:\/\/www\.paypalobjects\.com/;
```
Seavenly: Is it feasible to use `String.prototype.startsWith()` instead of a regex for this scenario? And for `__DEMO__` we can just accept anything since it's for internal use only?

Seavenly: Also, have we considered how this will work with the customization app since everything is now some form of URL?

src/services/banner/index.js
YSNIS: Yup - remnant of some previous code.

Seavenly (approved):

gillycheesesteak (approved):

PR #13, opened 2019-07-11
refactor: ps markup support

YSNIS:
Added personalization studio mock data to proxy imadserv and formatting it in markup fetch.


src/messages/services/banner/index.js
```diff
@@ -73,8 +91,13 @@ function fetcher(options) {
             });
             document.head.removeChild(script);
             delete window.__PP[callbackName];
+
+            const parsedMarkup = JSON.parse(markup.replace(/<\/?div>/g, ''));
             try {
-                resolve({ markup: JSON.parse(markup.replace(/<\/?div>/g, '')), options });
+                resolve({
+                    markup: parsedMarkup.touchpoint_messages ? mutateMarkup(parsedMarkup) : parsedMarkup,
```
gillycheesesteak: Let's add a comment here explaining the difference between these two options (ex: `v1 vs v2`)

gillycheesesteak (approved):

PR #14, opened 2019-07-15
refactor: log events grouped per message

roja079:


src/messages/models/Banner/index.js
```diff
@@ -54,49 +56,60 @@ const Banner = {
...
+            const returnVal = fn(args);
+            logger.info(name);
+            return returnVal;
+        });
+
         function render(totalOptions) {
-            const options = validateOptions(totalOptions);
-            const renderProm = getBannerMarkup(options) // Promise<Object( markup, options )>
+            logger.info(EVENTS.RENDER_START);
+            return pipe(
```
gillycheesesteak, line 64: Does it make sense to start with an object with `{ options, logger }` from the start here, instead of passing in the logger at every step?

src/messages/models/Banner/index.js
```diff
@@ -54,49 +56,60 @@ const Banner = {
...
+                passThrough(updateOptions), // Object()
+                assignToField('options'), // Object(options)
+                partial(objectAssign, { logger }), // Object(options, logger)
+                getBannerMarkup // Promise<Object(markup, options)>
+            )(totalOptions)
                 .then(
-                    insertMarkup // Promise<Object(meta, options)>
+                    pipe(
+                        partial(objectAssign, { logger }), // Promise<Object(markup, options, logger)>
+                        asyncAssign(logAfter(insertMarkup, EVENTS.INSERT)) // Promise<Object(meta)>
```
gillycheesesteak: Should this log happen after the insert is done (when the promise resolves)?

src/messages/controllers/render.js
```diff
@@ -43,46 +46,40 @@ export default function render(options, selector = '[data-pp-message]') {
...
-    logger.info(EVENTS.STARTING_MESSAGE_RENDER, {
-        url: window.location.href,
-        selector: selectorType
-    });
-
-    const updateFns = containers.map(container => {
-        const totalOptions = objectMerge(options, getInlineOptions(container));
+    return ZalgoPromise.all(
+        containers.map(container => {
+            const logger = (loggers.has(container)
```
gillycheesesteak: Does it make sense to create the `logger` instance inside `Banner.init` instead of creating it here and passing it in?

src/messages/models/Modal/index.js
```diff
@@ -6,20 +6,25 @@ import { ZalgoPromise } from 'zalgo-promise';
...
+    wrapper.setAttribute('data-pp-id', globalState.nextId);
+
     const [iframe, { insertMarkup }] = createContainer('iframe');
     const [parentOpen, parentClose] = initParent();
     const { track, clickUrl } = options;
     const [state, setState] = createState({
         status: 'CLOSED'
     });
+    const logger = Logger.create(globalState.nextId, '__internal__', 'Modal');
+    setGlobalState({ nextId: (globalState.nextId += 1) });
```
gillycheesesteak, line 32: Should we increment this id outside of the modal creation? Thinking hypothetically, if for example in the future we have a banner that does not create a modal?

src/messages/models/Banner/index.js
Seavenly: That would require some of these functions to be aware that they need to return the logger along with whatever other data they are actually creating/processing. It seems like a separate concern. Ideally I think we should have another helper utility that ensures data passed back from a function is appended to whatever was passed in.

src/messages/models/Banner/index.js
Seavenly: I think `logAfter` should be changed to `logBefore` actually to denote when the process starts. We'll know when it resolves when the next part of the pipeline then begins and logs it's start. Then any error messages that occur will appear after event name.

src/messages/controllers/render.js
Seavenly: That's a good point. I'll look into that.

src/messages/models/Modal/index.js
Seavenly: As we discussed, this is creating a new logger with an ID specific to the modal.

src/messages/services/logger/index.js
```diff
@@ -1,114 +1,103 @@
...
+            },
+            info(event, data = {}) {
+                setState({ logs: [...state.logs, { event, ...data }] });
+            },
+            error(data) {
+                logger.info(EVENTS.ERROR, data);
+            },
+            track: sendBeacon,
+            warn(...messages) {
+                console.warn('[Messaging.js]', ...messages);
```
gillycheesesteak: Are we intentionally using `[Messaging.js]` here instead of `[PayPal Messages]`?

src/messages/services/logger/index.js
```diff
@@ -1,114 +1,103 @@
...
 
-        return (immediate
-            ? ZalgoPromise.resolve()
-            : ZalgoPromise.all(pendingEvents).then(() => {
-                  pendingEvents.length = 0;
-              })
-        ).then(() => {
-            if (logs.length === 0) return;
+        function flush() {
+            if (state.count > FLUSH_MAX) return;
```
gillycheesesteak, line 51: Since we are intentionally flushing twice for banners/modals, do we want to increase the `FLUSH_MAX`?

src/messages/models/Modal/index.js
```diff
@@ -238,6 +243,8 @@ function createModal(options) {
     }
 
     function prepModal(ignoreCache = false) {
+        logger.start();
```
gillycheesesteak: `account` and other start event data are not being sent for the Modal log call

src/messages/services/logger/index.js
Seavenly: No. Probably just a bad copy/paste. Good catch.

src/messages/models/Modal/index.js
Seavenly: Yea. We need to at least send in the account here.

src/messages/services/logger/index.js
Seavenly: As we discussed, messages and modals each have separate instances of a logger.

gillycheesesteak (approved):

PR #41, opened 2019-08-21
refactor: update banner content for personalization studio

gillycheesesteak:


Seavenly:
Changes being pulled into #48

PR #48, opened 2019-08-27
feat: DE banners

gillycheesesteak:


src/interface/messages.js
```diff
@@ -18,7 +18,10 @@ export function setup() {
         // eslint-disable-next-line compat/compat
         const script = document.currentScript || document.querySelector('script[src$="messaging.js"]');
         if (script) {
-            Messages.setGlobalConfig(getInlineOptions(script));
+            // Default to en_US for non-SDK
+            Messages.setGlobalConfig({
+                ...getInlineOptions(script)
```
Seavenly: Is this change necessary?

src/messages/models/Banner/index.js
```diff
@@ -150,6 +150,10 @@ export default {
         }
 
         banner.renderProm = banner.renderProm.then(logger.end).catch(err => {
+            if (__LOCAL__) {
+                console.error(err);
```
Seavenly, line 180: Can this be removed?

src/messages/models/Banner/types.js
```diff
@@ -0,0 +1,8 @@
+export default {
```
Seavenly: Since we've extracted the types into their own module, should we also extract the `validateType()` function from `validateOptions.js` into the same location as more of a types helper module?

src/messages/models/Container/setSize.js
```diff
@@ -205,16 +205,13 @@ export default curry((container, { wrapper, options, logger, meta }) => {
         };
 
         if (parentContainerWidth < minBannerContentWidth && layout !== 'custom') {
-            if (
-                objectGet(options, 'style.logo.position') === 'top' &&
-                objectGet(options, 'style.logo.type') === 'primary'
-            ) {
+            if (container.getAttribute('data-pp-message-overflow') === 'fallback') {
```
Seavenly: Did we ever land anywhere with having a default set of options to check against based on locale to prevent unnecessary rerender attempts if the smallest style combination has already been selected?

src/messages/models/Container/stats.js
```diff
@@ -132,7 +132,7 @@ export default curry((container, { options: { amount }, events, track }) => {
     checkAdblock().then(detected => {
         payload.adblock = detected;
         payload.blocked = isHidden(container);
-        track(payload, container.hasAttribute('data-pp-message-hidden'));
+        track(payload, container.getAttribute('data-pp-message-hidden') === 'hidden');
```
Seavenly: Should this be `data-pp-message-overflow`?

src/messages/models/Template/index.js
```diff
@@ -364,34 +366,52 @@ function createTemplateNode(options, markup) {
     }
 
     const classNamePrefix = 'message';
+    const localeClass = `locale--${__MESSAGES__.__LOCALE__}`;
```
Seavenly: Should we include this constant as part of the `locale` module?

src/messages/models/Template/index.js
```diff
@@ -364,34 +366,52 @@ function createTemplateNode(options, markup) {
...
     ].map(getTemplateElement);
 
+    messagingContainer.classList.add(localeClass);
+
     appendText(headline, toMarkup('headline', mutationRules.headline));
     appendText(subHeadline, toMarkup('subHeadline', mutationRules.subHeadline));
     prependText(disclaimer, toMarkup('disclaimer', mutationRules.disclaimer));
-    appendImage(logoContainer, mutationRules.logo, 'PayPal Credit logo');
+    if (Array.isArray(mutationRules.logo)) {
+        mutationRules.logo.forEach(logo => appendImage(logoContainer, logo, 'PayPal Credit logo'));
```
Seavenly: Is it possible to have `appendImage()` handle the array check since both `appendText()` and `prependText()` handle that same logic internally.

src/messages/models/Template/styles/index.js
```diff
@@ -3,9 +3,15 @@ import flex from './flex';
...
-    'layout:flex': flex,
-    'layout:legacy': legacyHtml,
-    'layout:custom': custom
+    'layout:flex': flex
 };
+
+// Legacy and custom styles are only needed for US
+if (__MESSAGES__.__LOCALE__ === 'US') {
+    styles['layout:legacy'] = legacyHtml;
+    styles['layout:custom'] = custom;
```
Seavenly, line 14: If we're going to switching to a dynamic plugin type architecture, I wonder if these could be plugin rule sets as well.

tests/unit/spec/src/messages/models/Banner/validateOptions.test.js
```diff
@@ -83,26 +82,6 @@ describe('validateOptions', () => {
             expect(validated.amount).toBeUndefined();
         });
 
-        it('Warns invalid countryCode type', () => {
-            const validated = validateOptions(mockLogger, {
-                ...validOptions,
```
Seavenly, line 88: We need to remember that some of this code involving the use of the `countryCode` option may return since the webpack variable solution will not be supported as we'd hoped.

Seavenly (changes requested):

src/messages/models/Banner/index.js
gillycheesesteak: It was useful in my debugging, and will be tree-shaken in the prod build, so I think it makes sense to leave it in

demo/standalone.html
```diff
@@ -4,17 +4,18 @@
         <title>Messaging.js Dev Sandbox</title>
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no" />
-        <script src="./messaging.js" data-pp-account="DEV000NINONUS"></script>
+        <script src="./messaging.js" data-pp-account="DEV0000000IAZ" data-pp-country="DE"></script>
```
Seavenly: Is `data-pp-country="DE"` being used?

src/locale/DE/validOptions.js
```diff
@@ -0,0 +1,20 @@
...
+    text: {
+        logo: {
+            type: [Types.STRING, ['primary', 'alternative', 'inline', 'none']]
+        },
+        text: {
+            color: [Types.STRING, ['black', 'white']]
+        }
+    },
+    flex: {
+        color: [Types.STRING, ['blue', 'black', 'white', 'white-no-border', 'gray|grey']],
```
Seavenly: Are we removing `'white-no-border'` from DE?

src/locale/DE/validOptions.js
```diff
@@ -0,0 +1,20 @@
...
+        ratio: [Types.STRING, ['1x1', '1x4', '8x1', '20x1']]
+    },
+    custom: {
```
Seavenly: Should we remove this since it will not be a use-case at launch?

tests/unit/spec/src/messages/models/Banner/validateOptions.test.js
```diff
@@ -19,11 +19,10 @@ describe('validateOptions', () => {
         _legacy: false,
         account: '1234567890123',
         amount: 100,
-        countryCode: 'US',
         onRender: () => {}
     };
 
-    it('Allows valid id, _legacy, account, amount, countryCode, style', () => {
+    it('Allows valid id, _legacy, account, amount, country, style', () => {
```
Seavenly: Are we using a `country` prop anywhere?

tests/utils/setup.js
```diff
@@ -1 +1,5 @@
 import '@testing-library/jest-dom/extend-expect';
+
+window.__MESSAGES__ = {
```
Seavenly: Is this necessary because webpack isn't involved so the stand-in global variables aren't being replaced? Should we just import all the variables in `globals.js` into the window object?

src/locale/DE/validOptions.js
gillycheesesteak: Yes

tests/utils/setup.js
Seavenly: Alternatively, can they be injected with the `globals` field in the Jest config?

Seavenly (approved):

PR #66, opened 2019-10-10
feat: adding the landingUrl config option

roja079:


src/messages/models/Modal/index.js
```diff
@@ -325,19 +325,33 @@ export default {
...
-
                 if (target.tagName === 'IMG' && target.parentNode.tagName === 'A') {
-                    window.open(
-                        target.parentNode.href,
-                        'PayPal Credit Terms',
-                        'width=650,height=600,scrollbars=yes,resizable=no,location=no,toolbar=no,menubar=no,dependent=no,dialog=yes,minimizable=no'
-                    );
-
-                    evt.preventDefault();
+                    if (options.landingurl !== undefined && options.landingurl !== null) {
```
gillycheesesteak: This condition will never be true, because a legacy NI banner (the condition above) will never have a landing URL

src/messages/models/Modal/index.js
```diff
@@ -325,19 +325,33 @@ export default {
...
+                        evt.preventDefault();
+                    }
                 } else {
                     window.open(meta.clickUrl, '_blank');
                 }
             });
+        } else if (
+            options.style.layout === 'custom' &&
+            options.landingurl !== undefined &&
+            options.landingurl !== null
```
gillycheesesteak: You should be able to replace this with just `if (options.style.layout === 'custom' && options.landingurl')`

src/messages/models/Modal/index.js
```diff
@@ -325,19 +325,33 @@ export default {
...
+        ) {
+            events.on('click', evt => {
+                const { target } = evt;
+                if (target.tagName === 'IFRAME' || target.tagName === 'H1') {
```
gillycheesesteak: I don't think we want to validate the target at all, because the custom banner could have any kind of markup in it

src/messages/models/Template/mutations/ezp_any_eqz.js
```diff
@@ -43,7 +45,7 @@ export default {
             {
                 logo: Logo.ALTERNATIVE.COLOR,
                 headline: {
-                    replace: [['APR', 'APR.']],
+                    replace: [['APR', 'APR']],
```
gillycheesesteak: This line does nothing, it can be removed. This applies to any case where the two values passed to replace are the same.

src/messages/models/Modal/index.js
roja079: After making changes to the code i forgot to remove the if condition. Removed now.

src/messages/models/Modal/index.js
roja079: Done

src/messages/models/Modal/index.js
roja079: Done

src/messages/models/Template/mutations/ezp_any_eqz.js
roja079: Done

src/messages/models/Banner/index.js
```diff
@@ -90,6 +90,7 @@ const Banner = {
                 ) // Promise<Object(options, logger, wrapper, events, markup, template, meta)>
                 .then(
                     pipe(
+                        passThrough,
```
gillycheesesteak: What is the purpose of this line?

src/messages/models/Modal/index.js
```diff
@@ -338,6 +337,11 @@ export default {
                     window.open(meta.clickUrl, '_blank');
                 }
             });
+        }
+        if (options.style.layout === 'custom' && options.landingurl) {
```
gillycheesesteak: This needs an else from the previous condition

src/messages/models/Template/mutations/ezp_any_gtz.js
```diff
@@ -24,6 +24,7 @@ export default {
                 messageWidth: [250, 1000],
                 logo: Logo.ALT_NO_PP.COLOR,
                 headline: {
+                    replace: [['months', 'months']],
```
gillycheesesteak: We can remove all of the `replace` rules where the two values are the same - they have no effect

src/messages/models/Banner/validateOptions.js
```diff
@@ -152,7 +152,7 @@ function getValidStyleOptions(logger, options) {
  * @param {Object} options User options object
  * @returns {Object} Object containing only valid options
  */
-export default curry((logger, { account, amount, countryCode, style, offer, ...otherOptions }) => {
+export default curry((logger, { account, amount, countryCode, style, offer, landingurl, ...otherOptions }) => {
```
gillycheesesteak: Can we rename this field to `landingUrl`?

src/messages/models/Banner/index.js
roja079: That was kept for testing purpose, removed now.

src/messages/models/Modal/index.js
roja079: Done

src/messages/models/Template/mutations/ezp_any_gtz.js
roja079: Done

src/messages/models/Banner/validateOptions.js
roja079: Done

src/messages/models/Template/mutations/ni_non-us.js
```diff
@@ -8,7 +8,7 @@ export default {
             {
                 styles: [basicMediaQuery(320)],
                 logo: Logo.PRIMARY.COLOR,
-                headline: ['xsmall', { tag: 'medium', br: ['months'], replace: [['99+', '99+.']] }],
+                headline: ['xsmall', { tag: 'medium', br: ['months'], replace: [['99+', '99+']] }],
```
gillycheesesteak: The replace rule here has no effect and can be removed

src/messages/models/Template/mutations/ni_non-us.js
```diff
@@ -49,7 +50,7 @@ export default {
             {
                 styles: [basicMediaQuery(570)],
                 logo: Logo.ALTERNATIVE.COLOR,
-                headline: ['xsmall', { tag: 'medium', br: ['months'], replace: [['99+', '99+.']] }]
+                headline: ['xsmall', { tag: 'medium', br: ['months'], replace: [['99+', '99+']] }]
```
gillycheesesteak: Remove replace rule

banners/ezp_any_eqz.json
```diff
@@ -6,11 +6,11 @@
     },
     "data": {
         "headline": [
-            ["Pay over ${term} months at 0% APR", ["default", "small"]],
+            ["Pay over ${term} months at 0% APR.", ["default", "small"]],
             ["Enjoy ${term} Easy Payments with No Interest.", ["legacy-small"]]
         ],
         "subHeadline": [
-            ["Choose PayPal Credit and select ${term} Easy Payments to pay over time", ["legacy-medium"]],
+            ["Choose PayPal Credit and select ${term} Easy Payments to pay over time.", ["legacy-medium"]],
```
gillycheesesteak: I think we should leave all the legacy messages the same. We could update the replace rules to match, but since we don't expect custom banners to use these messages, I think it is easier to just revert the changes to the messages

PR #83, opened 2019-11-11
feat: font size options

gillycheesesteak:


src/locale/US/mutations/ezp_any_eqz.js
```diff
@@ -1,52 +1,43 @@
...
                 },
                 disclaimer: 'xsmall.2'
             }
         ],
-        ['logo.type:primary', { messageWidth: 190 }],
         [
             'logo.type:inline',
-            {
-                messageWidth: [255, 1000],
+            style => ({
```
Seavenly: Should we try and be a little forward thinking and destructure an object with various props?
`({ style }) => ({`
I also wonder if passing in the explicit data would add more clarity?
`({ textSize }) => ({`

src/messages/models/Template/index.js
```diff
@@ -362,6 +390,10 @@ function createTemplateNode(options, markup) {
         return styles;
     };
 
+    if (layout === 'text' && objectGet(options, 'style.text.fontFamily')) {
```
Seavenly, line 398: Is this only supposed to be available for only custom banners or all text messages?

src/messages/models/Banner/validateOptions.js
```diff
@@ -87,10 +87,19 @@ function populateDefaults(logger, defaults, options, prefix = 'style.') {
...
-    return {
+    const defaultValues = {
         layout: options.layout,
         ...populateDefaults(logger, localeStyleOptions[options.layout], options)
     };
+
+    if (options.layout === 'text') {
+        if (defaultValues.text.size > 16 || defaultValues.text.size < 10) {
+            logInvalid(logger, 'style.text.size', 'Text size should be between 10px and 16px.');
+            objectSet(defaultValues, 'text.size', 12);
```
Seavenly: Would we rather add acceptable integer values 10 - 16 to the validOptions object and validate similarly to other props?

src/messages/models/Banner/validateOptions.js
```diff
@@ -87,10 +87,19 @@ function populateDefaults(logger, defaults, options, prefix = 'style.') {
...
  */
 function getValidStyleOptions(logger, localeStyleOptions, options) {
-    return {
+    const defaultValues = {
         layout: options.layout,
         ...populateDefaults(logger, localeStyleOptions[options.layout], options)
     };
+
+    if (options.layout === 'text') {
+        if (defaultValues.text.size > 16 || defaultValues.text.size < 10) {
```
Seavenly: Size values that come in via inline props will be a string.

src/messages/models/Template/index.js
gillycheesesteak: It should be available for all text messages (but not an advertised parameter) to avoid needing a custom banner for a merchant that only wants to modify the font

Seavenly:
Is it possible to have a dynamic `max-width` on the message text, so that we can ensure the "See terms" is always on a second line for the primary logo option?

Seavenly (approved):

PR #94, opened 2019-11-27
fix: use normalized correlation ID header

Seavenly:


src/messages/services/logger/index.js
```diff
@@ -82,8 +82,10 @@ export const Logger = {
             })
                 .then(res => {
                     // Same correlation id duplicated in prod
-                    const [corrId] = (res.headers['Paypal-Debug-Id'] || '').split(',');
-                    setState({ history: [...state.history, corrId].slice(-5) });
+                    const [corrId] = (res.headers['paypal-debug-id'] || '').split(',');
```
gillycheesesteak, line 85: Will the header name always be lowercase?

Seavenly: I believe so. Verified in Chrome, Firefox, and IE 11.

gillycheesesteak (approved):

PR #99, opened 2019-12-06
test: add regression test suite

gillycheesesteak:


package.json
```diff
@@ -42,6 +45,7 @@
         "@testing-library/jest-dom": "^4.1.0",
         "babel-eslint": "^10.0.1",
         "babel-jest": "^24.8.0",
+        "core-js-compat": "^3.4.7",
```
Seavenly: This can be removed I believe.

demo/standalone.html
```diff
@@ -4,7 +4,7 @@
         <title>Messaging.js Dev Sandbox</title>
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no" />
-        <script src="./messaging.js" data-pp-account="DEV00000000NI"></script>
+        <script src="./messaging.js" data-pp-account="8SUQZGVVX324C" data-pp-currency="EUR"></script>
```
Seavenly: Reminder to swap this back.

demo/text.html
```diff
@@ -33,7 +33,7 @@
         <script>
             // prettier-ignore
             var devAccounts = [
-                'DEV00000000NI',
+                // 'DEV00000000NI',
```
Seavenly: Reminder to swap this back.

src/legacy/index.js
```diff
@@ -292,6 +292,7 @@ scripts.some(script => {
     const payerId = script.getAttribute('data-pp_payer_id');
     const dimensions = script.getAttribute('data-pp_dimensions');
     const account = payerId || pubId;
+    console.log(account, dimensions);
```
Seavenly: Can this be removed?

src/legacy/toNewPipeline.js
```diff
@@ -105,6 +105,7 @@ export default function toNewPipeline(ppScript) {
     const account = kvs.payer_id || kvs.pub_id;
     const [typeNI, typeEZP, dimension, style] = getSupportedAttributes(kvs.dimensions, kvs.style);
 
+    console.log(kvs);
```
Seavenly: Can this be removed?

.github/workflows/update-snapshots.yml
```diff
@@ -0,0 +1,25 @@
...
+            - uses: actions/checkout@v1
+            - name: Check for trigger commit
+              run: ./scripts/snapshot-tests/should-update.sh
+            - name: Use Node.js 10
+              if: env.UPDATE_SNAPSHOT == 1
+              uses: actions/setup-node@v1
+              with:
+                  node-version: 10.x
+            - name: if trigger commit found, update snapshots
+              if: env.UPDATE_SNAPSHOT == 1
```
Seavenly: Is it possible at all to nest steps under a single conditional?

scripts/snapshot-tests/setup-tests.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+echo "127.0.0.1 localhost.paypal.com" | sudo tee -a /etc/hosts
```
Seavenly: I'm not much of a bash ninja, but would this be simpler?
``` bash
echo "127.0.0.1 localhost.paypal.com" >> /etc/hosts
```

scripts/snapshot-tests/setup-tests.sh
```diff
@@ -0,0 +1,6 @@
...
+npm install
+npm run dev:standalone &
+sleep 20
```
Seavenly: Can we add a brief comment on why we're waiting here and pushing the above process into the background?

tests/functional/spec/createBannerTest.js
```diff
@@ -0,0 +1,85 @@
...
+    }
+};
+
+export default function createBannerTest(locale, legacy = false) {
+    return (viewport, config) => {
+        const testNameParts = getTestNameParts(locale, config);
+        test(testNameParts.slice(-1)[0], async () => {
+            await page.setViewport(viewport);
+
+            const testPage = legacy ? 'legacy_banner.html' : 'banner.html';
```
Seavenly: Would `legacy` be better named directly as `testPage` with a default value of `'banner.html'`?

tests/functional/spec/createBannerTest.js
```diff
@@ -0,0 +1,85 @@
...
+            await page.goto(`http://localhost.paypal.com:8080/${testPage}?config=${JSON.stringify(config)}`);
+
+            await waitForBanner(1000);
+
+            await new Promise(resolve => setTimeout(resolve, 500));
```
Seavenly: Can we add a comment why this timeout is necessary after a function called `waitForBanner`? Should it be moved into `waitForBanner`?

scripts/snapshot-tests/setup-tests.sh
gillycheesesteak: The `sudo` is required to modify `/etc/hosts`, and `sudo echo` does not work

.github/workflows/update-snapshots.yml
gillycheesesteak: From what I've seen in the Github Actions docs, the only way to maybe do this would be to create separate jobs, but I'm not sure that would work and even if it did, I think this would still be cleaner. See here: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idsteps

scripts/snapshot-tests/setup-tests.sh
Seavenly: Gotcha. That makes sense.

Seavenly (approved):

scripts/snapshot-tests/run-tests.sh
```diff
@@ -0,0 +1,15 @@
...
+    npm run test:func -- -u
+
+    # Uses GNU sed syntax
+    sed -i -e 's/DIRTY_SNAPSHOTS=1/DIRTY_SNAPSHOTS=0/g' .travis.yml
+    
+    git remote set-url origin "https://${GH_TOKEN}@github.com/${TRAVIS_REPO_SLUG}.git"
+
+    git checkout develop
+    git add .
+    git commit -m "update snapshots [skip ci]"
```
Seavenly: Can we prefix this with something like `chore(snapshots): `?

scripts/snapshot-tests/run-tests.sh
```diff
@@ -0,0 +1,15 @@
+if [[ "${DIRTY_SNAPSHOTS}" != "1" ]]; then
```
Seavenly: Would `$DIRTY_SNAPSHOTS` work instead of `"${DIRTY_SNAPSHOTS}"` for these checks?

gillycheesesteak: I used the quotes to make the script robust against the variable not being defined at all (see this answer: https://stackoverflow.com/questions/2237080/how-to-compare-strings-in-bash)

Looks like the braces are not necessary though

Seavenly (approved):

PR #136, opened 2020-02-10
refactor: adjust alternative message wrapping

merlinstardust:


banners/US/ni.json
```diff
@@ -6,10 +6,10 @@
     },
 
     "headline": [
-        ["Buy now. Pay over time.", ["default", "xsmall", "small"]],
+        ["Buy now. Pay over time.", ["default"]],
         [
             ["No Interest if paid in full in 6 months", ["on purchases of $99+", "weak"]],
-            ["medium", "large", "xlarge"]
+            ["xsmall", "small", "medium", "large", "xlarge"]
```
Seavenly, line 12: We do not want to change already established message/tag relationships as we have custom banners that rely on certain tags to deliver specific messages. Instead we should update the tag value inside of our mutation files.

src/locale/US/mutations/pala_multi_gtz.js
```diff
@@ -15,6 +15,7 @@ export default {
             'logo.type:inline',
             ({ textSize }) => ({
                 styles: [`.message__logo { width: ${textSize * 7}px }`],
+                messageWidth: [0, 1000],
```
Seavenly, line 18: This value effectively does nothing but reset this property. If we are resetting it in multiple places, perhaps we should consider moving the default `messageWidth` value to the logo types where it applies. Another option to try would be to set the value to `false` instead and that may override the previous value.

Seavenly (changes requested):
These changes should go into `develop` instead of `modal-apply-now`. Also probably need to pull the latest changes in `develop`.

PR #141, opened 2020-02-14
feat: standalone modal

merlinstardust:
This has a dependency on Zoid modals and the apply now button


src/messages/models/Banner/validateOptions.js
```diff
@@ -195,6 +196,24 @@ export default curry((logger, { account, amount, style, offer, ...otherOptions }
         validOptions.style = { layout: 'text' };
     }
 
+    if (typeof modal !== 'undefined') {
```
gillycheesesteak: Did we decided that we want to allow the user to select the type of modal that is displayed? My understanding was that we should still request the message to determine this

src/messages/models/Banner/validateOptions.js
```diff
@@ -195,6 +196,24 @@ export default curry((logger, { account, amount, style, offer, ...otherOptions }
...
+            if (!validateType(Types.STRING, modal.type)) {
+                logInvalidType(logger, 'modal.type', Types.STRING, modal.type);
+                validOptions.modal = { type: 'NI' };
+            }
+        } else {
+            logInvalidType(logger, 'modal', Types.OBJECT, modal);
+        }
+
+        if (otherOptions._legacy) {
+            logInvalid(logger, 'modal', 'Cannot have both legacy and modal options set');
```
gillycheesesteak: I don't think we need to log a message for this - the `_legacy` option is set by us internally, and any place where it is set, the user should not be configuring the modal option

src/messages/models/Container/index.js
```diff
@@ -5,22 +5,26 @@ import insertMarkup from './insertMarkup';
...
     }
 
-    const helperFns = objectEntries({
-        insertMarkup,
-        setSize,
-        runStats,
-        events
-    }).reduce((accumulator, [fnName, fn]) => ({ ...accumulator, [fnName]: fn(container) }), {});
+    const helpers = { runStats, events };
+    if (!isOnlyModal) {
```
gillycheesesteak: Can we make these helpers just a noop if the isOnlyModal option is true?

src/messages/models/Container/index.js
```diff
@@ -5,22 +5,26 @@ import insertMarkup from './insertMarkup';
 import runStats from './stats';
 import setSize from './setSize';
 
-export default function createContainer(type) {
-    const container = document.createElement(type);
+export default function createContainer(typeOrElement, isOnlyModal = false) {
```
gillycheesesteak: Instead of passing down `isOnlyModal` as a separate value everywhere, does it make sense to add it to the `options` object similar to how we use `_legacy`, something like `_modalOnly`

merlinstardust: This can be done but the options object would still need to be passed down to the container. Is that fine?

src/messages/models/Container/index.js
merlinstardust: What's the benefit of these being a noop?

src/messages/models/Banner/validateOptions.js
merlinstardust: Isn't it possible for the consumer to set the `_legacy` option during render? In which case shouldn't we warn them of the conflict? What if they think they can have both a legacy and modal only option?

src/messages/models/Container/index.js
gillycheesesteak: Yea, it just makes these call signatures more consistent, and in some places, removes an extra parameter in favor of using the options object that's already passed in

src/messages/models/Container/index.js
gillycheesesteak: I wonder if there's any harm in just leaving these helpers on, and just not calling them in the render pipeline if its only a modal. That keeps this code cleaner here

src/messages/models/Banner/validateOptions.js
gillycheesesteak: While it's technically possible to do that, it is not supported or advertised - that parameter is meant for us to use internally. So I would say that we don't need to warn about it

src/messages/models/Banner/validateOptions.js
```diff
@@ -149,7 +150,7 @@ export const validateStyleOptions = curry((logger, style) => {
  * @param {Object} options User options object
  * @returns {Object} Object containing only valid options
  */
-export default curry((logger, { account, amount, style, offer, ...otherOptions }) => {
+export default curry((logger, { account, amount, style, offer, modal, ...otherOptions }) => {
```
gillycheesesteak: Is the `modal` option still used/needed here?

merlinstardust: No, I missed that one. I'll remove it

gillycheesesteak (approved):

src/messages/controllers/render.js
```diff
@@ -70,6 +70,11 @@ export default function render(options, selector) {
 
             totalOptions.id = container.getAttribute('data-pp-id');
 
+            const modalOnlyAttribute = container.getAttribute('data-pp-modalonly');
+            if ((modalOnlyAttribute && modalOnlyAttribute !== 'false') || totalOptions.modalonly) {
```
gillycheesesteak: Should we just check for the existence of `data-pp-modalonly` here instead of checking for a value?

Seavenly:
Closing since the codebase has been drastically refactored.

PR #150, opened 2020-02-26
test: add modal functional tests

yeul:


tests/functional/content/modal-test.html
```diff
@@ -0,0 +1,64 @@
...
+
+    <body>
+        <!-- Hides banners behind a white box to prevent diff issues if banner design changes. -->
+        <div
+            class="hide-banner"
+            style="position: absolute; height: 500px; width: 100%; background: white; z-index: 1; pointer-events: none;"
+        ></div>
+        <div class="container banner-1"></div>
+        <br />
+        <div class="container banner-2" data-pp-account="DEV00000000NI" data-pp-style="text"></div>
```
gillycheesesteak: Should this be `data-pp-style-layout="text"`? Or can this attribute be removed?

tests/functional/content/modal-test.html
```diff
@@ -0,0 +1,64 @@
...
+            const query = window.location.search.substring(1);
+            const mappedParams = query.split('&').reduce((accumulator, param) => {
+                const [key, value] = param.split('=');
+                accumulator[key] = decodeURIComponent(value);
+                return accumulator;
+            }, {});
+            const config = JSON.parse(mappedParams.config);
+
+            if (config.style.text && config.style.text.color === 'white') {
+                document.body.style = 'background: #000';
```
gillycheesesteak: Is this necessary for the modal tests?

tests/functional/jest.config.js
```diff
@@ -2,5 +2,7 @@ module.exports = {
     testMatch: ['<rootDir>/**/?(*.)test.js?(x)'],
     globalSetup: 'jest-environment-puppeteer/setup',
     globalTeardown: 'jest-environment-puppeteer/teardown',
-    testEnvironment: 'jest-environment-puppeteer'
+    testEnvironment: 'jest-environment-puppeteer',
+    testTimeout: 180000,
+    preset: 'jest-puppeteer'
```
gillycheesesteak: Are these changes still necessary?

tests/functional/spec/DE/de_modalTests/de_modalCalc-text.test.js
```diff
@@ -0,0 +1,57 @@
...
+        `update finance terms when user updates amount passed into calculator ${bannerStyle.layout} ${viewport.width}`,
+        updateFinanceTerms(account, viewport, bannerStyle)
+    );
+    test(
+        `when an amount is passed into DE PP installments message, the correct financing terms are loaded in a table ${bannerStyle.layout} ${viewport.width}`,
+        deModalContentAndCalc(account, viewport, bannerStyle)
+    );
+    afterEach(async () => {
+        // eslint-disable-next-line no-undef
+        await jestPuppeteer.resetPage();
```
gillycheesesteak: Can we move this `afterEach` to some shared location that applies to all the tests?
Also, we should be able to tell eslint about the global `jestPuppeteer` object so we don't need this disable comment

tests/functional/spec/DE/de_modalTests/de_modalCalc-flex.test.js
```diff
@@ -0,0 +1,55 @@
...
+
+const toMatchImageSnapshot = configureToMatchImageSnapshot({
+    failureThresholdType: 'percent',
+    failureThreshold: 0.002,
+    customDiffConfig: {
+        threshold: 0.05
+    }
+});
+
+expect.extend({ toMatchImageSnapshot });
```
gillycheesesteak: Can we look for a way to apply this to all tests in a common place?

tests/functional/spec/utils/testStylesConfig.js
```diff
@@ -0,0 +1,26 @@
+// Permutations used in modal tests
+const viewports = [
+    { width: 1920, height: 1080 },
+    { width: 411, height: 731 }
+];
+
+const bannerStyles = [
```
gillycheesesteak, line 7: Can we format these in the same structure as the `style` object that gets passed to `openModal`, so that the style can just be passed along as it is without referencing each individual property?

tests/functional/spec/utils/us_modalTestDefs.js
```diff
@@ -0,0 +1,279 @@
...
+        )
+    );
+    const elementModal = await page.$("iframe[title='paypal_credit_modal']");
+
+    const modalFrame = await elementModal.contentFrame();
+    await modalFrame.evaluate(browserVal => Boolean(document.querySelector(browserVal)), selectors.modal.container);
+    await modalFrame.waitForSelector(selectors.modal.modalContent);
+    await modalFrame.waitForSelector(selectors.modal.modalMain);
+    await modalFrame.waitForSelector(selectors.modal.contentBody);
+    await modalFrame.waitForSelector(selectors.modal.contentBodyTitle);
```
gillycheesesteak: Are all of these `waitForSelector`s necessary?

tests/functional/spec/utils/globalModalTestDefs.js
```diff
@@ -0,0 +1,111 @@
...
+    const image = await page.screenshot(
+        {
+            clip: {
+                ...viewport,
+                x: 0,
+                y: 0
+            }
+        },
+        3
+    );
```
gillycheesesteak: Since we are always passing the same options here, can we move the actual screenshot into `modalSnapshot`?

tests/functional/content/modal-test.html
yeul: I've gone ahead and removed the class since it was unnecessary.

tests/functional/content/modal-test.html
yeul: The config is necessary, however, I've removed the if statement!

tests/functional/jest.config.js
yeul: I was able to remove the preset and the associated npm install from package.json, but the testTimeout is necessary as because puppeteer is trying to load in the iframes of the multiple banners on screen. It doesn't take 180000 seconds, but this is the buffer I've found that works consistently.

tests/functional/spec/DE/de_modalTests/de_modalCalc-text.test.js
yeul: I've added the jestPuppeteer global to the eslintrc. Also, I've add the afterEach inside of a new file called jest.setupTests.js and it is initiated inside of jest.config.js using a config option called setupFilesAfterEnv.

tests/functional/spec/utils/testStylesConfig.js
yeul: Well, I think the main reason why I passed in each individual option (other than trying to avoid having repeated arrays in each test file) was because modalSnapshot is taking those pieces (account, viewport, bannerStyle) and is creating a snapshot file based off of those permutations. Creating a unified style object without any sort of dot notation would sort of throw off how modalSnapshot is fundamentally filtering the snapshots into the appropriate place. Plus, I think it wouldn't necessarily be any neater since you would have to pass in something like styles.bannerStyle.layout or styles.viewport.width into modalSnapshot to get it working properly. Also, the describe.each blocks would probably look a bit more cluttered as well since they'd now have to be something like styles.viewports[0] vs. just viewports[0].

tests/functional/spec/utils/us_modalTestDefs.js
yeul: Just ran the tests again, they passed without them so I've removed them (and others)! I've left in some `waitForSelector`s in certain spots in some tests where there is a lot of clicking or inputing numbers into the calc form.

.eslintrc
```diff
@@ -5,7 +5,8 @@
         "browser": true
     },
     "globals": {
-        "__MESSAGES__": true
+        "__MESSAGES__": true,
+        "jestPuppeteer": true
```
gillycheesesteak: Let's add this to the `.eslintrc` in the tests/functional folder so it only applies to those files

tests/functional/spec/utils/us_modalTestDefs.js
```diff
@@ -0,0 +1,249 @@
...
+    const contentHeaderTitle = await modalFrame.evaluate(
+        _contentHeaderTitle => document.querySelector(_contentHeaderTitle).innerText,
+        selectors.modal.contentHeaderTitle
+    );
+    const contentBodyTitle = await modalFrame.evaluate(
+        _contentBodyTitle => document.querySelector(_contentBodyTitle).innerText,
+        selectors.modal.contentBodyTitle
+    );
+    // eslint-disable-next-line no-unused-expressions
+    expect(contentHeaderTitle).toContain('Buy now and pay over time with PayPal Credit') &&
```
gillycheesesteak: Why does this need to be one line?

tests/functional/spec/utils/us_modalTestDefs.js
```diff
@@ -0,0 +1,249 @@
...
+    modalSnapshot(`${testNameParts} ${bannerStyles[0].layout}`, viewport, image, account);
+};
+
+export const updateFinanceTerms = (account, viewport, bannerStyles) => async () => {
+    const testNameParts = 'update finance terms';
+    const elementModal = await page.$("iframe[title='paypal_credit_modal']");
+    const modalFrame = await elementModal.contentFrame();
+    await modalFrame.waitForSelector(selectors.calculator.calc, { visible: true });
+    await modalFrame.waitForSelector(selectors.calculator.calcForm);
+    await modalFrame.waitForSelector(selectors.calculator.calcInput, { visible: true });
```
gillycheesesteak: I still think we have a lot of repetitive `waitForSelector`s - remember that a lot of the inconsistency we saw was related to missing `await`s or strangeness with puppeteer - we should only need to wait for a single selector between interactions

tests/functional/spec/DE/de_modalTests/de_modalCalc-flex.test.js
yeul: Fixed! I've added it to `modalSnapshot.js`.

tests/functional/spec/utils/us_modalTestDefs.js
yeul: Fixed!

.eslintrc
yeul: Fixed!

tests/functional/spec/utils/globalModalTestDefs.js
yeul: Fixed! Moved into `modalSnapshot`.

tests/functional/spec/utils/testStylesConfig.js
yeul: After clarification, fixed!

tests/functional/spec/utils/us_modalTestDefs.js
yeul: No worries, I've removed more `waitForSelector`s.

tests/functional/spec/utils/us_modalTestDefs.js
```diff
@@ -0,0 +1,152 @@
...
+    const testNameParts = 'ni content in modal';
+    await page.waitForFunction(() =>
+        Array.from(document.querySelectorAll("iframe[title='paypal_credit_modal']")).find(
+            el => el.parentElement.parentElement.style.display !== 'none'
+        )
+    );
+    const elementModal = await page.$("iframe[title='paypal_credit_modal']");
+
+    const modalFrame = await elementModal.contentFrame();
+    await modalFrame.evaluate(browserVal => Boolean(document.querySelector(browserVal)), selectors.modal.container);
```
gillycheesesteak: I think this should be a `modalFrame.waitForFunction`

tests/functional/spec/utils/us_modalTestDefs.js
```diff
@@ -0,0 +1,152 @@
...
+};
+
+/**
+ * Runs inside us_ModalFunc-flex & us_ModalFunc-text for the US locale.
+ */
+
+export const clickHereSeeTerms = (account, viewport, bannerStyle) => async () => {
+    const testNameParts = 'see terms page on modal hyperlink click';
+    await page.waitFor(1000);
+    const elementModal = await page.$("iframe[title='paypal_credit_modal']");
```
gillycheesesteak, line 41: Should these lines where we use a simple `page.$` for the modal iframe be updated to use this method:
`await page.waitForFunction(() =>
        Array.from(document.querySelectorAll("iframe[title='paypal_credit_modal']")).find(
            el => el.parentElement.parentElement.style.display !== 'none'
        )
    );`

And if so, we can put that check into a utility method in a shared location for easy use everywhere

tests/functional/spec/de_initalizeModal.js
```diff
@@ -0,0 +1,33 @@
...
+    await page.waitForSelector("iframe[title='paypal_credit_modal']");
+
+    const elementHandle = await page.$('.banner-1 iframe');
+    const elementModal = await page.$("iframe[title='paypal_credit_modal']");
+
+    const frame = await elementHandle.contentFrame();
+    const modalFrame = await elementModal.contentFrame();
+
+    await frame.waitForSelector('.message__messaging', { visible: true });
+    await modalFrame.waitForSelector('.modal__content');
```
gillycheesesteak: Can we merge this and `us_initializeModal` into a single function? The only difference seems to be the content selector we wait for here, so we can have the country be a param that's passed in and decide which selector to use based on that.

Also, `initialize` is spelled wrong :)

tests/functional/spec/utils/us_modalTestDefs.js
yeul: I can actually just get rid of it since it's not in any of my other tests and works without it.

tests/functional/spec/utils/us_modalTestDefs.js
yeul: This more detailed check is actually already being done in a single location inside of `openModal` so it is running prior to each test inside of beforeEach.

tests/functional/spec/de_initalizeModal.js
yeul: I've made a new file called `initializeModal` that consolidates the two (with proper spelling this time lol) :^)

PR #157, opened 2020-03-04
feat: add support for NI qualifying message

yeul:


banners/US/ni.json
```diff
@@ -1,7 +1,7 @@
 {
     "meta": {
         "offerCountry": "US",
-        "offerType": "NI",
+        "offerType": "NI:${CREDIT_OFFERS_DS.qualifying_offer}",
```
gillycheesesteak: Why are we using the template variable here?

src/locale/US/mutations/ni_qual.js
```diff
@@ -0,0 +1,169 @@
+import Logo from '../logos';
+import { basicMediaQuery } from './ni';
+
+export default {
+    'layout:text': [
```
gillycheesesteak, line 10: I wonder if, while these mutations are so similar with the regular NI mutations, we put them in a common place that both files can import and this file can modify slightly

yeul: Like we talked about, there's an argument to be made for keeping it in two separate files vs. in one, but honestly I think keeping it in two separate files would avoid some level of confusion. There are things inside of ni.js that aren't inside of ni-qual.js (i.e. legacy) so keeping it separate sort of adds to that distinction. Plus, like you said, when something inevitably changes, we might end up having to refactor it back into two files anyways. I've added a comment at the top of ni.js and ni_qual.js to say when said mutation file applies (i.e. in ni_qual: Mutation used when NI:QUALIFYING-true). To me, two separate files feels like a logical separation, rather than an arbitrary one.

Seavenly:
Closing in favor of #173

PR #182, opened 2020-03-31
feat: detect message overflow 

yeul:
… message container height.


src/messages/models/Container/setSize.js
```diff
@@ -246,4 +247,54 @@ export default curry((container, { wrapper, options, logger, meta }) => {
...
+
+        if (typeof window.IntersectionObserver === 'undefined') {
+            const polyfillUrl = 'https://polyfill.io/v3/polyfill.js?features=IntersectionObserver';
+            dynamicImport(polyfillUrl)
+                .then(() => {
+                    containerHeightCheck();
+                })
+                .then(() => {
+                    const scriptEl = document.querySelector(`[src="${polyfillUrl}"]`);
+                    scriptEl.parentNode.removeChild(scriptEl);
```
Seavenly: Can this be handled inside `dynamicImport()` right before we resolve in the `'load'` event?

src/messages/models/Container/setSize.js
```diff
@@ -246,4 +247,54 @@ export default curry((container, { wrapper, options, logger, meta }) => {
...
+                });
+        } else containerHeightCheck();
```
Seavenly: Can we wrap this in brackets to improve readability as this is a more uncommon syntax?

src/messages/models/Container/setSize.js
```diff
@@ -246,4 +247,54 @@ export default curry((container, { wrapper, options, logger, meta }) => {
...
+         *
+         * If browser support exists, containerHeightCheck runs as normal.
+         */
+
+        if (typeof window.IntersectionObserver === 'undefined') {
+            const polyfillUrl = 'https://polyfill.io/v3/polyfill.js?features=IntersectionObserver';
+            dynamicImport(polyfillUrl)
+                .then(() => {
+                    containerHeightCheck();
+                })
```
Seavenly: Consider simplifying to `.then(containerHeightCheck)`

src/messages/models/Container/setSize.js
```diff
@@ -246,4 +247,54 @@ export default curry((container, { wrapper, options, logger, meta }) => {
...
+
+    /**
+     * IntersectionObserver checks to see if there is any intersection of elements up the tree that
+     * could cause the messages to be cut off.
+     * The element we are observing is the .messages div. If any ancestor elements of the .messages div
+     * is a container with a set height smaller than required for the message, the message will be set to display:none.
+     */
+
+    if (options.style.layout === 'text' && wrapper.parentNode.parentNode && logger) {
+        const containerHeightCheck = () => {
```
Seavenly: Can this be more generically named to something like `setupOverflowObserver` and then can we use this to check for both width and height issues, removing the need for our previous width checking logic?

src/messages/models/Container/setSize.js
```diff
@@ -246,4 +247,54 @@ export default curry((container, { wrapper, options, logger, meta }) => {
...
+                entries.forEach(entry => {
+                    if (entry.intersectionRatio < 0.9) {
+                        logger.warn(
+                            `Message hidden. PayPal Credit Message of layout type text requires a height of at least ${parseInt(
+                                container.getAttribute('height'),
+                                10
+                            ) + 1}px. Current container is ${el.parentNode.clientHeight}px. Message hidden.`
+                        );
+                        // eslint-disable-next-line no-param-reassign
+                        wrapper.style.display = 'none';
```
Seavenly: This should follow the same hiding logic as the width so that we can allow for a fallback message attempt.

src/messages/models/Container/setSize.js
```diff
@@ -246,4 +247,54 @@ export default curry((container, { wrapper, options, logger, meta }) => {
...
+                    }
+                });
+            }, {});
+            if (document.querySelector('.messages')) observer.observe(document.querySelector('.messages'));
```
Seavenly: We cannot guarantee that the merchant will use `.messages` as their selector. We should likely observe each message using use `wrapper.parentNode`.

src/messages/models/Container/setSize.js
```diff
@@ -246,4 +247,54 @@ export default curry((container, { wrapper, options, logger, meta }) => {
...
+         * If browser does not support IntersectionObserver, the dynamicImport function is called,
+         * adds the appropriate polyfill needed for legacy browser support, then executes containerHeightCheck.
+         * Script tag containing polyfill is subsequently removed after load.
+         *
+         * If browser support exists, containerHeightCheck runs as normal.
+         */
+
+        if (typeof window.IntersectionObserver === 'undefined') {
+            const polyfillUrl = 'https://polyfill.io/v3/polyfill.js?features=IntersectionObserver';
+            dynamicImport(polyfillUrl)
```
Seavenly: If we can move the `script` cleanup inside the helper utility, let's combine these into a single line.

yeul:
Adding some screenshots to show this in action:
1. Demonstrating a fallback message due to parent container width too small for full message:
Screen Shot 2020-05-26 at 1 33 37 PM
2. Message hidden due to parent container width too small for fallback:
Screen Shot 2020-05-26 at 1 34 00 PM
3. Message hidden due to parent container height too small for fallback:
Screen Shot 2020-05-26 at 1 34 28 PM
4. Message hidden due to parent container height too small for full message:
Screen Shot 2020-05-26 at 1 34 50 PM
5. IE11 dynamic polyfill working and script tag with polyfill subsequently removed:
Screen Shot 2020-05-26 at 1 43 28 PM
Screen Shot 2020-05-26 at 1 45 07 PM

src/messages/models/Container/setSize.js
```diff
@@ -215,35 +216,91 @@ export default curry((container, { wrapper, options, logger, meta }) => {
...
+             */
+            const root = Array.prototype.slice
+                .call(document.querySelectorAll('*'))
+                .filter(myEl => {
+                    return (
+                        myEl.getBoundingClientRect().height !== window.innerHeight &&
+                        !myEl.tagName.match(/(STYLE|SCRIPT|HEAD|LINK|META|TITLE)/g)
+                    );
+                })
+                .shift();
```
Seavenly: We should likely traverse up the DOM tree from the message wrapper to get the exact path against which we should check. And we can recursively check until the `el.nodeType !== Node.ELEMENT_NODE`. Also, I think we can rely on checking against the computed style height being `100%` instead of having the borwser compute the height.

src/messages/models/Container/setSize.js
```diff
@@ -215,35 +216,91 @@ export default curry((container, { wrapper, options, logger, meta }) => {
...
+            // eslint-disable-next-line compat/compat
+            const observer = new IntersectionObserver(
+                (entries, io) => {
```
Seavenly: Consider using a full variable name such as `observer`, especially when `io` could mean intersection observer or comonnly i/o.

src/messages/models/Container/setSize.js
```diff
@@ -215,35 +216,91 @@ export default curry((container, { wrapper, options, logger, meta }) => {
...
+                                )
+                            ) {
+                                logger.error({ name: ERRORS.MESSAGE_HIDDEN });
+                                logger.warn(
+                                    `Message hidden. PayPal Credit Message fallback requires minimum dimensions of ${minWidth}px x ${parseInt(
+                                        container.getAttribute('height'),
+                                        10
+                                    ) + 1}px. Your current container is ${parentContainerWidth}px x ${
+                                        entry.target.parentNode.getBoundingClientRect().height
+                                    }px. Message hidden.`
```
Seavenly: Should we just extract the expression into their own variables so that prettier will collapse the string to one line? They look to be common between both the `if` and `else` block so we could lift them above that.

src/messages/models/Container/setSize.js
```diff
@@ -215,35 +216,91 @@ export default curry((container, { wrapper, options, logger, meta }) => {
...
+                                );
+
+                                // Highest priority styles, will re-render from attribute observer
+                                objectEntries(minSizeOptions).forEach(([key, val]) => {
+                                    const attributeKey = `data-pp-${key.replace(/\./g, '-')}`;
+                                    wrapper.parentNode.setAttribute(attributeKey, val);
+                                });
+                            }
+                        }
+                        io.disconnect();
```
Seavenly: If multiple entries were to be passed in `disconnect()` would be called multiple times. Consider moving this outside of the `forEach`.

src/messages/models/Container/setSize.js
```diff
@@ -215,35 +216,91 @@ export default curry((container, { wrapper, options, logger, meta }) => {
...
+
+        const setupOverflowObserver = () => {
+            const el = wrapper.parentNode;
+            /**
+             * Assign root to the first element with a height *not* equal to the window height.
+             * In most cases, the bounding box (root) will be assigned to <html>, but this accounts for cases
+             * where <html> or <body> might have a set height, potentially hiding the message when it is out of the viewport.
+             */
+            const root = Array.prototype.slice
+                .call(document.querySelectorAll('*'))
```
Seavenly: Consider using `Array.from` instead of the slice/call technique. Note that `Array.from` would need to be imported from `core-js` since it's not supported in ie11.

Seavenly:
Closing in favor of changes incorporated in `zoid-message`

PR #185, opened 2020-04-06
test: stats payload tests

merlinstardust:


tests/functional/spec/payload.test.js
```diff
@@ -0,0 +1,244 @@
...
+        const requests = await runTest({
+            config,
+            testPage,
+            callback: async ({ bannerFrame, modalFrame }) => {
+                await clickBanner(bannerFrame);
+                await modalFrame.click(selectors.button.btn);
+            }
+        });
+
+        const request = requests.find(r => r.bdata.link && request.bdata.link.includes('Apply Now'));
```
gillycheesesteak: I think this should be `r.bdata.link.includes()`

tests/functional/jest.config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-    testMatch: ['<rootDir>/**/?(*.)test.js?(x)'],
+    testMatch: ['<rootDir>/**/?(*.)test.js?(x)', '<rootDir>/spec/payload.test.js'],
```
gillycheesesteak: Does the first match pattern not catch this file?

merlinstardust: For some reason, this was the only way I could catch this file. I feel like it should match the pattern of the first one but it skips it.

package.json
```diff
@@ -74,6 +74,8 @@
         "prettier": "^1.14.3",
         "prettier-eslint": "^9.0.1",
         "puppeteer": "^2.0.0",
+        "puppeteer-request-spy": "^1.3.1",
```
Seavenly: Do we need to include this obscure package? Or can we accomplish what we need with a few lines of code using the Puppeteer API?

merlinstardust: It might be possible but I'm not sure how. My research pointed me to this as the quickest way to see requests that are sent out.

Seavenly: It looks like we're trying to capture certain requests based on keyword matching. Could this similarly help to accomplish it? https://pptr.dev/#?product=Puppeteer&version=v2.1.1&show=api-event-request

merlinstardust: Fixed finally!

PR #186, opened 2020-04-10
feat: Add rendered content and size to payload tracking

merlinstardust:


src/messages/models/Container/stats.js
```diff
@@ -116,6 +156,7 @@ export default curry((container, { options: { amount, account, merchantId, place
         browser_width: window.innerWidth,
         browser_height: window.innerHeight,
         visible: isInViewport(container),
+        renderedMessage,
```
gillycheesesteak, line 159: The rendered sizes (small, large, etc) will need to be added to the `uuid` field in `sendBeacon.js` in order to be properly captured by downstream data processing. The additional information in here could make sense to add to calls to `/messagingLogger`, but for the beacon, we want to keep it as brief as possible, so probably just something like `SMALL,LARGE,LARGE` with a defined order of headline,sub-headline,disclaimer

src/messages/services/logger/sendBeacon.js
```diff
@@ -7,10 +7,16 @@ import { curry } from '../../../utils/index';
...
-export default curry(({ uuid, urls, messageRequestId }, payload, bannerHidden = false) => {
+export default curry(({ uuid: mainUuid, urls, messageRequestId }, payload, bannerHidden = false) => {
     const beacon = new window.Image();
 
     if (typeof payload === 'object') {
+        const { headline, subHeadline, disclaimer } = payload.renderedMessage;
+        const renderedMessageUuidString = payload.renderedMessage
+            ? `::message.size:${headline.size},${subHeadline.size},${disclaimer.size}`
+            : '';
+        const uuid = `${mainUuid}${renderedMessageUuidString}`;
```
Seavenly, line 18: @gillycheesesteak Is there any concern with this value flowing in with different values during different points of the render cycle?

PR #204, opened 2020-05-21
feat: add support for gpl uk text messages

yeul:


src/locale/GB/logos.js
```diff
@@ -0,0 +1,45 @@
+import { PayPalLogo, PPLogo, LOGO_COLOR } from '@paypal/sdk-logos';
+import { dom } from 'jsx-pragmatic';
+
+function gplPPLogo(logoColor) {
+    return PayPalLogo({ logoColor }).render(dom({ doc: document }));
+}
+
+function gplPPMonogram(logoColor) {
```
Seavenly: Consider adding the `.getAttribute('src')` inside these 2 helper functions since it's always being called directly after the function is called each time. Also, there is nothing specific to "gpl" in the functions so something like `getPPLogoBase64(logoColor)` could be more appropriate.

src/messages/models/Template/index.js
```diff
@@ -364,7 +364,11 @@ function createTemplateNode(options, markup) {
         span.textContent = `${withText} `;
         const strong = document.createElement('strong');
         strong.textContent = productName;
-        span.appendChild(strong);
+
+        if (getLocaleClass() === 'locale--GB') {
```
Seavenly: Should we alter the `strong` `createElement()` type based on the locale to be either a `strong` or `span` tag and change its name from `strong` to `productNode` or something along those lines?

yeul: Good idea, I've added that in. Let me know if that was what you were thinking.

banners/GB/pl.json
```diff
@@ -0,0 +1,18 @@
...
+        "variables": {
+            "offer_id": "${CREDIT_OFFERS_DS.offer_id}",
+            "vendor_financing_id": "${CREDIT_OFFERS_DS.vendor_financing_id}",
+            "financing_code": "${CREDIT_OFFERS_DS.financing_code}"
+        }
+    },
+
+    "headline": [["Make 3 monthly payments with 0% interest on eligible purchases.", ["default"]]],
+
+    "subHeadline": [["", ["default", "small"]]],
```
gillycheesesteak: Do we need to include this empty string or can this element just be `"subHeadline": []`?

src/messages/models/Template/index.js
```diff
@@ -353,20 +353,58 @@ function createTemplateNode(options, markup) {
...
+         * Used for stylizing the brand text for GPL UK inline messages.
+         */
+
+        if (insertPPText) {
+            const paypalNode = document.createElement('span');
+            paypalNode.className = 'pp-text-logo';
+            span.appendChild(paypalNode);
+        }
+
+        const productNode = document.createElement(`${el}`);
```
Seavenly:
```suggestion
        const productNode = document.createElement(nodeType);
```

src/messages/models/Template/index.js
```diff
@@ -353,20 +353,58 @@ function createTemplateNode(options, markup) {
...
+        span.textContent = `${withText} `;
+
+        let el = 'strong';
+        switch (getLocaleClass()) {
+            case 'locale--GB':
+                el = 'span';
+                break;
+            default:
+                break;
+        }
```
Seavenly: Consider simplifying this:
```suggestion
        const nodeType = getLocaleClass() === 'locale--GB' ? 'span' : 'strong';
```

src/locale/GB/mutations/pl.js
```diff
@@ -0,0 +1,127 @@
...
+
+export default {
+    'layout:text': [
+        [
+            'default',
+            ({ textSize }) => ({
+                styles: [
+                    fallbackMediaQuery(textSize * 13),
+                    `.message__logo { width: ${textSize * 4}px }`,
+                    `.message__logo:first-child { width: ${textSize * 1.25}px; margin-right: 5px; }`
```
gillycheesesteak: I believe the `margin-right: 5px` can be removed, as it is replicated in CSS

src/locale/GB/mutations/pl.js
```diff
@@ -0,0 +1,127 @@
...
+                logo: Logo.PRIMARY.COLOR,
+                headline: [
+                    {
+                        tag: 'medium',
+                        replace: [['purchases.', 'purchases']],
+                        br: ['on']
+                    },
+                    { tag: 'xsmall' }
+                ],
+                disclaimer: ['default']
```
gillycheesesteak: This line should be unnecessary because it matches the `default` config. Same for other variations

src/locale/GB/mutations/plq.js
```diff
@@ -0,0 +1,141 @@
...
+                    `.message__logo:first-child { width: ${textSize * 1.25}px; }`
+                ],
+                logo: Logo.PRIMARY.COLOR,
+                headline: [
+                    {
+                        tag: 'medium',
+                        replace: [['month.', 'month']],
+                        br: ['of']
+                    },
+                    { tag: 'xsmall' }
```
gillycheesesteak: Since this exact config is repeated for many variations, consider making it the default and overriding with the current default where necessary

src/messages/models/Template/index.js
```diff
@@ -353,20 +353,51 @@ function createTemplateNode(options, markup) {
     prependText(disclaimer, toMarkup('disclaimer', mutationRules.disclaimer));
     appendImage(logoContainer, mutationRules.logo, 'PayPal Credit logo');
 
+    // Appends the message with the word "with" along with the product name. i.e. "with PayPal Credit"
+    const appendWithProductText = (insertPPText = false) => {
+        const [withText, productName] = getLocalProductName();
+        const span = document.createElement('span');
+        span.textContent = `${withText} `;
+
+        const nodeType = getLocaleClass() === 'locale--GB' ? 'span' : 'strong';
```
gillycheesesteak: Can we always use a span and add styles to US/DE to make it bold? Or, alternatively, always use `strong` and add styles in GB to make it normal weight?

src/messages/models/Template/index.js
```diff
@@ -353,20 +353,51 @@ function createTemplateNode(options, markup) {
...
-        headline.appendChild(span);
+        appendWithProductText();
+    }
+
+    if (
+        (objectGet(options, 'style.logo.type') === 'primary' ||
+            objectGet(options, 'style.logo.type') === 'alternative') &&
+        getLocaleClass() === 'locale--GB'
+    ) {
+        appendWithProductText();
```
gillycheesesteak: Consider simplifying this to
```
if (getLocaleClass() === 'locale--GB') {
   appendWithProductText(objectGet(options, 'style.logo.type') === 'inline');
} else if (objectGet(options, 'style.logo.type') === 'none') {
   appendWithProductText();
}

File: tests/functional/spec/GB/text.test.js [hunk 11]
```diff
@@ -0,0 +1,79 @@
...
+            const getConfig = style => ({
+                account,
+                style: {
+                    layout: 'text',
+                    ...style
+                }
+            });
+
+            // Each valid logo type
+            ['primary', 'alternative', 'inline', 'none'].forEach(type => {
```
gillycheesesteak: Let's test `top`, `left`, and `right` for logo type === `primary`

utils/devServerProxy.js
```diff
@@ -67,11 +70,19 @@ module.exports = app => {
...
+                formattedMonthlyPayment = `£${bestOffer.monthly}`;
+            }
+
             const morsVars = {
                 financing_code: Math.random()
                     .toString(36)
                     .slice(2),
-                formattedMonthlyPayment: country === 'DE' ? `${bestOffer.monthly}€` : `$${bestOffer.monthly}`,
+                formattedMonthlyPayment,
                 formattedTotalCost: country === 'DE' ? `${terms.formattedAmount}€` : `$${terms.formattedAmount}`,
```
gillycheesesteak: Let's format this properly based on country too, even if we aren't using it yet. Maybe have a small utility for localizing a currency amount

utils/devServerProxy.js
```diff
@@ -67,12 +70,23 @@ module.exports = app => {
...
 
+            let formattedMonthlyPayment = `$${bestOffer.monthly}`;
+            let formattedTotalCost = `$${terms.formattedAmount}`;
+
+            if (country === 'DE') {
+                formattedMonthlyPayment = `${bestOffer.monthly}€`;
+                formattedTotalCost = `${terms.formattedAmount}€`;
+            } else if (country === 'GB') {
+                formattedMonthlyPayment = `£${bestOffer.monthly}`;
+                formattedTotalCost = `£${terms.formattedAmount}`;
```
gillycheesesteak: I think this could be cleaner with a separate function `localizeCurrency(country, amount)` that returns the correct format, and just call that for both values once

tests/functional/spec/GB/acounts.js
```diff
@@ -0,0 +1 @@
+export default ['DEV000000GBPL', 'DEV00000GBPLQ'];
```
gillycheesesteak: Filename is spelled wrong :)

utils/devServerProxy.js
```diff
@@ -116,12 +130,20 @@ module.exports = app => {
             const terms = getTerms(country, Number(amount));
             const [bestOffer] = terms.offers || [{}];
 
+            let formattedMonthlyPayment = `$${bestOffer.monthly}`;
+
+            if (country === 'DE') {
+                formattedMonthlyPayment = `${bestOffer.monthly}€`;
+            } else if (country === 'GB') {
+                formattedMonthlyPayment = `£${bestOffer.monthly}`;
+            }
```
gillycheesesteak: Formatting utility described above could be used here as well

tests/functional/spec/GB/acounts.js
yeul: Oopsie daisy 😅

PR #206, opened 2020-06-11
refactor: consolidate modal requests

gillycheesesteak:


src/modal/parts/Calculator/DE/Calculator.jsx
```diff
@@ -13,14 +14,20 @@ const Calculator = () => {
             <div className="calculator">
                 <h3 className="calculator__title">Monatliche Raten berechnen</h3>
                 <form className="calculator__form" onSubmit={submit}>
-                    <input className="calculator__input" value={value} onInput={changeInput} />
+                    <input
+                        className="calculator__input"
+                        value={value}
+                        onInput={changeInput}
+                        disabled={isLoading}
```
Seavenly: Can this be removed for accessibility?

src/modal/parts/Calculator/DE/Calculator.jsx
```diff
@@ -13,14 +14,20 @@ const Calculator = () => {
...
+                        style={{ opacity: isLoading ? '0.5' : '1' }}
+                    />
                     <p className="calculator__instructions">
                         {!terms.error &&
                             terms.formattedMinAmount &&
                             terms.formattedMaxAmount &&
                             `Geben Sie einen Betrag zwischen ${terms.formattedMinAmount}€ und ${terms.formattedMaxAmount}€ ein.`}
                     </p>
-                    <Button size="md" type="submit">
+                    <Button size="md" type="submit" disabled={isLoading}>
```
Seavenly: Same as above.

src/modal/parts/Calculator/US/Calculator.jsx
```diff
@@ -13,8 +13,15 @@ const Calculator = () => {
...
                 <h2 className="calculator__title">Enter a purchase amount to calculate your monthly Easy Payments.</h2>
                 <form className="calculator__form" onSubmit={submit}>
-                    <input className="calculator__input" type="text" value={value} onInput={changeInput} />
-                    <Button className="calculator__button" type="submit" secondary>
+                    <input
+                        className="calculator__input"
+                        type="text"
+                        value={value}
+                        onInput={changeInput}
+                        disabled={isLoading}
```
Seavenly: Same as above.

src/modal/lib/hooks/calculator.js
```diff
@@ -1,10 +1,35 @@
...
 import { useXProps } from './helpers';
 import { ServerContext } from '../context';
 import { request } from '../../../utils';
 
+const termsCache = new Map();
+
+const getCacheKey = params =>
+    ['amount', 'country', 'client_id', 'payer_id', 'merchant_id'].map(key => params[key]).join('_');
+
+const termsFetcher = async (params, csrf) => {
```
Seavenly: Could we use the existing `memoize` helper function to achieve the same thing as creating another type of cache here? Or we could possibly refactor a bit and use `useMemo` to store the request promise and read it when needed.

gillycheesesteak: I couldn't think of a way to use the existing memoize while also using the terms that are server-rendered and having them be cached as well. Thoughts?

PR #207, opened 2020-06-11
fix: remove modal double scrollbar

gillycheesesteak:


src/messages/components/Modal/index.js
```diff
@@ -14,8 +14,63 @@ function getModalType(offerCountry, offerType) {
...
+            const placeholderViewport = document.createElement('meta');
+            placeholderViewport.setAttribute('name', 'viewport');
+            placeholderViewport.setAttribute('content', '');
+
+            if (!existingViewport) {
+                // If no viewport exists, some browsers will not respect the change of simply removing the hijack viewport
+                // Add a blank viewport to replace the hijack in this case
+
+                document.head.appendChild(placeholderViewport);
+            }
```
Seavenly: Does `appendChild` force an immediate browser repaint/cycle? I would think that appending a child and then immediately removing the child could be squashed into no action by the browser.

src/messages/components/Modal/index.js
```diff
@@ -14,8 +14,63 @@ function getModalType(offerCountry, offerType) {
...
+const useViewportHijack = () => {
+    const [viewportState, setViewportState] = createState({});
+
+    return [
+        () => {
+            // Save existing body style
+            const bodyStyle = document.body.getAttribute('style');
+
+            // Create hijack viewport
+            const newViewport = document.createElement('meta');
```
Seavenly: Instead of manually creating the meta elements, could we use the `jsx-pragmatic` to create them?

PR #220, opened 2020-06-30
feat: add black message variants for US PPC

yeul:


src/locale/US/mutations/logoUtil.js
```diff
@@ -0,0 +1,26 @@
+import Logo from '../logos';
+
+/**
+ * Assigns appropriate logo based on message config options.
+ * Used in mutations files where config settings are the same across many accounts.
+ */
+
+export const textLogosArr = [
```
Seavenly: Consider renaming this file to something like `common.js` since this file doesn't contain actionable utility functions. Also consider using more specific variable names such as `textLogoMutations` and `flexLogoMutations`.

Seavenly (approved):

PR #228, opened 2020-07-14
refactor: change countryCode to buyerCountry

yeul:


src/messages/models/Banner/validateOptions.js
```diff
@@ -205,6 +206,14 @@ export default curry((logger, { account, amount, style, offer, ...otherOptions }
         validOptions.style = { layout: 'text' };
     }
 
+    if (typeof buyerCountry !== 'undefined' && buyerCountry !== 'US') {
+        if (typeof amount !== 'undefined') {
+            logger.warn(
+                'When using a non-US buyerCountry override with an amount, please add corresponding currency code to your config.'
```
gillycheesesteak, line 212: Should we check whether a currency has been already configured before showing this warning?

src/utils/elements.js
```diff
@@ -26,6 +26,7 @@ export function getInlineOptions(container) {
         .filter(({ nodeName }) => stringStartsWith(nodeName, 'data-pp-'))
         .reduce((accumulator, { nodeName, nodeValue }) => {
             if (nodeValue) {
+                if (nodeName === 'data-pp-buyercountry') nodeName = 'data-pp-buyerCountry'; // eslint-disable-line no-param-reassign
```
gillycheesesteak: Can we generalize this as an "attribute name override" map in case we have other properties in the future where we want this kind of behavior?

src/messages/models/Banner/validateOptions.js
```diff
@@ -19,7 +19,7 @@ const VALID_OPTIONS = {
     onApply: [Types.FUNCTION],
     currency: [Types.STRING, ['USD', 'EUR']],
     placement: [Types.STRING, ['', 'home', 'category', 'product', 'cart', 'payment']],
-    countryCode: [Types.STRING, [undefined, 'US', 'DE']]
+    buyerCountry: [Types.STRING, [undefined, 'US', 'DE']]
```
gillycheesesteak: Should we add GB to this list? Are there other buyer countries that we want to allow for US-based merchants? Such as MX, CA?

yeul: I've gone ahead and added GB to the list since we will soon have GB messaging available in prod. Adding something like MX or CA wouldn't actually work because we don't have messaging for those countries. Technically this is for merchants who aren't located in the country where the majority of their customers are. It would make sense for US, DE, and GB because a Mexican merchant might have most of their customers in the US so they want to pass in `buyerCountry: 'US'` since that's where their customers are. Same for DE and GB. A merchant might be based in Switzerland or Austria with most of their customers from Germany so it would make sense to pass in buyerCountry: 'DE'. Something like MX or CA would only work if we had MX or CA specific messages. But it's definitely something we will probably do in the future.

PR #235, opened 2020-07-15
fix: correct validation for font sizes

merlinstardust:


demo/text.html
```diff
@@ -45,7 +46,9 @@
                     'DEV0000000PSZ',
                     'DEV0000000PSG',
                     'DEV0000000PMZ',
-                    'DEV0000000PMG',
+                    'DEV0000000PMG'
+
+                    // DE
```
yeul, line 51: How about we put the GB accounts ( `DEV000000GBPL` and `DEV00000GBPLQ`) under a `// GB` comment.

src/messages/models/Banner/validateOptions.js
```diff
@@ -94,6 +94,7 @@ function getValidVal(logger, typeArr, val, location) {
 function populateDefaults(logger, defaults, options, prefix = 'style.') {
     return objectEntries(defaults).reduce((accumulator, [key, val]) => {
         if (Array.isArray(val)) {
+            // console.log({key, val, oKey: options[key],  pfKey: `${prefix}${key}`})
```
yeul: Is this leftover from your testing? If so, can it be removed?

yeul (approved):

PR #240, opened 2020-07-24
fix: preselecting tabs based on offer type of banner

paypal-are:
Allows appropriate tab to be pre-selected based on the offer type of the banner presented to the user.

### DIFF
--- a/src/messages/components/Modal/component.js
+++ b/src/messages/components/Modal/component.js
@@ -58,11 +58,6 @@ export default getGlobalComponent('__paypal_credit_modal__', () =>
                 sendToChild: true,
                 required: false
             },
-            messageOfferType: {
-                type: 'string',
-                queryParam: true,
-                required: true
-            },
             country: {
                 type: 'string',
                 queryParam: true,
@@ -85,11 +80,6 @@ export default getGlobalComponent('__paypal_credit_modal__', () =>
             },
 
             // Callbacks
-            track: {
-                type: 'function',
-                queryParam: false,
-                required: false
-            },
             onClick: {
                 type: 'function',
                 queryParam: false,
@@ -105,11 +95,15 @@ export default getGlobalComponent('__paypal_credit_modal__', () =>
                 queryParam: false,
                 required: false
             },
-
+            onReady: {
+                type: 'function',
+                queryParam: false,
+                required: false
+            },
             // Computed Props
             offer: {
                 type: 'string',
-                value: ({ props }) => determineInitialTab(props.messageOfferType),
+                value: ({ props }) => determineInitialTab(props.offer),
                 required: false
             },
             payerId: {
--- a/src/messages/components/Modal/component.js
+++ b/src/messages/components/Modal/component.js
@@ -17,7 +17,21 @@ const getGlobalComponent = (namespace, fn) => {
 
 // Determine pre-selected tab based on the offer type of the banner.
 // Currently only applicable to the US
-const determineInitialTab = (type = 'NI') => (startsWith(type, 'EZP') || startsWith(type, 'PALA') ? 'EZP' : 'NI');
+const determineInitialTab = (type = 'NI') => {
+    switch (true) {
+        case [
+            'EZP:ANY:EQZ',
+            'EZP:ANY:GTZ',
+            'PALA:MULTI:EQZ',
+            'PALA:MULTI:GTZ',
+            'PALA:SINGLE:EQZ',
+            'PALA:SINGLE:GTZ'
+        ].includes(type):
+            return 'EZP';
+        default:
+            return 'NI';
+    }
+};
 
 export default getGlobalComponent('__paypal_credit_modal__', () =>
     create({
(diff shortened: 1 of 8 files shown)


src/modal/content/Content.jsx
```diff
@@ -9,10 +9,56 @@ import INST from './DE/INST';
 import PL from './GB/PL';
 import Tabs from '../parts/Tabs';
 
-const Content = () => {
+// Utilize a combination of offer type and country code to determine pre-selected tab.
+const determineInitialTab = (type = 'NI', country = 'US') => {
+    return {
```
gillycheesesteak: I think we can simplify this by using a switch on country. For DE and GB, since they are always the same, we can just return the value. For US, we can just use an array of EZP offer types and check whether the type passed in is in that array

src/modal/Modal.jsx
```diff
@@ -16,7 +16,7 @@ const Modal = ({ serverData }) => (
                 {common}
             </style>
             <Container>
-                <Content />
+                <Content modalType={serverData.type} country={serverData.country} />
```
gillycheesesteak: Instead of passing this info in as props, we can use the `useServerData` hook within the Content component to extract these values

src/modal/parts/Tabs.jsx
```diff
@@ -4,14 +4,27 @@ import { useState, useEffect } from 'preact/hooks';
 
 import { useTransitionState, useXProps } from '../lib/hooks';
 
-const Tabs = ({ tabs }) => {
-    const [currentTab, selectTab] = useState(0);
+const getInitialTabIndex = (initialTabKey, tabs) => {
+    let selected = 0;
+    tabs.forEach(({ tabKey }, index) => {
```
gillycheesesteak: Consider using `tabs.findIndex()`

src/modal/content/Content.jsx
```diff
@@ -29,13 +54,16 @@ const Content = () => {
                         {USEzp}
                     </style>
                     <Tabs
+                        initialTabKey={determineInitialTab(type, country)}
```
Seavenly: I think we could likely leverage passing in the `offer` messages option into the modal as a prop to simplify this, especially since the tabs are only applicable for US.

src/messages/components/Modal/index.js
```diff
@@ -84,7 +84,9 @@ export default {
             });
         } else {
             const modalType = getModalType(meta.offerCountry, meta.offerType);
```
gillycheesesteak: We should be able to remove this variable here, and move the tracking call below so it can make use of the correct modal type

src/messages/components/Modal/index.js
```diff
@@ -84,7 +84,9 @@ export default {
...
-            const { renderProm, show } = renderModal({ options, meta, track, wrapper, type: modalType });
+            // The type passed in here is the offer type from the messages call.
+            // The modal type is returned from a separate call and passed in as a prop
+            const { renderProm, show } = renderModal({ options, meta, track, wrapper, type: meta.offerType });
```
gillycheesesteak: Let's rename the `type` variable to something more specific, like `messageOfferType`

src/messages/components/Modal/component.js
```diff
@@ -67,6 +85,11 @@ export default getGlobalComponent('__paypal_credit_modal__', () =>
             },
 
             // Callbacks
+            track: {
```
Seavenly: Instead of passing the tracker into the modal, can we follow similar to the changes in our `zoid-message` feature branch?
[https://github.com/paypal/paypal-messaging-components/blob/87d1e7abc16280edc617e6dc9671a5bf413d9bda/src/zoid/modal/component.js#L72](https://github.com/paypal/paypal-messaging-components/blob/87d1e7abc16280edc617e6dc9671a5bf413d9bda/src/zoid/modal/component.js#L72)
https://github.com/paypal/paypal-messaging-components/blob/87d1e7abc16280edc617e6dc9671a5bf413d9bda/src/controllers/modal/interface.js#L11

src/messages/components/Modal/component.js
```diff
@@ -84,6 +107,11 @@ export default getGlobalComponent('__paypal_credit_modal__', () =>
             },
 
             // Computed Props
+            offer: {
+                type: 'string',
+                value: ({ props }) => determineInitialTab(props.messageOfferType),
```
Seavenly: Does `messageOfferType` get consumed anywhere else besides here? Or could we simplify this by doing something like `({ props }) => determineInitialTab(props.offer)` where it computes itself?

src/messages/components/Modal/index.js
```diff
@@ -5,25 +5,15 @@ import { memoizeOnProps } from '../../../utils';
...
-}
-
 const renderModal = memoizeOnProps(
-    ({ options, meta, track, wrapper, type }) => {
+    ({ options, meta, track, wrapper }) => {
         const [hijackViewport, replaceViewport] = useViewportHijack();
 
-        const { render, hide, updateProps } = Modal({
-            type,
+        const createOnReadyHandler = () => ({ modalType }) =>
```
gillycheesesteak: Is there a reason we don't just create the handler inline here?

paypal-are: None that I know of. It's consistent with [this code](https://github.com/paypal/paypal-messaging-components/blob/87d1e7abc16280edc617e6dc9671a5bf413d9bda/src/controllers/modal/interface.js#L11).

I can make that change.

PR #242, opened 2020-07-27
fix: apply now link consistent with sdk environment

paypal-are:


yeul (approved):

src/modal/lib/hooks/applyNow.js
```diff
@@ -6,14 +6,17 @@ import { ServerContext } from '../context';
 
 export default clickTitle => {
     const { payerId } = useContext(ServerContext);
-    const { onClick, refId } = useXProps();
+    const { onClick, refId, env } = useXProps();
+
     const [, handleClose] = useTransitionState();
 
+    const urlBase = env === 'production' ? 'paypal' : 'sandbox.paypal';
```
Seavenly: Can we also have a 3rd case for local and stage to use the stage URL?
- production: `paypal`
- sandbox: `sandbox.paypal`
- local or stage (maybe staging?): `msmaster.qa.paypal`

paypal-are: Sure thing. Would the `local` environment point to a value taken from `globals.js`?

PR #244, opened 2020-08-03
test: zoid message unit tests

Seavenly:


src/controllers/message/interface.js
```diff
@@ -21,14 +20,26 @@ export default options => ({
         const { messagesMap } = globalState;
         const containers = getAllBySelector(selector);
 
-        if (selector.length === 0) {
-            return Logger.warn('Invalid selector', selector);
+        if (containers.length === 0) {
+            // Prevent firing off extra log when loaded inside zoid iframe
+            if (selector !== '[data-pp-message]') {
```
gillycheesesteak: Do we want to use a different method to determine whether this is loaded inside the zoid iframe? For example, if a merchant uses the default selector, but does not include an element on the page, this warning may be useful

PR #249, opened 2020-08-10
refactor: isolate localized modal content

gillycheesesteak:


src/components/modal/parts/Container.jsx
```diff
@@ -36,48 +18,18 @@ const Modal = ({ children }) => {
     }, [transitionState]);
 
     useLayoutEffect(() => {
-        if (transitionState === 'CLOSED') {
+        if (transitionState === 'CLOSING') {
```
Seavenly: Does changing this to `CLOSING` have the potential to cause the content to scroll up while animating down causing a bit of a visual jank?

webpack.config.dev.js
```diff
@@ -81,10 +81,15 @@ module.exports = (env = {}) => {
         vars: globals(env)
     });
 
-    COMPONENTS_DEV_CONFIG.entry = {
-        'smart-credit-message': './src/components/message/index.js',
-        'smart-credit-modal': './src/components/modal/index.js'
-    };
+    COMPONENTS_DEV_CONFIG.entry = ['US', 'DE', 'GB'].reduce(
```
Seavenly, line 86: Do we need a similar change to `webpack.config.js`?

src/components/modal/parts/Container.jsx
gillycheesesteak: The browser seems to not like updating the scroll position of the element when it is already hidden. Based on my testing, this is currently broken in production. I haven't noticed any jank with this change testing locally

PR #254, opened 2020-08-17
feat: add support for server-side rendered custom LE banners

yeul:


server/message/index.jsx
```diff
@@ -98,13 +104,91 @@ export default ({ options, markup, locale }) => {
         </span>
     );
 
+    const injectSpaceNodes = spans => {
+        return spans.reduce((accumulator, span) => [...accumulator, span, ' '], []).slice(0, -1);
+    };
+
+    const CustomSpan = ({ textContent, className = '' }) => <span className={className}>{textContent}</span>;
```
gillycheesesteak: I think we can use the `Text` part that already exists for this

server/message/index.jsx
```diff
@@ -98,13 +104,91 @@ export default ({ options, markup, locale }) => {
...
+                ''
+            );
+        });
+        return populatedMarkup;
+    };
+
+    createCustomTemplateNode({ data: markup, meta: markup.meta, template: options.customMarkup });
+
+    // Shared mutations, styles, and fonts between custom and non-custom messages/banners.
+    const Styles = () => {
```
gillycheesesteak: Can we split this out into its own file in `parts`?

server/message/index.jsx
```diff
@@ -98,13 +104,91 @@ export default ({ options, markup, locale }) => {
...
+
+        const spans = uniformText.map(text => {
+            if (Array.isArray(text)) {
+                const [textContent, className] = text;
+                return <CustomSpan className={className} textContent={textContent} />;
+            }
+            return <CustomSpan textContent={text} />;
+        });
+
+        return injectSpaceNodes(spans);
```
gillycheesesteak: Instead of using `injectSpaceNodes`, we can use the `spaced` prop of `Text` to add a space after the span

server/message/index.jsx
```diff
@@ -98,13 +104,91 @@ export default ({ options, markup, locale }) => {
...
+                <style
+                    className="styles__mutations"
+                    dangerouslySetInnerHTML={{ __html: mutationStyleRules.join('\n') }}
+                />
+                <style className="styles__misc" dangerouslySetInnerHTML={{ __html: miscStyleRules.join('\n') }} />
+            </>
+        );
+    };
+
+    if (options.style.layout === 'custom') {
```
gillycheesesteak, line 99: Can we create a `CustomMessage` part that includes any of the functions that are only related to custom messages?

PR #256, opened 2020-08-18
test: Improve functional tests

merlinstardust:


package.json
```diff
@@ -32,7 +32,8 @@
         "preinstall": "rm -rf node_modules",
         "start": "npm run dev",
         "test": "jest",
-        "test:func": "jest --projects tests/functional",
+        "test:func": "jest --verbose --config tests/functional/jest.config.js",
+        "test:func:payload": "jest --verbose --config tests/functional/jest.payload.config.js",
```
gillycheesesteak, line 38: What purpose does separating the payload tests serve?

scripts/snapshot-tests/run-tests.sh
```diff
@@ -1,17 +1,25 @@
...
-    git commit -m "chore(snapshots): update snapshots [skip ci]"
-    git push
-fi
\ No newline at end of file
+        git checkout develop
+        git add .
+        git commit -m "chore(snapshots): update snapshots [skip ci]"
+        git push
+    else
+        npm run test:func:payload
```
gillycheesesteak: Does this mean we are only running the payload tests when the `DIRTY_SNAPSHOTS` flag is set? Is that what we want?

tests/functional/spec/payload.test.js
```diff
@@ -65,6 +65,7 @@ describe('payload testing', () => {
     const testPage = 'banner.html';
 
     test('initial payload', async () => {
+        console.log('initial payload'); // eslint-disable-line no-console
```
gillycheesesteak: Can we add an `.eslintrc` (or edit the existing one, if there is one) for just the `tests` folder that disables this eslint rule for everything in the folder?

tests/functional/spec/createBannerTest.js
```diff
@@ -36,6 +36,13 @@ const getTestNameParts = (locale, { account, style: { layout, ...style } }) => {
 
 const waitForBanner = async timeout => {
     try {
+        // console.log('waitFor load');
+        // await page.waitForNavigation({ waitUntil: 'load' }); // consider nav done when the load event is fired
+        // console.log('waitFor domcontentloaded');
+        // await page.waitForNavigation({ waitUntil: 'domcontentloaded' }); // consider nav done when the DOMContentLoaded event is fired
```
gillycheesesteak: Do we still need this commented code?

gillycheesesteak (commented):
Do these changes address the issue where the uploaded album has fewer images than there are failed snapshots?

package.json
merlinstardust: The main purpose of this was to allow for payload testing even when snapshots need to be updated and it's not on the develop branch. The current script will not run any functional tests if `DIRTY_SNAPSHOTS` is set and it's not a dev branch. Right now it's just payload tests, but I could see other non-snapshot functional tests being added to this in the future.

tests/functional/spec/payload.test.js
merlinstardust: I added these to ensure Travis doesn't time out since these tests can take a bit sometimes. I can add an .eslintrc to the tests folder for this, but then it won't catch any extra console.logs that might be added during local debugging and could lead to clutter. Thoughts?

gillycheesesteak: My personal feeling is that console.logs in test files are fine, so I don't mind disabling that rule for the tests file, but I'm open to leaving it like this because your explanation makes sense

tests/functional/spec/US/modal/us_modalTestDefs.js
```diff
@@ -57,15 +57,15 @@ export const applyNowBtn = (account, viewport, bannerStyle) => async () => {
...
     await page.waitFor(800);
 
-    await modalSnapshot(`${testNameParts} ${bannerStyle.layout}`, viewport, account);
+    await modalSnapshot(`${groupString} ${testNameParts}`, viewport, account);
 };
 
 /**
  * Runs inside modalCalc.test.js for the US locale.
- * Passes in bannerStyles instead of bannerStyle.
+ * Passes in groupString instead of groupString.
```
yeul: We could probably get rid of this comment here since it seems redundant now.

tests/functional/spec/utils/testStylesConfig.js
```diff
@@ -21,8 +21,16 @@ const bannerStyles = [
 
 const amounts = [1, 10000, undefined, 500];
 
+const getGroupString = ({ viewport, bannerStyle, amount }) => {
```
yeul: Let's add a comment above this function explaining what this is trying to do. Since if you open up test files `getGroupString` might be kind of vague.

yeul:
can you add the "test:" prefix to the front of your PR title?

yeul:
It seems we still get some diff-ing failures from certain calculator tests because either the amount isn't completely entered in or the logo is at a slightly different height between the two versions. i.e.:

Screen Shot 2020-08-31 at 5 22 27 PM
Screen Shot 2020-08-31 at 5 22 34 PM

We should probably increase the delay before the snapshot is taken here since Puppeteer is super finicky and inconsistent in the timing area.

PR #273, opened 2020-09-08
fix: allow for inline text size attribute

yeul:


src/old/messages/models/Banner/validateOptions.js
```diff
@@ -195,6 +195,10 @@ export default curry((logger, { account, amount, style, offer, ...otherOptions }
 
     if (validateType(Types.OBJECT, style) && validateType(Types.STRING, style.layout)) {
         validOptions.style = style;
+        const sizeVal = validOptions.style?.text?.size;
+        if (typeof sizeVal === 'string') {
```
Seavenly: Can we do similar checks like `amount` above so that we can filter out coerced `NaN` values?

PR #275, opened 2020-09-08
feat: add Pay in 4 text messages

yeul:


server/locale/US/PPC/validOptions.js
```diff
@@ -1,4 +1,4 @@
-import { Types } from '../../types';
```
merlinstardust, line 1: Do we have root imports? Might be a good a time to add them. `~/server/types` would be a lot nicer.

server/locale/US/Pi4/logos.js
```diff
@@ -0,0 +1,109 @@
...
+
+function getPPMonogramBase64(logoColor) {
+    if (logoColor === 'monochrome') {
+        return getSrc(PPMonochrome({ logoColor }));
+    }
+    return getSrc(PPLogo({ logoColor }));
+}
+
+export default {
+    PRIMARY: {
```
merlinstardust, line 23: Can these be named similarly to what I have on updates with `STACKED`, `SINGLE_LINE`, and so on? Would be clearer to know which is which.

If the `PRIMARY` or similar label is absolutely necessary for something (like custom banners), it could be added as another key-value of `PRIMARY: STACKED` or what have you

server/locale/US/Pi4/mutations/index.js
```diff
@@ -0,0 +1,17 @@
...
+import pi4_no_amount from './pi4_no_amount';
+import pi4_non_qual from './pi4_non_qual';
+import pi4q from './pi4q';
+
+export default function getMutations(id, type, { amount }) {
+    switch (id) {
+        case 'PI4Q':
+            return pi4q[type];
+        case 'PI4':
+            return typeof amount === 'undefined' ? pi4_no_amount[type] : pi4_non_qual[type];
```
merlinstardust, line 17: Are these different enough to warrant separate files? Looking at the diff between them, seems like a lot is extremely similar.

But if it does warrant separate files for mutations, I feel like there should be a corresponding banner

server/locale/US/Pi4/mutations/index.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable eslint-comments/disable-enable-pair */
+// mutations import here.
+/* eslint-disable camelcase */
+import pi4_no_amount from './pi4_no_amount';
+import pi4_non_qual from './pi4_non_qual';
+import pi4q from './pi4q';
```
merlinstardust, line 6: I think these file name should align with the `banners/...` file names. That way it's clear which banner/info goes with which mutations

server/locale/US/Pi4/mutations/mediaQueries.js
```diff
@@ -0,0 +1,103 @@
+/**
+ * Wrap text under logo for default/primary left and alternative config types.
+ */
+export function textWrap(breakpoint, textSize) {
+    return `@media screen and (max-width: ${breakpoint}px) {
```
merlinstardust: Is it at all possible to invert this (and the other queries) and the corresponding styles so `min-width` is used and it's a mobile first approach?

We use mobile first in almost every other media query except for the messageDisclaimer one and that's at a super small breakpoint too. So if possible, use min-width / mobile-first instead of max-width

server/locale/US/Pi4/mutations/mediaQueries.js
```diff
@@ -0,0 +1,103 @@
...
+    .message__headline > .tag--medium > span {
+        white-space: normal;
+    }
+
+    .message__headline > .tag--xsmall {
+        display: none;
+    }
+
+    @media screen and (max-width: ${breakpoint}px) {
+
```
merlinstardust: Minor thing but there's an extra newline here. To be consistent, there should be either no extra newlines or one at each end of the media block

server/locale/US/Pi4/mutations/mediaQueries.js
```diff
@@ -0,0 +1,103 @@
...
+    .message__headline > .tag--medium > span {
+        white-space: normal;
+    }
+
+    .message__headline > .tag--small {
+        display: none;
+    }
+
+    @media screen and (max-width: ${breakpoint}px) {
+
```
merlinstardust: Also extra newline here

server/locale/US/Pi4/styles/text/base.css
```diff
@@ -0,0 +1,11 @@
+.message__disclaimer > .tag--default > span {
```
merlinstardust: Are these styles not being inherited from the existing base css that I added in `message-updates` branch?

server/locale/index.js
```diff
@@ -2,42 +2,42 @@ import US from './US';
 import DE from './DE';
 import GB from './GB';
 
-const getLocaleSettings = offerCountry => {
+const getLocaleSettings = (offerCountry, id) => {
```
merlinstardust: What goes in for id? Can this be clarified with a comment and example or use a different word? 

I've started to realize that ID is actually meaningless since literally everything is an identifier

merlinstardust: Looking below, it looks like ID is offerType, so I think you should just use that

tests/functional/spec/US/text.test.js
```diff
@@ -10,118 +10,127 @@ describe('US > text', () => {
...
-                    viewport,
-                    getConfig({
-                        logo: {
-                            type
-                        }
-                    })
-                );
-            });
+                // Each font size option logo.type-alternative_logo.position-left
+                [10, 12, 16].forEach(size => {
```
merlinstardust, line 51: Do we not want to do 11, 13, 14, and 15 too?

tests/functional/spec/US/text.test.js
```diff
@@ -10,118 +10,127 @@ describe('US > text', () => {
...
     accounts.forEach(account => {
-        describe(account, () => {
-            const getConfig = style => ({
-                account,
-                style: {
-                    layout: 'text',
-                    ...style
-                }
-            });
+        const amounts = [undefined, 5];
```
merlinstardust: This array of amounts should be defined closer to where it's used.

tests/functional/spec/US/text.test.js
```diff
@@ -10,118 +10,127 @@ describe('US > text', () => {
...
+        const accountTest = amount =>
```
merlinstardust, line 13: This seems weird to defined this long function inside of the `accounts.forEach` block. Any way to define it before the block so you can just do the below or similar.

```
accounts.forEach(account => {
  amounts.forEach(amount => {
    if (account === 'DEV0000000PI4') accountTest(account, amount)
    else accountTest(account, amount)
  }
}
```

Other option would be to change getConfig to just have a conditional in the `getConfig` function:

```
const getConfig = style => {
    const config = {
        account,
        style: {
            layout: 'text',
            ...style
        }
    };
    if (account === 'DEV0000000PI4') {
        config.amount = 5;
    }
    return config;
);
```

utils/devServerProxy/miscellaneous.js
```diff
@@ -20,8 +20,9 @@ export const localizeNumber = country => (amount, fractionDigits = 2) => {
 
     switch (country) {
         case 'DE':
-        case 'GB':
             return baseFormat.replace(/^([\d,]+)(\.)(\d+)$/, (match, p1, p2, p3) => `${p1.replace(/,/g, '.')},${p3}`);
```
merlinstardust, line 23: Why did this line get moved up to DE?

merlinstardust (changes requested):
One other thing I noticed, and I'll send you a screenshot, is that when I put in an amount for PI4Q, it put a weird bracket in the calculated amount. Also the amount seems to be dividing by 12 when it should be by 4.

utils/devServerProxy/miscellaneous.js
yeul: This was moved up to DE because it was causing GB to have a ',' instead of a '.' in its amount. The only time there should be a comma is when the country is DE since that is how they format their currency. This was only happening locally in our dev server so it's more of a QoL consistency thing.

tests/functional/spec/US/text.test.js
yeul: This actually existed prior to my changes and was just the way the text.test file was set up. I believe the logic behind it when Ryan created it was that it was safe to test the smallest, the default, and the largest as they are most likely to be used and if those are okay, the likelihood of the others being okay too is high. I don't think we need 4x the extra text screenshots, personally.

server/locale/US/Pi4/mutations/index.js
yeul: I named it after the `offerType` inside the pay_in_four and pay_in_four_q banners since PI4 and PI4Q are used in devServerProxy and it keeps with that convention. I could totally add a clarifying comment though, otherwise the names would be super long i.e. pay_in_four_non_qual.js and pay_in_four_no_amount.js. How about this?

```
/**
 * pi4_no_amount and pi4_non_qual correspond to the message pay_in_four.json
 * pi4q corresponds to the message pay_in_four_q.json
 */
```

server/locale/US/Pi4/styles/text/base.css
yeul: Yup they totally are, this was a placeholder until your changes came through so I cam remove it now!

server/locale/US/Pi4/mutations/index.js
merlinstardust: The comment helps 👍

tests/functional/spec/US/text.test.js
merlinstardust:
```
accounts.forEach(account => {
  if (account === 'DEV0000000PI4') {
    amounts.forEach(amount => {
      accountTest(account, amount);
    }
  }
  else {
    accountTest(account, amount)
  }
}
```

server/locale/index.js
yeul: I've gone ahead and renamed it to `offerType`.

merlinstardust (approved):

PR #284, opened 2020-09-15
feat: US GPL modal

gillycheesesteak:


src/components/modal/content/DE/parts/Calculator.jsx
```diff
@@ -1,25 +1,25 @@
 /** @jsx h */
 import { h, Fragment } from 'preact';
 
-import { useCalculator } from '../../../lib';
+import { useCalculator, useContent } from '../../../lib';
```
merlinstardust, line 4: This triple `../` makes a good case for us using an aliased root import. It's a little hard to follow

src/components/modal/content/DE/parts/Calculator.jsx
```diff
@@ -1,25 +1,25 @@
...
 
 const Calculator = () => {
     const { terms, value, isLoading, submit, changeInput } = useCalculator();
+    const { content } = useContent('INST');
 
     return (
         <Fragment>
             <div className="calculator">
-                <h3 className="title">Monatliche Raten berechnen</h3>
+                <h3 className="title">{content.calculator.title}</h3>
```
merlinstardust: Given how often the subfields of `content` and `terms` are called, it would be good to extract them to simplify the code a bit

src/components/modal/content/DE/parts/Disclaimer.jsx
```diff
@@ -1,25 +1,23 @@
 /** @jsx h */
 import { h } from 'preact';
+import { useContent } from '../../../../lib';
```
merlinstardust, line 3: Oof... this quadruple `../` is even worse. We definitely need some root importing

src/components/modal/content/DE/parts/Disclaimer.jsx
```diff
@@ -1,25 +1,23 @@
...
-                : `Anspruchsberechtigte Kunden müssen PayPal ein SEPA-Lastschriftmandat erteilen sowie über ein deutsches PayPal-Privatkonto mit bestätigtem Bankkonto als Zahlungsquelle verfügen. Angaben gemäß § 6a PAngV: effektiver Jahreszins ${offer.apr}%, fester Sollzinssatz ${offer.nominalRate}%, Nettodarlehensbetrag von ${minAmount} EUR bis ${maxAmount} EUR, Laufzeit ${offer.term} Monate. Bonität vorausgesetzt. Kreditgeber: PayPal (Europe) S.à.r.l. et Cie, S.C.A., 22–24 Boulevard Royal, L-2449 Luxembourg. Repräsentatives Beispiel gem. § 6a Abs. 4 PAngV: Nettodarlehensbetrag 1.000,00 EUR, Gesamtbetrag 1.052,54 EUR, monatl. Rate 87,71 EUR, 12 Raten, Laufzeit 12 Monate, fester Sollzinssatz 9,56%, effektiver Jahreszins 9,99%. `;
+        const disclosure = Number(offer.apr.replace(/[,.]/g, '')) === 0 ? content.disclosure0 : content.disclosure;
```
merlinstardust: 1. Would be good to abstract `Number(offer.apr.replace(/[,.]/g, '')) === 0` to something like `isZeroApr` just to make this a little clearer. Took me a minute to figure out what this was doing

2. Can `disclosure0` be renamed to `disclosureFor0Apr` or similar? My first thought was this some kind of strangely represented array

src/components/modal/content/DE/parts/INST.jsx
```diff
@@ -3,34 +3,21 @@ import { h } from 'preact';
...
-        <div className="calculator-container">
-            <Calculator country="DE" />
-        </div>
-    </section>
-);
+    return (
+        <section className="content-body">
+            <Carousel items={content.carousel} />
+
+            {/* <!-- Calculator --> */}
```
merlinstardust: While I usually prefer more comments and not less, this one seems unnecessary. The `Calculator` component and `calculator-container` class make it very obvious

src/components/modal/content/DE/parts/TermsTable.jsx
```diff
@@ -1,33 +1,22 @@
...
-    );
+    const { content } = useContent('INST');
+
+    const genericError = <h3 className="error">{content.terms.genericError}</h3>;
 
     if (terms.error || !terms.maxAmount) {
         return genericError;
     }
 
     if (+terms.amount < terms.minAmount && terms.type === 'pala') {
```
merlinstardust: Similar to above, the subfields of `terms` could be pulled out.

Also, what's the initial `+` do? It's been a while since I have had to use that

src/components/modal/content/DE/styles/header.scss
```diff
@@ -0,0 +1,55 @@
+.header-wrapper {
```
merlinstardust, line 1: +1000 for SCSS. So beautiful.

src/components/modal/parts/Tabs.jsx
```diff
@@ -22,14 +23,19 @@ const Tabs = ({ tabs }) => {
         }
     }, [transitionState, initialTab]);
 
+    const hasHeader = tabs.some(tab => Boolean(tab.header));
     // TODO: Accessibility
```
merlinstardust, line 43: So I know this was already here, but do you know what needs to be done in terms of accessibility and can you add it here with this comment?

I don't think it should be incorporated into your story, but I think if the comment had more detail, the next dev to see it would be more encouraged to try to make this accessible

merlinstardust (changes requested):

src/components/modal/content/DE/parts/Calculator.jsx
Seavenly: We would love to do this, but I believe we are currently limited by the common SDK webpack config that we utilize.

PR #285, opened 2020-09-15
refactor: instrumentation updates for tag sizes

paypal-are:
# Description #

- Updates banner json messages to include `offerTerm` in support of adding that field to the `flf1` fpti tag.
- Examines size of message sections in order to instrument fallback message.

## Examples ##
Banner Updates
```
{
    "meta": {
        "offerCountry": "US",
        "offerType": "NI",
        "offerTerm": "${CREDIT_OFFERS_DS.total_payments}",
...
```

Fallback instrumentation message
```
headline:MEDIUM::subheadline:NONE::disclaimer:XSMALL
```


src/utils/fallbackInstrumentation.js
```diff
@@ -0,0 +1,58 @@
+import objectEntries from 'core-js-pure/stable/object/entries';
+
+// Using spread operator here (e.g [...node.children] results in [HtmlCollection]
+// rather than [child, child, ...] for some reason
+const getChildren = node => Array.prototype.slice.call(node.children);
+
+// Map class name of visible element to size tag.
+const toTagSize = classList => {
+    const tagMap = {
```
gillycheesesteak: I think this can be simplified to reverse the way this class name is generated (`tag--<tagName>`)

src/utils/fallbackInstrumentation.js
```diff
@@ -0,0 +1,58 @@
...
+};
+
+const getTagSize = node => {
+    // Provide NONE as value for section in payload if not shown
+    if (!node) {
+        return 'NONE';
+    }
+
+    // Get the tag size of the visible element
+    return getChildren(node).reduce((acc, next) => {
```
gillycheesesteak: Do these reduce calls make more sense as `Array.find`? Either way, we have to use the polyfills for those methods

banners/US/pala_single_gtz.json
```diff
@@ -2,6 +2,7 @@
     "meta": {
         "offerCountry": "US",
         "offerType": "PALA:SINGLE:GTZ",
+        "offerTerm": "${CREDIT_OFFERS_DS.total_payments}",
```
gillycheesesteak, line 5: Make sure to add this field in the message content for other locales

src/utils/stats.js
```diff
@@ -35,7 +37,9 @@ export function runStats({ container, index }) {
         pos_y: Math.round(containerRect.top).toString(),
         browser_width: window.innerWidth.toString(),
         browser_height: window.innerHeight.toString(),
-        visible: isInViewport(container).toString()
+        visible: isInViewport(container).toString(),
+        // Query iframe to determine size of message sections
+        banner_type: instrumentFallback(document.querySelector(`[data-pp-id="${index}"] iframe`))
```
Seavenly: Since we already have the `data-pp-id` element as `container`, I would imagine `container.querySelector('iframe')` could be slightly more performant.

src/utils/fallbackInstrumentation.js
```diff
@@ -0,0 +1,47 @@
+import arrayFind from 'core-js-pure/stable/array/find';
+
+// Using spread operator here (e.g [...node.children] results in [HtmlCollection]
+// rather than [child, child, ...] for some reason
+const getChildren = node => Array.prototype.slice.call(node.children);
```
Seavenly: Consider using `Array.from()` (`arrayFrom` in `core-js-pure`) to be more declarative.

src/utils/fallbackInstrumentation.js
```diff
@@ -0,0 +1,46 @@
...
+    const visibleElement = arrayFind(
+        getChildren(node),
+        element => window.getComputedStyle(element).getPropertyValue('display') !== 'none'
+    );
+
+    // Get the tag size of the element shown
+    return toTagSize(visibleElement.classList);
+};
+
+export function instrumentFallback(container) {
```
Seavenly: What is the significance of this name? Would something like `getActiveTags` be more appropriate?

src/utils/fallbackInstrumentation.js
paypal-are: Linter didn't like `Array.from()`:

` 5:29  error  Prefer the spread operator over `Array.from()`  unicorn/prefer-spread`

I'll switch it to `arrayFrom`.

src/utils/stats.js
paypal-are: `bannerType` is being passed directly to stats now, so this line no longer queries the DOM.

src/utils/fallbackInstrumentation.js
```diff
@@ -0,0 +1,47 @@
+import arrayFind from 'core-js-pure/stable/array/find';
+import arrayFrom from 'core-js-pure/stable/array/from';
+
+// Using spread operator here (e.g [...node.children] results in [HtmlCollection]
+// rather than [child, child, ...] for some reason
+const getChildren = node => arrayFrom(node.children);
```
Seavenly: Since this is only being called once, do you think that `arrayFrom(node.children)` is declarative enough to replace with `getChildren(node)` below?

src/utils/index.js
```diff
@@ -9,6 +9,7 @@ export * from './logger';
 export * from './stats';
 export * from './experimentation';
 export * from './observers';
+export * from './fallbackInstrumentation';
```
Seavenly: Can we rename the file to better match the function it exports?

PR #286, opened 2020-09-15
refactor: remove merchant.js builds

yeul:


server/message/index.jsx
```diff
@@ -1,5 +1,6 @@
 /** @jsx h */
 /** @jsxFrag Fragment */
+// eslint-disable-next-line no-unused-vars
```
gillycheesesteak: Run a fresh `npm i` and you should be able to remove these disables for `Fragment`

PR #331, opened 2020-10-06
refactor: update Flex to Pay in 3 for GB text messages

yeul:


server/locale/GB/mutations/mediaQueries.js
```diff
@@ -1,5 +1,5 @@
-// Media query used to fallback to "Pay Later with Flex."
-export function fallbackMediaQuery(breakpoint) {
+// Media query used to fallback to "Buy now, pay later."
+export function xSmallTagMediaQuery(breakpoint) {
```
merlinstardust: Hmmm... So this code looks good but I'm wondering if it would be better if the names were combined into `xSmallFallbackMediaQuery`.

Also, I'm not sure if media query needs to be included since that's stated with the import.

server/locale/GB/mutations/mediaQueries.js
```diff
@@ -34,6 +34,37 @@ export function fallbackMediaQuery(breakpoint) {
...
+        
+        .message__headline .tag--small > span {
+            white-space: nowrap;
+        }
+    }
+`;
+}
+
 // Moves PayPal logo to position right for logo primary message config.
 export function gbPLContentMediaQuery(breakpoint) {
```
merlinstardust: Since we tend to use GBPL as basically a single word, and I am strangely pedantic about `camelCaseCode`, I think this should be `gbplContentMediaQuery`.

Similarly, the alt query below should also be `gbplAltContentMediaQuery`.

And as per above, I'm not sure if `MediaQuery` is necessary to specify. Or if you think we should, we should start doing this for all other media queries. Or we could import it as an object and instead do `mediaQueries.gbplAltContent`. Thoughts?

merlinstardust (changes requested):
Code looks good other than the two comments, which are more something to discuss if we need to change it at all

merlinstardust (approved):

PR #339, opened 2020-10-08
test: Improve functional tests v2

merlinstardust:
2nd attempt at improving snapshot tests. This is based off of #256 and seems to be improved with SSR changes


.travis.yml
```diff
@@ -18,16 +30,72 @@ before_install:
...
+              - ./scripts/snapshot-tests/update-snapshots.sh
+
+        # - name: 'Functional Tests'
+        #   if: env(DIRTY_SNAPSHOTS) = 0
+        #   script:
+        #       - npm run test:func -- --testPathPattern $TEST_PATH_PATTERN
+        #       - ./scripts/snapshot-tests/collect-diffs.sh
+
+        # DE
+        - name: 'Functional DE Banner Tests'
```
gillycheesesteak: I'm worried about the scalability of this approach - however, I see some commented lines above that may be related, so if there's already an idea about that, ignore this

scripts/snapshot-tests/run-tests.sh
```diff
@@ -1,18 +1,48 @@
...
+    cat $DIFF_FOLDERS_LIST
+    echo ''
+
+    echo ''
+    find ./tests/functional/ -type f -name '*png' | grep -h __diff_output__ > $DIFF_FILES_LIST
+    diffFileCount=$(wc -l < $DIFF_FILES_LIST)
+    echo "DIFF FILES FOUND ($diffFileCount found)"
+    cat $DIFF_FILES_LIST
+    echo ''
+
```
gillycheesesteak: Is this code repeated from `collect-diffs.sh`? Can that be consolidated?

scripts/snapshot-tests/run-tests.sh
```diff
@@ -1,18 +1,48 @@
...
 
-    git checkout develop
-    git add .
-    git commit -m "chore(snapshots): update snapshots [skip ci]"
-    git push
+        git checkout develop
+        git add .
+        git commit -m "chore(snapshots): update snapshots [skip ci]"
+        git push
+    else
```
gillycheesesteak: Same comment as above but for `update-snapshots.sh`

.travis.yml
merlinstardust: Yeah, I'm waiting to hear from support on using the env variables in the script phase. Ideally, I would like to do something like one of these options

```
npm run test:func -- --testPathPattern env(TEST_PATH_PATTERN)
npm run test:func -- --testPathPattern $TEST_PATH_PATTERN
```

But if that's not an option, I can use the env var from inside a single-line bash script instead.

```
# .travis.yml
./scripts/snapshot-tests/run-by-pattern.sh

# run-by-pattern.sh
npm run test:func -- --testPathPattern $TEST_PATH_PATTERN
```

I didn't do that initially because a single-line bash script was more verbose and failures inside of bash scripts can sometimes be ignored. I have a fix for the ignored failures (`set -ev`), so it's fine, but the bash script really shouldn't be needed.

scripts/snapshot-tests/run-tests.sh
merlinstardust: For both these comments, I do plan to remove `run-tests.sh` but I wanted to ensure that the `.travis.yml` changes were working before I did so.

PR #372, opened 2020-10-26
fix: add local modalContent route

merlinstardust:


src/components/modal/lib/utils.js
```diff
@@ -20,7 +20,7 @@ export const getContent = memoizeOnProps(
             )
             .slice(1);
 
-        return request('GET', `${window.location.origin}/credit-presentment/modalContent?${query}`).then(
+        return request('GET', `${window.location.origin}/credit-presentment/calculateTerms?${query}`).then(
```
gillycheesesteak: The change to the modalContent endpoint was intentional to make sure that other reactive values in the modal are updated with recalculation - what bug is this intended to fix?

merlinstardust: In the calculator hook reducer, the `terms` case kept getting a `Cannot read formattedAmount of undefined` error, which was causing the tests for the modal calculator test to fail.

PR #373, opened 2020-10-26
test: SSR unit tests

gillycheesesteak:


merlinstardust:
Good set to start with. I think having unit tests for MutatedText, Styles, mutations, and the locale getters would be good as well

PR #397, opened 2020-12-03
refactor: Add ignore_cache option

jadutter:
Add ignore cache option for messages

### DIFF
--- a/src/zoid/message/validation.js
+++ b/src/zoid/message/validation.js
@@ -169,12 +169,8 @@ export default {
     },
     ignoreCache: ({ props: { ignoreCache } }) => {
         if (typeof ignoreCache !== 'undefined') {
-            const options = [true, false];
-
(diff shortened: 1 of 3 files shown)


src/zoid/message/validation.js
```diff
@@ -165,6 +165,21 @@ export default {
...
 
+        return undefined;
+    },
+    ignoreCache: ({ props: { ignoreCache } }) => {
+        if (typeof ignoreCache !== 'undefined') {
+            const options = [true, false];
+
+            if (!validateType(Types.BOOLEAN, ignoreCache)) {
+                logInvalidType('ignoreCache', Types.BOOLEAN, ignoreCache);
+            } else if (!arrayIncludes(options, ignoreCache)) {
```
Seavenly: We should be able to safely remove this if check since a boolean by definition only supports the 2 values and we're ensuring it's a boolean right before this step.

Seavenly (approved):

PR #398, opened 2020-12-08
feat: Add fontFamily and fontSrc

jadutter:
Add fontFamily and fontSrc option to text layout.

Allow merchants to pass fontFamily and fontSrc options in order to customize the font for upstream messages.

### DIFF
--- a/server/locale/DE/validOptions.js
+++ b/server/locale/DE/validOptions.js
@@ -9,7 +9,7 @@ export default {
             color: [Types.STRING, ['black', 'white']],
             size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
             fontFamily: [Types.STRING],
-            fontSource: [Types.STRING]
+            fontSource: [Types.ANY]
         },
         preset: [Types.STRING, [undefined, 'smallest']]
     },
@@ -18,10 +18,8 @@ export default {
         ratio: [Types.STRING, ['1x1', '1x4', '8x1', '20x1']],
         preset: [Types.STRING, [undefined, 'smallest']],
         text: {
-            color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
-            // size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
             fontFamily: [Types.STRING],
-            fontSource: [Types.STRING]
+            fontSource: [Types.ANY]
         }
     }
 };
--- a/server/locale/DE/validOptions.js
+++ b/server/locale/DE/validOptions.js
@@ -16,6 +16,12 @@ export default {
     flex: {
         color: [Types.STRING, ['blue', 'black', 'white', 'gray|grey']],
         ratio: [Types.STRING, ['1x1', '1x4', '8x1', '20x1']],
-        preset: [Types.STRING, [undefined, 'smallest']]
+        preset: [Types.STRING, [undefined, 'smallest']],
+        text: {
+            color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
+            // size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
+            fontFamily: [Types.STRING],
+            fontSrc: [Types.STRING]
+        }
     }
 };
--- a/server/locale/DE/validOptions.js
+++ b/server/locale/DE/validOptions.js
@@ -8,7 +8,8 @@ export default {
         text: {
             color: [Types.STRING, ['black', 'white']],
             size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
-            fontFamily: [Types.STRING]
+            fontFamily: [Types.STRING],
+            fontSrc: [Types.STRING]
         },
         preset: [Types.STRING, [undefined, 'smallest']]
     },
--- a/server/locale/DE/validOptions.js
+++ b/server/locale/DE/validOptions.js
@@ -9,7 +9,7 @@ export default {
             color: [Types.STRING, ['black', 'white']],
             size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
             fontFamily: [Types.STRING],
-            fontSrc: [Types.MULTI_STRING]
+            fontSrc: [Types.STRING]
         },
         preset: [Types.STRING, [undefined, 'smallest']]
     },
@@ -21,7 +21,7 @@ export default {
             color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
             // size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
             fontFamily: [Types.STRING],
-            fontSrc: [Types.MULTI_STRING]
+            fontSrc: [Types.STRING]
         }
     }
 };
--- a/server/locale/DE/validOptions.js
+++ b/server/locale/DE/validOptions.js
@@ -9,7 +9,7 @@ export default {
             color: [Types.STRING, ['black', 'white']],
             size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
             fontFamily: [Types.STRING],
-            fontSrc: [Types.STRING]
+            fontSource: [Types.STRING]
         },
         preset: [Types.STRING, [undefined, 'smallest']]
     },
@@ -21,7 +21,7 @@ export default {
             color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
             // size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
             fontFamily: [Types.STRING],
-            fontSrc: [Types.STRING]
+            fontSource: [Types.STRING]
         }
     }
 };
--- a/server/locale/DE/validOptions.js
+++ b/server/locale/DE/validOptions.js
@@ -9,7 +9,7 @@ export default {
             color: [Types.STRING, ['black', 'white']],
             size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
             fontFamily: [Types.STRING],
-            fontSrc: [Types.STRING]
+            fontSrc: [Types.MULTI_STRING]
         },
         preset: [Types.STRING, [undefined, 'smallest']]
     },
@@ -21,7 +21,7 @@ export default {
             color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
             // size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
             fontFamily: [Types.STRING],
-            fontSrc: [Types.STRING]
+            fontSrc: [Types.MULTI_STRING]
         }
     }
 };
--- a/server/locale/GB/validOptions.js
+++ b/server/locale/GB/validOptions.js
@@ -10,7 +10,7 @@ export default {
             color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
             size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
             fontFamily: [Types.STRING],
-            fontSource: [Types.STRING]
+            fontSource: [Types.ANY]
         },
         preset: [Types.STRING, [undefined, 'smallest']]
     },
@@ -19,10 +19,8 @@ export default {
         ratio: [Types.STRING, ['1x1', '1x4', '8x1', '20x1']],
         preset: [Types.STRING, [undefined, 'smallest']],
         text: {
-            color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
-            // size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
             fontFamily: [Types.STRING],
-            fontSource: [Types.STRING]
+            fontSource: [Types.ANY]
         }
     }
 };
--- a/server/locale/GB/validOptions.js
+++ b/server/locale/GB/validOptions.js
@@ -10,7 +10,7 @@ export default {
             color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
             size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
             fontFamily: [Types.STRING],
-            fontSrc: [Types.MULTI_STRING]
+            fontSrc: [Types.STRING]
         },
         preset: [Types.STRING, [undefined, 'smallest']]
     },
@@ -22,7 +22,7 @@ export default {
             color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
             // size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
             fontFamily: [Types.STRING],
-            fontSrc: [Types.MULTI_STRING]
+            fontSrc: [Types.STRING]
         }
     }
 };
--- a/server/locale/GB/validOptions.js
+++ b/server/locale/GB/validOptions.js
@@ -9,7 +9,8 @@ export default {
         text: {
             color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
             size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
-            fontFamily: [Types.STRING]
+            fontFamily: [Types.STRING],
+            fontSrc: [Types.STRING]
         },
         preset: [Types.STRING, [undefined, 'smallest']]
     },
--- a/server/locale/GB/validOptions.js
+++ b/server/locale/GB/validOptions.js
@@ -10,7 +10,7 @@ export default {
             color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
             size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
             fontFamily: [Types.STRING],
-            fontSrc: [Types.STRING]
+            fontSrc: [Types.MULTI_STRING]
         },
         preset: [Types.STRING, [undefined, 'smallest']]
     },
@@ -22,7 +22,7 @@ export default {
             color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
             // size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
             fontFamily: [Types.STRING],
-            fontSrc: [Types.STRING]
+            fontSrc: [Types.MULTI_STRING]
         }
     }
 };
--- a/server/locale/GB/validOptions.js
+++ b/server/locale/GB/validOptions.js
@@ -17,6 +17,12 @@ export default {
     flex: {
         color: [Types.STRING, ['blue', 'black', 'white', 'gray|grey', 'monochrome', 'grayscale|greyscale']],
         ratio: [Types.STRING, ['1x1', '1x4', '8x1', '20x1']],
-        preset: [Types.STRING, [undefined, 'smallest']]
+        preset: [Types.STRING, [undefined, 'smallest']],
+        text: {
+            color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
+            // size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
+            fontFamily: [Types.STRING],
+            fontSrc: [Types.STRING]
+        }
     }
 };
--- a/server/locale/GB/validOptions.js
+++ b/server/locale/GB/validOptions.js
@@ -10,7 +10,7 @@ export default {
             color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
             size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
             fontFamily: [Types.STRING],
-            fontSrc: [Types.STRING]
+            fontSource: [Types.STRING]
         },
         preset: [Types.STRING, [undefined, 'smallest']]
     },
@@ -22,7 +22,7 @@ export default {
             color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
             // size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
             fontFamily: [Types.STRING],
-            fontSrc: [Types.STRING]
+            fontSource: [Types.STRING]
         }
     }
 };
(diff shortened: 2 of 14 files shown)


server/locale/DE/validOptions.js
```diff
@@ -8,13 +8,20 @@ export default {
...
         color: [Types.STRING, ['blue', 'black', 'white', 'gray|grey']],
         ratio: [Types.STRING, ['1x1', '1x4', '8x1', '20x1']],
-        preset: [Types.STRING, [undefined, 'smallest']]
+        preset: [Types.STRING, [undefined, 'smallest']],
+        text: {
+            color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
+            // size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
+            fontFamily: [Types.STRING],
+            fontSrc: [Types.MULTI_STRING]
+        }
```
merlinstardust, line 23: For flex, I think we can simplify this to just have it be `flex.fontFamily` and `flex.fontSrc`. Color options can just be handled at `flex.color`.

server/locale/GB/validOptions.js
```diff
@@ -9,13 +9,20 @@ export default {
...
     flex: {
         color: [Types.STRING, ['blue', 'black', 'white', 'gray|grey', 'monochrome', 'grayscale|greyscale']],
         ratio: [Types.STRING, ['1x1', '1x4', '8x1', '20x1']],
-        preset: [Types.STRING, [undefined, 'smallest']]
+        preset: [Types.STRING, [undefined, 'smallest']],
+        text: {
+            color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
+            // size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
+            fontFamily: [Types.STRING],
+            fontSrc: [Types.MULTI_STRING]
```
merlinstardust: As above, we can flatten this and not need `flex.text`, which could also be confusing

server/locale/US/GPL/validOptions.js
```diff
@@ -19,11 +20,23 @@ export default {
             ['blue', 'black', 'white', 'white-no-border', 'gray|grey', 'monochrome', 'grayscale|greyscale']
         ],
         ratio: [Types.STRING, ['1x1', '1x4', '8x1', '20x1']],
-        preset: [Types.STRING, [undefined, 'smallest']]
+        preset: [Types.STRING, [undefined, 'smallest']],
+        text: {
+            color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
+            // size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
+            fontFamily: [Types.STRING],
+            fontSrc: [Types.MULTI_STRING]
```
merlinstardust: As above, we can flatten this and not need `flex.text`

server/locale/US/GPL/validOptions.js
```diff
@@ -19,11 +20,23 @@ export default {
...
     custom: {
         markup: [Types.STRING],
         ratio: [Types.ANY],
-        preset: [Types.STRING, [undefined, 'smallest']]
+        preset: [Types.STRING, [undefined, 'smallest']],
+        text: {
+            color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
+            size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
+            fontFamily: [Types.STRING],
+            fontSrc: [Types.MULTI_STRING]
```
merlinstardust: We shouldn't need these fields in custom, since custom gets defined with a URL pointing to custom HTML files. Ideally, if they're using `fontFamily` or `fontSrc`, they won't need to repeat it in custom

server/locale/US/PPC/validOptions.js
```diff
@@ -19,11 +20,23 @@ export default {
             ['blue', 'black', 'white', 'white-no-border', 'gray|grey', 'monochrome', 'grayscale|greyscale']
         ],
         ratio: [Types.STRING, ['1x1', '1x4', '8x1', '20x1']],
-        preset: [Types.STRING, [undefined, 'smallest']]
+        preset: [Types.STRING, [undefined, 'smallest']],
+        text: {
+            color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
+            // size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
+            fontFamily: [Types.STRING],
+            fontSrc: [Types.MULTI_STRING]
```
merlinstardust: As above, we can flatten this and not need `flex.text`

server/locale/US/PPC/validOptions.js
```diff
@@ -19,11 +20,23 @@ export default {
...
     custom: {
         markup: [Types.STRING],
         ratio: [Types.ANY],
-        preset: [Types.STRING, [undefined, 'smallest']]
+        preset: [Types.STRING, [undefined, 'smallest']],
+        text: {
+            color: [Types.STRING, ['black', 'white', 'monochrome', 'grayscale|greyscale']],
+            size: [Types.NUMBER, [12, 10, 11, 13, 14, 15, 16]],
+            fontFamily: [Types.STRING],
+            fontSrc: [Types.MULTI_STRING]
```
merlinstardust: As above, custom shouldn't need these

server/locale/DE/validOptions.js
merlinstardust: `fontSrc` should be fully spelled out as `fontSource`. It's better to use full spellings unless the abbreviation is extremely long.

server/message/index.jsx
```diff
@@ -36,6 +36,34 @@ const applyCascade = curry((style, flattened, type, rules) =>
         type === Array ? [] : {}
     )
 );
+const getFontSrcRule = fontSrc => {
```
merlinstardust: This function is getting repeated a lot here and in `server/validateStyle` and in `src/zoid/message/validation`, so I think it should be abstracted to not repeat as much code.

server/message/index.jsx
```diff
@@ -36,6 +36,34 @@ const applyCascade = curry((style, flattened, type, rules) =>
...
+        otf: 'otf' // opentype
+        // eot: 'eot', // embedded opentype
+        // svg: 'svg' // svg
+    };
+    if (fontSrc?.constructor === Array) {
+        const ruleVal = fontSrc
+            .map(url => {
+                const match = urlPat.exec(url);
+                const ext = match ? match[2] : null;
+                const fmt = fontFormats[ext];
```
merlinstardust: `ext` and `fmt` should be `extension` and `format`

server/message/index.jsx
```diff
@@ -36,6 +36,34 @@ const applyCascade = curry((style, flattened, type, rules) =>
...
+const getFontSrcRule = fontSrc => {
+    /* eslint-disable-next-line security/detect-unsafe-regex */
+    const urlPat = RegExp(/^(?<url>https?:\/\/.+?\.(?<ext>woff|woff2|otf|tff))$/);
+    const fontFormats = {
+        woff: 'woff', // woff
+        // woff2: 'woff2', // woff2
+        ttf: 'ttf', // truetype
+        otf: 'otf' // opentype
+        // eot: 'eot', // embedded opentype
+        // svg: 'svg' // svg
```
merlinstardust: Since we're ignoring `woff2`, `eot`, and `svg`, these can be removed.

server/message/index.jsx
```diff
@@ -36,6 +36,34 @@ const applyCascade = curry((style, flattened, type, rules) =>
...
+    };
+    if (fontSrc?.constructor === Array) {
```
merlinstardust: This check shouldn't be necessary since we're already forcing an array at the validate step

server/message/index.jsx
merlinstardust: Add newline before the function

server/message/index.jsx
```diff
@@ -62,14 +90,40 @@ export default ({ options, markup, locale }) => {
...
+    if (fontSrc) {
+        customFontStyleRules.push(`
+            @font-face {
+                font-family: '${fontFamily}'; 
+                ${fontSrcRule}
+            }`);
+    }
+    if (fontFamilyRule || textSizeRule) {
+        customFontStyleRules.push(`${fontSelector}{ ${fontFamilyRule}${textSizeRule} }`);
+    }
```
merlinstardust: All these font lines have made it more difficult to read. I think moving all of it in to a function will make it more readable

server/message/index.jsx
```diff
@@ -62,14 +90,40 @@ export default ({ options, markup, locale }) => {
...
     const mutationStyleRules = mutationRules.styles ?? [];
+    const customFontStyleRules = [];
     const miscStyleRules = [];
+    const textSize = layout === 'flex' ? undefined : style.text?.size;
+    const fontSrc = style.text?.fontSrc;
+    const fontFamily = fontSrc ? 'MerchantCustomFont' : style.text?.fontFamily;
+    const fontSelector = `
+.message__messaging, 
+.message__messaging .message__headline span, 
+.message__messaging .message__disclaimer span`;
```
merlinstardust: Indent these and move the closing backtick to the next line so it looks cleaner

server/message/index.jsx
```diff
@@ -62,14 +90,40 @@ export default ({ options, markup, locale }) => {
...
-        // miscStyleRules.push(`.message__headline { font-size: ${textSize}px; }`);
-        // miscStyleRules.push(`.message__disclaimer { font-size: ${textSize}px; }`);
         miscStyleRules.push(`.message__messaging { font-size: ${textSize}px; }`);
     }
+    if (fontSrc) {
+        customFontStyleRules.push(`
+            @font-face {
+                font-family: '${fontFamily}'; 
+                ${fontSrcRule}
+            }`);
```
merlinstardust: This can fit on one line

server/validateStyle.js
```diff
@@ -14,7 +14,38 @@ const logInvalidOption = (addLog, location, options, val) =>
             /^[a-z0-9]+$/i.test(val) ? val : 'REDACTED'
         }".`
     );
-
+function validateFontSrc(fontSrc) {
```
merlinstardust: Add newline before and after the function

server/message/index.jsx
```diff
@@ -62,14 +90,40 @@ export default ({ options, markup, locale }) => {
...
+    const fontSrc = style.text?.fontSrc;
+    const fontFamily = fontSrc ? 'MerchantCustomFont' : style.text?.fontFamily;
+    const fontSelector = `
+.message__messaging, 
+.message__messaging .message__headline span, 
+.message__messaging .message__disclaimer span`;
+    const textSizeRule = textSize ? `font-size: ${textSize}px; ` : '';
+    const fontFamilyRule = fontFamily
+        ? `font-family: '${fontFamily}', PayPal-Sans-Big, PayPal-Sans, Arial, sans-serif; `
+        : '';
```
merlinstardust: In the case that a merchant wants to use a generic font family, then `fontFamily` should not have quotes as generic families are keywords. It should still have quotes when not generic.

Here's an array of generic font-families I found on MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/font-family

```
['serif', 'sans-serif', 'monospace', 'cursive', 'fantasy', 'system-ui', 'ui-serif', 'ui-sans-serif', 'ui-monospace', 'math', 'emoji', 'fangsong']
```

src/zoid/message/validation.js
```diff
@@ -103,6 +103,33 @@ export default {
...
+                        otf: 'otf' // opentype
+                        // eot: 'eot', // embedded opentype
+                        // svg: 'svg' // svg
+                    };
+                    const ext = url?.split('.').slice(-1);
+                    return url && url.indexOf('http') === 0 && fontFormats[ext];
+                });
+            }
+            if (fontSrc) {
+                console.log(`fontSrc:`, fontSrc);
```
merlinstardust: Remove console.log artifacts

server/validateStyle.js
```diff
@@ -42,7 +73,18 @@ function getValidVal(addLog, typeArr, val, location) {
...
-
+        if (location === 'style.text.fontSrc') {
+            return validateFontSrc(val);
+        }
+        if (type === Types.MULTI_STRING) {
+            if (typeof val === 'string') {
+                return [val];
+            }
+            if (val?.constructor === Array) {
+                console.log(val);
```
merlinstardust: Remove console.log artifacts

merlinstardust (changes requested):
The code itself looks pretty good. I'm still having trouble trying to get a fontSrc to load on my machine. Would be worth having another dev or two also try to test it on their machine.

Also, be sure to commit via the command line, so the pre-commit linting check runs. I found the following after running `npm run lint`

```
server/validateStyle.js
  29:24  error    'fontSrc' is already declared in the upper scope  no-shadow
  84:17  warning  Unexpected console statement                      no-console

src/zoid/message/validation.js
  127:17  warning  Unexpected console statement                          no-console
  128:17  error    Assignment to property of function parameter 'style'  no-param-reassign
  130:24  error    Assignment to property of function parameter 'style'  no-param-reassign

✖ 5 problems (3 errors, 2 warnings)
```

server/locale/GB/validOptions.js
jadutter: I’ll admit, I found it confusing too. It took me a bit to recognize we were structuring the object default slightly differently than the input object we’re comparing it against. I figured it was better to follow the convention we were using.

If we moved flex.text.fontSource to flex.fontSource, then I think text.text.fontSource should likewise move to text.fontSource.

Ditto for fontFamily.

server/message/index.jsx
jadutter: I'll add a function to check if they're using one of those generic names.

server/message/index.jsx
```diff
@@ -36,8 +40,111 @@ const applyCascade = curry((style, flattened, type, rules) =>
...
+        svg: 'svg' // svg
+    };
+    const urlPattern = RegExp(`^(?<url>https?://.+?.(?<ext>${[...Object.values(fontFormats)].join('|')}))$`);
+    const payload = {
+        fontSource: val,
+        validValues: [],
+        error: null
+    };
+    try {
+        const fontSource = val;
```
Seavenly: Can the parameter `val` just be renamed to `fontSource`?

server/message/index.jsx
```diff
@@ -36,8 +40,111 @@ const applyCascade = curry((style, flattened, type, rules) =>
...
+    }
+    const fontFormats = {
+        woff: 'woff', // woff
+        woff2: 'woff2', // woff2
+        ttf: 'ttf', // truetype
+        otf: 'otf', // opentype
+        eot: 'eot', // embedded opentype
+        svg: 'svg' // svg
+    };
+    const urlPattern = RegExp(`^(?<url>https?://.+?.(?<ext>${[...Object.values(fontFormats)].join('|')}))$`);
```
Seavenly: It doesn't look like we're actually using the named capture groups. Would it be sufficient without them? Also, are we not supporting relative paths?

server/message/index.jsx
```diff
@@ -36,8 +40,111 @@ const applyCascade = curry((style, flattened, type, rules) =>
...
+    const payload = {
+        fontSource: val,
+        validValues: [],
+        error: null
+    };
+    try {
+        const fontSource = val;
+        payload.fontSource = fontSource;
+        const ruleVal = (fontSource?.constructor === Array ? fontSource : [fontSource])
```
Seavenly: Consider using the more declarative `Array.isArray()`

server/message/index.jsx
```diff
@@ -36,8 +40,111 @@ const applyCascade = curry((style, flattened, type, rules) =>
...
+            .slice(0, 10)
+            .filter(e => typeof e === 'string')
```
Seavenly: At this point, these input aspects should have been vetted. Can these be removed?

server/message/index.jsx
```diff
@@ -36,8 +40,111 @@ const applyCascade = curry((style, flattened, type, rules) =>
...
+            .map(url => {
+                const match = urlPattern.exec(url);
+                const extension = match ? match[2] : null;
```
Seavenly:
```suggestion
                const extension = urlPattern.exec(url)?.[2];
```

server/message/index.jsx
```diff
@@ -36,8 +40,111 @@ const applyCascade = curry((style, flattened, type, rules) =>
...
+                const format = fontFormats[extension];
+                if (url && format) {
```
Seavenly: We should be able to safely assume `url` exists at this point.
``` suggestion
                if (format) {
```

server/message/index.jsx
```diff
@@ -36,8 +40,111 @@ const applyCascade = curry((style, flattened, type, rules) =>
...
+                    payload.validValues.push(url);
+                    return `url('${url}') format('${format}')`;
+                }
+                return '';
+            })
+            .filter(e => e)
```
Seavenly:
```suggestion
            .filter(Boolean)
```

server/message/index.jsx
```diff
@@ -36,8 +40,111 @@ const applyCascade = curry((style, flattened, type, rules) =>
...
+        ttf: 'ttf', // truetype
+        otf: 'otf', // opentype
+        eot: 'eot', // embedded opentype
+        svg: 'svg' // svg
+    };
+    const urlPattern = RegExp(`^(?<url>https?://.+?.(?<ext>${[...Object.values(fontFormats)].join('|')}))$`);
+    const payload = {
+        fontSource: val,
+        validValues: [],
+        error: null
```
Seavenly: Can we just omit the `error` value completely unless it exists?

server/message/index.jsx
```diff
@@ -36,8 +40,111 @@ const applyCascade = curry((style, flattened, type, rules) =>
...
+const getFontFamilyRule = (addLog, val) => {
+    if (!val) {
+        return '';
+    }
+    const payload = {
+        fontFamily: val,
+        validValue: null
+    };
+    /* eslint-disable-next-line security/detect-unsafe-regex */
+    const fontPattern = RegExp(/^(?<font>([a-zA-Z0-9._-]\s*)+?)$/);
```
Seavenly: Do we need to validate the font here when it should have been vetted prior to this?

server/message/index.jsx
```diff
@@ -36,8 +40,111 @@ const applyCascade = curry((style, flattened, type, rules) =>
...
+        cursive: 'cursive',
+        fantasy: 'fantasy',
+        'system-ui': 'system-ui',
+        'ui-serif': 'ui-serif',
+        'ui-sans-serif': 'ui-sans-serif',
+        'ui-monospace': 'ui-monospace',
+        math: 'math',
+        emoji: 'emoji',
+        fangsong: 'fangsong'
+    };
```
Seavenly: It looks like this is needed because the generic font families can't be surrounded with `'`s, but specific font names should be to capture spaces. Can we add a comment explaining this reasoning Do we need to have `math`, `emoji`, and `fangsong` in the list?

server/message/index.jsx
```diff
@@ -36,8 +40,111 @@ const applyCascade = curry((style, flattened, type, rules) =>
...
+    const fontFamily = typeof val === 'string' ? genericFamilies[val] ?? `'${val}'` : '';
```
Seavenly: It should be safe to assume `val` is a string from prior vetting.

src/utils/elements.js
```diff
@@ -37,15 +37,31 @@ export function isElement(el) {
...
+            try {
+                return flattenedToObject(name, JSON.parse(value));
+            } catch (err) {
+                console.error(err.stack);
+            }
+            try {
+                return flattenedToObject(name, JSON.parse(value.replace(/'/g, '"')));
+            } catch (err) {
+                console.error(err.stack);
+            }
```
Seavenly: Instead of 2 try blocks consider just using `return flattenedToObject(name, JSON.parse(value.replace(/'/g, '"')));` since it will cover both cases.

src/utils/elements.js
```diff
@@ -37,15 +37,31 @@ export function isElement(el) {
 export function getInlineOptions(container) {
     // Allows for data attributes dependent on camel casing to function properly.
     const attributeNameOverride = {
-        'data-pp-buyercountry': 'data-pp-buyerCountry'
+        'data-pp-buyercountry': 'data-pp-buyerCountry',
+        'data-pp-style-text-fontfamily': 'data-pp-style-text-fontFamily',
+        'data-pp-style-text-fontsource': 'data-pp-style-text-fontSource'
+    };
+    const getOptionValue = (name, value) => {
+        if (name === 'style-text-fontSource' && value?.indexOf('[') >= 0) {
```
Seavenly: Consider using `String.prototype.startsWith` as opposed to `indexOf() >= 0`. It requires importing the polyfill manually from `core-js-pure` but it's generally preferred due to its declarative nature.

server/message/index.jsx
jadutter: No, we wouldn't want to support relative pathing. Since this goes into our iframe, it would grab the path relative to our iframe. 
To support relative urls, we'd likely need to modify it so that fontSource values get checked and appended with the page origin before the request went out to cpnw. 

I'd initially tried pulling out the named capture groups, but it seems they're not supported in regular babel. I wasn't sure if it was acceptable to add `@babel/plugin-transform-named-capturing-groups-regex` to the sdk, so I refactored it to pull out the items by index. 
I'd be more than happy to use named capture groups for this instead.

server/message/index.jsx
jadutter: Yes. In previous commits, I was changing what was stored in fontSource, so I was trying not to modify the input parameter. I'll simplify this.

server/message/index.jsx
jadutter: I made the change to my local copy after you mentioned the same thing on cpnw.

server/message/index.jsx
jadutter: Updated.

server/message/index.jsx
jadutter: Updating here, and other locations where I was filtering for truthy values.

server/message/index.jsx
jadutter: We don't need it anymore. I'll remove it.

server/message/index.jsx
jadutter: This list was pulled from MDN's list of generic names, without further thought. 

I think we can safely drop `math` and `emoji`, but we may want to retain `fangsong` to future proof needing to support Chinese fonts.

server/message/index.jsx
jadutter: Keeping this as is since we cannot add the babel plugin for regex named groups.

PR #399, opened 2020-12-10
feat: add GPL France messages and modals

yeul:


jadutter (approved):

banners/FR/gplq.json
```diff
@@ -0,0 +1,20 @@
...
+        "lander": "",
+        "variables": {
+            "offer_id": "${CREDIT_OFFERS_DS.offer_id}",
+            "vendor_financing_id": "${CREDIT_OFFERS_DS.vendor_financing_id}",
+            "financing_code": "${CREDIT_OFFERS_DS.financing_code}",
+            "formattedPeriodicPayment": "${CREDIT_OFFERS_DS.formattedPeriodicPayment}"
+        }
+    },
+
+    "headline": [[["Payez en 4X."], ["default", "small"]]],
```
merlinstardust, line 15: I think there should be an option text for non-qualifying amounts here?

`Payez en 4X pour les achats de 50 € à 2000 €.`

server/locale/FR/mutations/gpl.js
```diff
@@ -0,0 +1,90 @@
...
+                headline: [
+                    {
+                        tag: 'default',
+                        br: ['les']
+                    }
+                ]
+            })
+        ],
+        [
+            'logo.type:alternative',
```
merlinstardust, line 50: Is alternative right supposed to have the logo on the right side? If it is, I'm not seeing it show up on the right. If not, then ignore this.

server/locale/FR/mutations/common.js
```diff
@@ -0,0 +1,20 @@
+import Logo from '../../../message/logos';
+
+/**
+ * Assigns appropriate logo based on message config options.
+ * Used in mutations files where config settings are the same across many accounts.
+ */
+
+export const textLogoMutations = [
+    ['text.color:white && logo.type:primary', { logo: Logo.PP_PAYPAL.WHITE }],
```
merlinstardust: Since this file only ever contains logo mutations, what do you think about switching the file name to `logos.js` or `logoMutations.js`?

Also, given that we're using the same definition for almost every locale (except DE, I think), maybe we should just have this be a shared mutations file by default? And then it can be overridden by a specific locale?

merlinstardust (changes requested):
Code looks solid. Only had a few comments/questions

banners/FR/gplq.json
yeul: Eventually, but not for launch, there will be a third message created for non-qualifying amounts that have the min and max in the message.

server/locale/FR/mutations/gpl.js
yeul: Nope! We're just following the same convention as GPL US and GB where alternative right is ignored. It might become a thing when we add text alignment to the messages though, but for now it should always be on the left.

server/locale/FR/mutations/common.js
yeul: Good idea, I'll call it `logoMutations` and move it to the message folder and fix the imports for the GPL messages.

yeul:
- Correct link to @paypal-are's GPL FR Flex messages PR for reference: https://github.com/yeul/paypal-messaging-components/pull/9
- Correct link to my GPL FR modal PR for reference: https://github.com/yeul/paypal-messaging-components/pull/8

PR #402, opened 2020-12-15
feat: standalone modal support

Seavenly:


src/components/modal/lib/utils.js
```diff
@@ -28,16 +28,28 @@ export const getContent = memoizeOnProps(
...
         arrayIncludes(
-            ['EZP:ANY:EQZ', 'EZP:ANY:GTZ', 'PALA:MULTI:EQZ', 'PALA:MULTI:GTZ', 'PALA:SINGLE:EQZ', 'PALA:SINGLE:GTZ'],
+            [
+                'EZP:ANY:EQZ',
+                'EZP:ANY:GTZ',
+                'PALA:MULTI:EQZ',
+                'PALA:MULTI:GTZ',
+                'PALA:SINGLE:EQZ',
+                'PALA:SINGLE:GTZ',
+                'EZP'
```
merlinstardust: Small thing but can EZP be put first so all the offers starting with EZP are together?

merlinstardust (approved):
Existing code and tests look good mostly except for the one small change I had. Plus I had a few thoughts beyond that

1. Can you add a functional test for this?
2. Since this feature is called standalone modal, I feel like we should rename the `demo/standalone.html` file to something else to prevent confusion.
3. Maybe this should get its own demo page like `demo/standalone_modal.html`? Thoughts?

merlinstardust (changes requested):
Accidentally hit approve on the last one, so comments are the same, but clicking request this time.

src/controllers/message/setup.js
```diff
@@ -7,27 +7,29 @@ export default function setup() {
     // Populate global config options
     const script = getScript();
     if (script) {
-        const { merchantid, ...inlineScriptOptions } = getInlineOptions(script);
+        const { ...inlineScriptOptions } = getInlineOptions(script);
```
gillycheesesteak: Can this be changed to `const inlineScriptOptions = ...`

gillycheesesteak (approved):

PR #410, opened 2021-01-07
feat: set text as explicit default

merlinstardust:


src/zoid/message/validation.js
```diff
@@ -113,13 +115,12 @@ export default {
             }
         }
 
-        if (validateType(Types.OBJECT, style)) {
-            logInvalidType('style.layout', Types.STRING, style.layout);
-        } else if (typeof style !== 'undefined') {
-            logInvalidType('style', Types.OBJECT, style);
+        if (validateType(Types.OBJECT, styleInput)) {
```
jadutter: Can we combine these if statements?

Delete lines 116-118 so that
``` if (validateType(Types.OBJECT, styleInput)) {```
only runs once on line 106?

File: src/zoid/message/validation.js [hunk 2]
```diff
@@ -42,10 +42,10 @@ const logInvalid = memoize((location, message) =>
 const logInvalidType = (location, expectedType, val) =>
     logInvalid(
         location,
-        `Expected type "${expectedType.toLowerCase()}" but instead received "${typeof val}" (${val}).`
+        `Expected type '${expectedType.toLowerCase()}' but instead received '${typeof val}' (${val}).`
     );
 const logInvalidOption = (location, options, val) =>
-    logInvalid(location, `Expected one of ["${options.join('", "').replace(/\|[\w|]+/g, '')}"] but received "${val}".`);
+    logInvalid(location, `Expected one of ['${options.join('', '').replace(/\|[\w|]+/g, '')}'] but received '${val}'.`);
```
Seavenly: Why are we switching from `"` to `'` in our logging output? Also, this seems like a potential side-effect from a find/replace:
`options.join('', '')`.

jadutter (approved):

PR #421, opened 2021-01-19
feat: add text align parameter to US text messages

yeul:


yeul:
**Note**: Requires CPNW PR 187

PR #458, opened 2021-03-15
feat: DE GPL text messages

yeul:


server/message/parts/MutatedText.jsx
```diff
@@ -17,6 +17,33 @@ const MutatedText = ({ tagData, options }) => {
...
+     * 3. (\.|,)
+     * Either a comma or period decimal
+     * 4. 00
+     * The 00 after the decimal
+     * 5. €?
+     * An optional €
+     * 6. -
+     * A dash
+     * 7. For the rest of the pattern refer to 1-5.
+     */
```
gillycheesesteak, line 34: +100000

.travis.yml
```diff
@@ -10,7 +10,7 @@ node_js:
 
 env:
     global:
-        - DIRTY_SNAPSHOTS=0
+        - DIRTY_SNAPSHOTS=1
```
gillycheesesteak: Can we leave this as 0 until we are ready to merge, so that we can confirm that the only failing snapshots are the ones we expect?

PR #461, opened 2021-03-18
feat: unique iframe titles

jadutter:
increase accessibility by creating unique titles for message and modal iframes


src/zoid/message/containerTemplate.jsx
```diff
@@ -68,7 +68,8 @@ export default ({ uid, frame, prerenderFrame, doc, event, props, container }) =>
             `;
         }
     });
-
+    // eslint-disable-next-line no-param-reassign
+    frame.title = getComponentTitle('Paypal Message ');
```
paypal-are: Is the extra space after `PayPal Message` and later `PayPal Modal` intended?

src/utils/global.js
```diff
@@ -35,6 +35,14 @@ export const nextIndex = () => {
     return globalState.index - 1;
 };
 
+export const getComponentTitle = componentName => {
```
paypal-are: This function could be made simpler by initializing the component counts in `createDefaultState`.

We could update the count with something like:
```
setGlobalState({ ...globalState, ...{ [componentKey]: globalState[componentKey] + 1 } });
```

src/interface/experiment.js
```diff
@@ -50,6 +50,7 @@ export const Messages = config => ({
     }
 });
 
+Messages.__VERSION__ = __MESSAGES__.__VERSION__;
```
Seavenly: If we'd like to more easily expose the version of our library I would recommend we follow suit with the SDK and use `version` as the property name. Also, consider using our utility function `getLibraryVersion()` so that we can minimize the locations where we hardcode the global variables.

src/utils/global.js
paypal-are: Additionally, I thought we had decided to omit the incrementing counters altogether as they didn't add any value to accessibility?

jadutter: They do add to accessibility. If there are multiple banners on a page, it's useful to a user to know which one currently has focus. The rule of thumb for accessibility seemed to be, if the content is visually unique, then the title should be unique.

src/interface/experiment.js
jadutter: I'll make those two changes.

src/zoid/message/containerTemplate.jsx
jadutter: Yes and no. I was adding the space there instead of within the getComponentTitle function, but there was no particular reason for not adding the space inside the function.
getComponentTitle is going away, so I'll consider this resolved.

src/zoid/message/containerTemplate.jsx
```diff
@@ -4,6 +4,8 @@ import { EVENT } from 'zoid/src';
 
 import { overflowObserver } from '../../utils';
 
+let titleCounter = 0;
```
paypal-are: Personal preference, but I would use a closure to keep track of this variable rather than declare it as global:

```
const incTitleCount = () => {
    let count = 0;

    return title => `${title} ${count++}`;
};

const getTitle = incTitleCounter();
```
...

```
<node el={frame} title={  getTitle('PayPal Message') }` ... />
```

src/interface/experiment.js
```diff
@@ -50,6 +51,7 @@ export const Messages = config => ({
     }
 });
 
+Messages.version = getLibraryVersion();
```
paypal-are, line 54: Seems harmless, but I'm not sure if we should include changes unrelated to the story in this PR.

src/zoid/message/containerTemplate.jsx
jadutter: Thank you. I like this better.

paypal-are (approved):

PR #476, opened 2021-03-30
fix: send script UID to server

Seavenly:


src/old/messages/components/Modal/component.js
```diff
@@ -125,6 +126,11 @@ export default getGlobalComponent('__paypal_credit_modal_old__', () =>
                 queryParam: true,
                 sendToChild: false,
                 value: getTargetMeta
+            },
+            scriptUID: {
```
gregjopa, line 130: This usage of `getCurrentScriptUID` looks good to me. We can see that zoid uses the same function when defining the namespace: https://github.com/krakenjs/zoid/blob/6ae489c32c6f605230be070801a92ddf62f0a1dd/src/lib/global.js#L8

gregjopa (approved):
Looks good to me!

gillycheesesteak (approved):

bluepnume (approved):
LGTM!

PR #486, opened 2021-04-06
refactor: demo update

jadutter:
Modified demo to aid in customizing the look of documentation screenshots.

- Changed demo snapshots:
    - padding query parameter to set padding around banners and modals
    - click query parameter on modal to automatically hide the banner, and show the modal
    - added multi-banner.html to be able to compare text banners of different sizes 

- Changed text/flex pages:
    - able to show/hide headers (that show account and configuration)
    - limit labels to only show the parts that change
    - show default behavior by setting a property as undefined


merlinstardust (changes requested):
For the demo changes, can you add a brief explanation in the `docs/` section on the changes and how `VIEW` should be used?

PR #508, opened 2021-04-30
fix: improve automatic message rendering

danzhaas:


src/utils/observers.js
```diff
@@ -24,8 +24,17 @@ export const getInsertionObserver = createGlobalVariableGetter(
                      * Use Array.prototype.slice.call() to turn nodeList into a regular array before using forEach().
                      */
                     Array.prototype.slice.call(mutation.addedNodes).forEach(node => {
-                        if (isElement(node) && node.hasAttribute('data-pp-message')) {
+                        if (!isElement(node)) {
+                            // return;
+                        } else if (node.hasAttribute('data-pp-message')) {
```
Seavenly:
```suggestion
                        if (isElement(node)) {
                            if (node.hasAttribute('data-pp-message')) {
```

Seavenly: Can we nest the if statements instead of having an empty block?

src/utils/observers.js
```diff
@@ -24,8 +24,17 @@ export const getInsertionObserver = createGlobalVariableGetter(
...
                             newMessageContainers.push(node);
+                        } else {
+                            const targetedChildNodes = node.querySelectorAll('[data-pp-message]');
+                            if (targetedChildNodes.length > 0) {
+                                Array.prototype.slice.call(targetedChildNodes).forEach(targetedChildNode => {
+                                    newMessageContainers.push(targetedChildNode);
+                                });
+                            }
```
Seavenly:
```suggestion
                           import arrayFrom from 'core-js-pure/stable/array/from'

                           arrayFrom(node.querySelectorAll('[data-pp-message]')).forEach(targetedChildNode => newMessageContainers.push(targetedChildNode));
```

Seavenly: Consider using the more declarative `arrayFrom` helper.

danzhaas: I like arrayFrom because it is much more readable.  All for this change.  There is another instance of Array.prototype.slice.call on line 26, I'd say it's within the scope of this issue and would opt to change it. What do you think?

src/utils/observers.js
danzhaas: Yes, I agree it's more readable.

src/utils/observers.js
```diff
@@ -23,9 +24,15 @@ export const getInsertionObserver = createGlobalVariableGetter(
                      * IE11 does not support nodeList.prototype.forEach().
                      * Use Array.prototype.slice.call() to turn nodeList into a regular array before using forEach().
```
Seavenly: This comment should be safe to remove now based on your change.

tests/unit/spec/src/controllers/message/setup.test.js
```diff
@@ -2,7 +2,7 @@ import setup from 'src/controllers/message/setup';
 
 import insertMockScript from 'utils/insertMockScript';
 import Messages from 'src/controllers/message/interface';
-import { getGlobalState } from 'src/utils';
+import { getGlobalState, destroyGlobalState } from 'src/utils';
 
 // TODO: Re-enable skipped tests after ramp
```
danzhaas, line 8: You had shown me how to re-enable these two skipped tests when I asked about them earlier in this story ('Renders messages on load' on line 35 and 'Renders message with script attribute options' on line 68).  Do you want me to re-enable these as well?  I don't know if they belong in this PR as I'd guess those two tests are out of the scope of this story (AC: "Improve the messaging insertion observer to check for data-pp-message elements inside the added node").

Seavenly: That's fine. You can leave them skipped for now.

tests/unit/spec/src/controllers/message/setup.test.js
```diff
@@ -18,13 +18,14 @@ describe('message setup', () => {
     afterEach(() => {
         Messages.mockClear();
         Messages().render.mockClear();
+        destroyGlobalState();
+        document.getElementsByTagName('html')[0].innerHTML = '';
```
Seavenly: These 2 lines should be able to be replaced by `destroy()` from `src/controllers/message/destroy`

Seavenly (approved):

PR #517, opened 2021-05-06
refactor: persistent deviceID storage

paypal-are:


src/components/message/Message.jsx
```diff
@@ -22,12 +22,18 @@ const Message = () => {
...
         onHover,
         onMarkup,
+        deviceID,
         resize
     } = useXProps();
     const { markup, meta, parentStyles, warnings, setServerData } = useServerData();
     const dimensionsRef = useRef({ width: 0, height: 0 });
     const buttonRef = useRef();
 
+    // If deviceID fresh use value from inside iframe, otherwise use deviceID from parent page.
```
gillycheesesteak: Let's wrap this in a useEffect so it only runs on the first render

src/utils/sdk.js
```diff
@@ -107,14 +107,48 @@ export function getSessionID() {
...
+// Separate utility for checking storageID without populating if nonexistant.
+export function readStorageID() {
+    return getRawStorage().id;
+}
+
+// Allows usage of parent page deviceID if none available within iframe
+export function writeStorageID(storageID) {
+    if (isLocalStorageEnabled()) {
+        try {
+            // Needed in order to satisfy no-unused-expressions
```
gillycheesesteak: Consider using the simpler code and just disabling the eslint rule

gillycheesesteak (approved):

PR #532, opened 2021-06-01
refactor: prerender space

jadutter:
Reserve space for message iframe using the zoid prerender iframe.


src/zoid/message/prerenderTemplate.jsx
```diff
@@ -0,0 +1,24 @@
...
+        (
+            <html>
+                <style>{styleContent}</style>
+                <body>
+                    <span>Loading...</span>
+                </body>
+            </html>
+        ).render(dom({ doc }));
+    const el = createEl({ doc });
+    return el.innerHTML;
```
Seavenly: Can this return the element reference similar to the `containerTemplate.jsx`?

src/zoid/message/prerenderTemplate.jsx
```diff
@@ -0,0 +1,24 @@
...
+        height: 100%;
+        color: #0000;
+    }
+    `;
+    const createEl = ({ doc }) =>
+        (
+            <html>
+                <style>{styleContent}</style>
+                <body>
+                    <span>Loading...</span>
```
Seavenly: I don't believe we want any content at this time, we just want the space to be reserved.

src/zoid/message/containerTemplate.jsx
```diff
@@ -48,34 +148,26 @@ export default ({ uid, frame, prerenderFrame, doc, event, props, container }) =>
...
         }
     });
     const messageTitle = getTitle(frame.title);
     return (
         <span id={uid}>
             <style>{baseStyles}</style>
-            <node el={frame} title={messageTitle} onRender={setupAutoResize} />
-            <node el={prerenderFrame} title={`Prerender ${messageTitle}`} />
+            <node class={`prerender`} el={prerenderFrame} title={`Prerender ${messageTitle}`} />
+            <node class={`render`} el={frame} title={messageTitle} onRender={setupAutoResize} />
```
Seavenly: We should refrain from using generic CSS classes on these elements because they are placed onto the merchant page where they could easily overlap with other existing merchant styles. We could instead use something like `#${uid} > iframe:nth-of-type(1)` to target those elements as needed.

src/zoid/message/containerTemplate.jsx
```diff
@@ -48,34 +148,26 @@ export default ({ uid, frame, prerenderFrame, doc, event, props, container }) =>
...
+    console.log('baseStyles', props?.style?.layout ?? 'text', baseStyles);
 
     event.on('styles', ({ styles }) => {
         if (typeof styles === 'string') {
             const style = container.querySelector(`#${uid} style`);
 
             style.textContent = `
                 ${baseStyles}
-                ${styles.replace(/\.pp-flex[^\s:,.#]*/g, `#${uid}`)}
+                ${styles.replace(/\.pp-(text|flex)[^\s:,.#]*/g, `#${uid}`)}
```
Seavenly: Do `.pp-text` styles ever come down from the server?

src/zoid/message/containerTemplate.jsx
```diff
@@ -3,12 +3,114 @@ import { node, dom } from 'jsx-pragmatic/src';
...
+        showLoadedMessage.LOADED > showLoadedMessage.RENDERED
+    ) {
+        showBanner({ uid, prerenderFrame, container });
+    }
+};
+showLoadedMessage.RENDERED = false;
+showLoadedMessage.LOADED = false;
+
 export default ({ uid, frame, prerenderFrame, doc, event, props, container }) => {
+    prerenderFrame.srcdoc = prerenderTemplate({ doc, props });
```
Seavenly: Can we use the zoid `prerenderTemplate` option instead?

src/zoid/message/containerTemplate.jsx
```diff
@@ -3,12 +3,114 @@ import { node, dom } from 'jsx-pragmatic/src';
...
+        }
+    `;
+};
+const showBanner = ({ uid, prerenderFrame, container }) => {
+    // prerenderFrame.parentNode.removeChild(prerenderFrame);
+    const style = container.querySelector(`#${uid} style`);
+    style.textContent = style.textContent.replace(/(#.+?>\s*iframe:nth-of-type\(2\)\s*\{(\n|.+?))opacity:\s*0;/g, `$1`);
+    // style.textContent = style.textContent.replace(/(#.+?>\s*iframe:nth-of-type\(1\)\s*\{)/g, `$1\n\topacity: 0;`);
+};
+const showLoadedMessage = ({ uid, prerenderFrame, container, event }) => {
```
Seavenly: Can you elaborate on what's going on here?

src/zoid/message/containerTemplate.jsx
```diff
@@ -3,12 +3,114 @@ import { node, dom } from 'jsx-pragmatic/src';
 import { EVENT } from 'zoid/src';
 
 import { getOverflowObserver, createTitleGenerator } from '../../utils';
+import prerenderTemplate from './prerenderTemplate';
 
 const getTitle = createTitleGenerator();
 
+import getParentStyles from '../../../server/getParentStyles.js';
```
Seavenly: Is this going to be used?

src/zoid/message/containerTemplate.jsx
```diff
@@ -3,12 +3,114 @@ import { node, dom } from 'jsx-pragmatic/src';
...
+            max-height: 96px;
+            min-height: 42px;
+            max-width: 768px;
+            min-width: 250px;
+        `,
+        '20x1': `
+            max-height: 59px;
+            min-height: 42px;
+            max-width: 1169px;
+            min-width: 350px;
```
Seavenly: Will these create the true ratios? Also, are the mobile ratios accounted for somewhere?

src/zoid/message/containerTemplate.jsx
```diff
@@ -3,12 +3,114 @@ import { node, dom } from 'jsx-pragmatic/src';
...
+        `
+    };
+    if (layout === 'text') {
+        const { size: textSize = 12 } = textOptions ?? {};
+        cssStyles = `
+        min-height: ${textSize * 3}px;
+        max-height: ${textSize * 6}px;
```
Seavenly: Does the ``parentHeight = `${textSize * 3}px`;`` conflict with this?

src/zoid/message/prerenderTemplate.jsx
jadutter: I remember. This is a placeholder until I've finalized what the rest of the PR should look like. It's helpful helpful for visualization purposes during the dev process.

src/zoid/message/containerTemplate.jsx
jadutter: Ah. I'll make sure to change it back to `#${uid} > iframe:nth-of-type(1)` before finalizing the PR. I found it helpful to target the iframe by class while developing as I was experimenting with the order of the elements, and I didn't want to lose track of which element the css was targeting.

src/zoid/message/containerTemplate.jsx
jadutter: I'd initially modified `getParentStyles` to produce all the prerender css and imported then imported `getParentStyles` into `containerTemplate` for use. I'll revert this change.

src/zoid/message/containerTemplate.jsx
jadutter: `showBanner` would handle the transition from "showing" the transparent prerender iframe, to showing the message banner iframe. For now, I'm accomplishing this by setting prerender and render to absolute, with the banner on top of the prerender. When we're sure the banner is ready to be shown, we remove the css option `opacity: 0`, so that the banner appears on top of the prerender iframe.

src/zoid/message/containerTemplate.jsx
jadutter: It was an early approach that I'm leaning towards not taking. I'm thinking I'll leave getParentStyles alone and use a separate function to get the css to use in containerTemplate.

src/zoid/message/containerTemplate.jsx
jadutter: The `parentHeight` is used on the element that contains the prerender and render iframes. The parent is set to a relative position, while the prerender and render iframes are set to an absolute.

src/zoid/message/prerenderTemplate.jsx
jadutter: I'd had issues getting the zoid option to work initially. I've since fixed and simplified this.

src/zoid/message/containerTemplate.jsx
jadutter: > true ratios

I'm not certain that I know what you mean by that... I pulled these values based on the behaviors I observed; how big vs how small I could make a flex banner when adjusting the screen size using the device toolbar in responsive mode. I did only check it against `DEV0000000GPL` though, so I'll check to see if there's issues for the other countries/accounts.

demo/standalone.html
```diff
@@ -4,11 +4,11 @@
         <title>Messaging.js Dev Sandbox</title>
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no" />
-        <script src="//localhost.paypal.com:8080/messaging.js" data-pp-account="DEV00000000NI"></script>
+        <meta name="description" content="Demonstrate paypal messaging components." />
     </head>
-
     <body>
         <div class="messages" data-pp-message></div>
+        <script src="//localhost.paypal.com:8080/messaging.js" data-pp-account="DEV00000000NI"></script>
```
Seavenly, line 11: Is there any particular reason we moved this to the body?

src/zoid/message/containerTemplate.jsx
```diff
@@ -70,12 +116,18 @@ export default ({ uid, frame, prerenderFrame, doc, event, props, container }) =>
             `;
         }
     });
+    event.on(EVENT.RENDERED, () => {
+        const style = container.querySelector(`#${uid} style`);
+        style.textContent = style.textContent.replace(/(#.+?iframe:nth-of-type\(2\)\{ *)display: none;/g, `$1`);
```
Seavenly: Is this safe to be removed with the state of the current CSS rules?

demo/standalone.html
```diff
@@ -4,11 +4,11 @@
         <title>Messaging.js Dev Sandbox</title>
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no" />
-        <script src="//localhost.paypal.com:8080/messaging.js" data-pp-account="DEV00000000NI"></script>
+        <meta name="description" content="Demonstrate paypal messaging components." />
```
Seavenly: If we'd like to add this, we should probably universally add it to the other demo pages as well.

demo/standalone.html
jadutter: Having it at the top was negatively impacting the [lighthouse score](https://web.dev/render-blocking-resources), since it then has to wait for the script to load before rendering any html elements.

demo/standalone.html
jadutter: This was another small change I saw I could make to improve the [lighthouse score](https://web.dev/meta-description/). I'm all for making the similar changes to the other demo pages. The standalone page just happened to be the one I was doing the CLS testing on.

src/zoid/message/containerTemplate.jsx
jadutter: Oh, thanks for pointing this out. When I took another look, I realized that removing the prerender frame was causing the css to then act on the banner iframe because of the new order. 

I'll be making a change to fix this.

src/zoid/message/containerTemplate.jsx
```diff
@@ -6,11 +6,72 @@ import { getOverflowObserver, createTitleGenerator } from '../../utils';
...
+    // when we remote the prerender iframe, we'll remove this css as well
+    const prerenderStyle = `
+        #${uid} > iframe:nth-of-type(1){
+            ${cssStyles}
+        }
+        #${uid} > iframe:nth-of-type(2){
+            display: none;
+            min-height: 10px;
+        }
+    `.replace(/[\s\n]+/g, ' ');
```
Seavenly: Does minifying this CSS in memory provide any benefit since the full string has already been bundled with the code?

src/zoid/message/containerTemplate.jsx
```diff
@@ -70,12 +119,21 @@ export default ({ uid, frame, prerenderFrame, doc, event, props, container }) =>
             `;
         }
     });
+    event.on(EVENT.RENDERED, () => {
+        // remove the css setting the prerender and hiding the banner
+        const style = container.querySelector(`#${uid} style:nth-of-type(2)`);
+        style.parentNode.removeChild(style);
```
Seavenly: Are we overcomplicating this by needing to remove styles? Could we use a more specific CSS selector like this to account for when the prerender iframe is removed?
`#${uid} > iframe:nth-of-type(1):not(:last-of-type)`

src/zoid/message/containerTemplate.jsx
jadutter: I was minifying it because of limitations I was encountering with javascript regex. It continued to be difficult and refuse to match the selector and attribute immediately after. I got around this by changing the css to single line so that the regex `.+?` would work.

src/zoid/message/containerTemplate.jsx
jadutter: I'm switching to use that selector since removing the style element caused the rendered height to bounce briefly.

src/zoid/message/containerTemplate.jsx
jadutter: This is being removed though since we're no longer changing the css and instead removing the prerender element.

src/zoid/message/containerTemplate.jsx
jadutter: Fixed by changing the css selector and only removing the prerender element.

Seavenly:
Closing for now due to inconsistent results

PR #533, opened 2021-06-02
refactor: modal prefetch removal

rosman21:
I want to have as little network requests as possible when I integrate Pay Later messages via the PP JS SDK. 

Getting the modal details on a click event now rather than prerender. 
adding a prerender frame for the modal to show the end user the modal is in a loading state


package.json
```diff
@@ -39,6 +39,7 @@
         "test:func:ciupdate": "jest --verbose --config tests/functional/jest.allFunctional.config.js --updateSnapshot"
     },
     "dependencies": {
+        "@paypal/common-components": "^1.0.18",
```
paypal-are, line 42: Did we discuss with the team before adding this new dependency?

src/zoid/message/component.js
```diff
@@ -199,7 +199,7 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
                         // Set visible to false to prevent this update from popping open the modal
                         // when the user has previously opened the modal
                         modal.updateProps({ refIndex: index, offer: offerType, visible: false });
-                        modal.render('body');
+                        // modal.render('body');
```
paypal-are: Similarly, we can remove this line rather than leave it commented.

src/controllers/modal/interface.js
```diff
@@ -19,8 +19,9 @@ import { getModalComponent } from '../../zoid/modal';
 const memoizedModal = memoizeOnProps(
     ({ account, merchantId, currency, amount, buyerCountry, offer, onReady, onCalculate, onApply, onClose }) => {
         addPerformanceMeasure('firstModalRenderDelay');
-
-        const { render, hide, updateProps, state, event } = getModalComponent()({
+        // need to remove hide from here in order for prerender frame to show until modal network request resolves
+        // const { render, hide, updateProps, state, event } ...
```
paypal-are: If these lines are unused we can remove them rather than leave them as comments.

src/controllers/modal/interface.js
```diff
@@ -19,8 +19,9 @@ import { getModalComponent } from '../../zoid/modal';
 const memoizedModal = memoizeOnProps(
     ({ account, merchantId, currency, amount, buyerCountry, offer, onReady, onCalculate, onApply, onClose }) => {
         addPerformanceMeasure('firstModalRenderDelay');
-
-        const { render, hide, updateProps, state, event } = getModalComponent()({
+        // need to remove hide from here in order for prerender frame to show until modal network request resolves
```
paypal-are: We should remove these lines if they are unnecessary rather than leave them as comments.

src/zoid/message/component.js
```diff
@@ -108,9 +108,34 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
...
+                        // possibly even render some type of message in the modal to show an error occured?
+                        if (document.querySelector(`#${modal.prerenderDetails.uid}`)) {
+                            document.querySelector(`#${modal.prerenderDetails.uid}`).style.display = 'block';
+                            modal.prerenderDetails.prerenderElement.classList.remove(
+                                `${modal.prerenderDetails.uid}-invisible`
+                            );
+                            modal.prerenderDetails.prerenderElement.classList.add(
+                                `${modal.prerenderDetails.uid}-visible`
+                            );
+                            // no way to check if prerenderer is fully loaded so wait 100ms before attempting to see if the elements exist.
```
paypal-are: Do we foresee this being an issue? It seems as though this behavior will be non-deterministic and difficult to debug in the event of an error. Is the pre-renderer guaranteed to be loaded after 100ms?

src/zoid/modal/containerTemplate.jsx
```diff
@@ -1,18 +1,41 @@
...
+    prerenderFrame.classList.add(CLASS.VISIBLE);
     event.on(EVENT.RENDERED, () => {
-        prerenderFrame.style.setProperty('display', 'none');
+        // first wait 500ms to show proper transitions
+        setTimeout(() => {
+            prerenderFrame.classList.add(CLASS.INVISIBLE);
+            frame.classList.remove(CLASS.INVISIBLE);
+            prerenderFrame.classList.remove(CLASS.VISIBLE);
+            frame.classList.add(CLASS.VISIBLE);
+        }, 500);
```
paypal-are: Are these values specified as part of the design or arbitrary wait times? Are we certain no race conditions can occur here (i.e. the transitions happen too early or late?)

src/zoid/message/component.js
rosman21: the reason this happens is because at the time of re-clicking the message the dom doesn't generate the pre-renderer yet. however there is probably a better way of handling this now that I am looking at it. Is it guranteed to load after 100? Maybe, maybe not so I do see this not being a good solution

src/zoid/modal/containerTemplate.jsx
rosman21: not specified, we just want the spinner to show for a small amount of time before the content modal actually gets loaded in. When these times are not in place the transition happens way too fast and is not pleasing at all.

paypal-are: Are we able to tie the presence of the spinner directly to the loading of the modal content (i.e. remove the spinner whenever the promise is fulfilled?)

rosman21: yea the `destroyElement` in the next line completely removes the spinner iframe after content modal is loaded.

src/zoid/modal/containerTemplate.jsx
```diff
@@ -1,18 +1,41 @@
...
 
 import { createTitleGenerator } from '../../utils';
 
 const getTitle = createTitleGenerator();
 
-export default ({ uid, frame, prerenderFrame, doc, event }) => {
+export default ({ uid, frame, prerenderFrame, doc, event, state }) => {
+    const CLASS = {
+        VISIBLE: `${uid}-visible`,
+        INVISIBLE: `${uid}-invisible`
```
paypal-are: Is there a difference between an element having class `VISIBLE` and not having class `INVISIBLE`? (i.e are we able to toggle INVISIBLE to hide elements rather than keep track of two classes?)

rosman21: not necessarily, I used checkout components as a guide. If invisible is not on the iframe it will still show. so maybe toggling invisible would make more sense.

https://github.com/paypal/paypal-checkout-components/blob/ea2b2118cd16166f22d881b9dc42a937a68929ec/src/zoid/buttons/container.jsx#L124

src/components/modal/styles/modal-frame.scss
```diff
@@ -34,17 +34,17 @@
...
-    top: 0;
-    left: 0;
-    width: 100%;
-    height: 100%;
-    background: rgba(108, 115, 120, 0.85);
-    transition: opacity 350ms;
-    opacity: 0;
-    z-index: -1;
-}
+// .overlay {
```
paypal-are: Let's get rid of these styles if we know we no longer need them.

src/zoid/message/component.js
```diff
@@ -108,9 +108,37 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
...
+                            document.querySelector(`#${modal.prerenderDetails.uid}-top`).style.opacity = 1;
+                            modal.prerenderDetails.prerenderElement.classList.remove(
+                                `${modal.prerenderDetails.uid}-invisible`
+                            );
+                            modal.prerenderDetails.prerenderElement.classList.add(
+                                `${modal.prerenderDetails.uid}-visible`
+                            );
+                            if (modal.prerenderDetails.prerenderElement.contentDocument) {
+                                // wait for prerenderer to exist
+                                ZalgoPromise.delay(100).then(() => {
```
paypal-are: Since these delay values are arbitrary, I'd prefer if these were declared as constants at the beginning of the scope so that we avoid peppering the code with magic numbers. This will allow us to easily adjust the times later if need be.

PR #537, opened 2021-06-10
refactor: remove Preact from messages

grablack:


webpack.config.js
```diff
@@ -35,7 +37,12 @@ module.exports = (env = {}) => {
         vars: globals({
             ...env,
             TARGET: 'components'
-        })
+        }),
+        plugins: [
+            new IgnorePlugin({
+                contextRegExp: /preact/
+            })
```
Seavenly: I think we need to completely separate our message and modal configs now and remove the split chunking. I think this would remove the need for this plugin then possibly.

src/components/message/index.js
```diff
@@ -1,16 +1,6 @@
...
 export function setupMessage({ markup, meta, parentStyles, warnings }) {
-    render(
-        <XPropsProvider>
-            <ServerDataProvider data={{ markup, meta, parentStyles, warnings }}>
-                <Message />
-            </ServerDataProvider>
-        </XPropsProvider>,
-        document.body
-    );
+    Message(markup, meta, parentStyles, warnings, document.body);
```
Seavenly: Instead of passing in a DOM element (body) and then mutating it, consider a more pure signature:
``` javascript
const message = createMessage({ markup, meta, parentStyles, warnings });

document.body.appendChild(message);
```

src/components/message/Message.js
```diff
@@ -0,0 +1,169 @@
+import objectEntries from 'core-js-pure/stable/object/entries';
+import { request, getActiveTags, ppDebug, getOrCreateStorageID } from '../../utils';
+
+const addToDom = (button, markup) => {
```
Seavenly: Any particular reason we've created a function instead of just conveniently using `button.innerHTML = markup`?

src/components/message/Message.js
```diff
@@ -0,0 +1,169 @@
...
+
+    const handleHover = () => {
+        if (typeof onHover === 'function') {
+            onHover({ meta });
+        }
+    };
+
+    // if there is a button reuse that instead of making a new one
+    const buttonElement =
+        frame && frame.querySelector('button') ? frame.querySelector('button') : document.createElement('button');
```
Seavenly: This function should only ever be called once upon the initial iframe setup correct? I think we should be able to safely assume that the body does not have a button element and that we need to create the main element.

src/components/message/Message.js
```diff
@@ -0,0 +1,169 @@
...
+
+    buttonElement.setAttribute('type', 'button');
+    buttonElement.setAttribute('aria-label', 'PayPal Credit Message');
```
Seavenly: Can we update this to `PayPal Pay Later Message`?

src/components/message/Message.js
```diff
@@ -0,0 +1,169 @@
...
+            deviceID: getOrCreateStorageID()
+        });
+    }
+
+    if (typeof onMarkup === 'function') {
+        onMarkup({ meta, styles: parentStyles, warnings });
+    }
+
+    if (typeof onProps === 'function') {
+        onProps(() => {
```
Seavenly: Any particular reason we are using `window.xprops` instead of the first argument which contains the xprops? If so can we leave a brief comment.

src/components/message/Message.js
```diff
@@ -0,0 +1,169 @@
...
+                .reduce(
+                    (acc, [key, val]) =>
+                        `${acc}&${key}=${encodeURIComponent(typeof val === 'object' ? JSON.stringify(val) : val)}`,
+                    ''
+                )
+                .slice(1);
+
+            ppDebug('Updating message with new props...', { inZoid: true });
+
+            request('GET', `${window.location.origin}/credit-presentment/renderMessage?${query}`).then(({ data }) => {
```
Seavenly: We should be explicit about which change in props triggers this request to prevent unnecessary render calls.

src/components/message/Message.js
```diff
@@ -0,0 +1,169 @@
...
+                        meta: data.meta ?? meta,
+                        styles: data.parentStyles ?? parentStyles,
+                        warnings: data.warnings ?? warnings
+                    });
+                }
+
+                if (!frame) {
+                    return button;
+                }
+                return frame;
```
Seavenly: Are these return values being used anywhere?

src/components/message/Message.js
```diff
@@ -0,0 +1,169 @@
...
+                .slice(1);
+
+            ppDebug('Updating message with new props...', { inZoid: true });
+
+            request('GET', `${window.location.origin}/credit-presentment/renderMessage?${query}`).then(({ data }) => {
+                button = addToDom(button, data.markup ?? markup);
+                // don't attempt append if there is no document body provided
+                if (frame) {
+                    frame.appendChild(button);
+                }
```
Seavenly: Why do we need to re-append the button element? Shouldn't it already be in the body?

src/components/message/Message.js
```diff
@@ -0,0 +1,169 @@
...
+
+                if (typeof onMarkup === 'function') {
+                    // resizes the parent message div
+                    onMarkup({
+                        meta: data.meta ?? meta,
+                        styles: data.parentStyles ?? parentStyles,
+                        warnings: data.warnings ?? warnings
+                    });
+                }
```
Seavenly: We need to be explicit about what prop changes trigger these callbacks to match the original `useLayoutEffect` dependencies.

src/components/message/Message.js
```diff
@@ -0,0 +1,169 @@
...
+            meta,
+            activeTags: getActiveTags(button),
+            // Utility will create iframe deviceID if it doesn't exist.
+            deviceID: getOrCreateStorageID()
+        });
+    }
+
+    if (typeof onMarkup === 'function') {
+        onMarkup({ meta, styles: parentStyles, warnings });
+    }
```
Seavenly: I think to better match the original Preact, we may want to wrap these in a `requestAnimationFrame` to ensure that the message has rendered inside the iframe before calling these.

src/components/message/Message.js
```diff
@@ -0,0 +1,169 @@
...
+    let buttonWidth = button.offsetWidth;
+    let buttonHeight = button.offsetHeight;
+    // Zoid will not fire a resize event if the markup has the same dimensions meaning the render event
+    // in the overflow observer will not fire. This forces the resize event to fire for every render
+    // so that the render complete logs will always fire
+    if (dimensionsRef.current.width === buttonWidth && dimensionsRef.current.height === buttonHeight) {
+        resize({ width: buttonWidth, height: buttonHeight });
+    } else {
+        dimensionsRef.current = { width: buttonWidth, height: buttonHeight };
+    }
```
Seavenly: This was originally wrapped in a `useDidUpdateLayoutEffect`, which did not run on the first render of the message. It should only be run when the props change or after a rerender request.

src/components/message/Message.js
grablack: You're right. I originally had the button being replaced. I've updated this.

src/components/message/Message.js
grablack: The return isn't being used. I've updated this to return the button to align with the pure signature comment above.

src/components/message/Message.js
grablack: I've updated this

src/components/message/Message.js
grablack: That is correct. I've updated it to only create the button element.

src/components/message/Message.js
grablack: I think I wrote it like that when I was using it in another spot and trying to debug something else and thought that was the issue. I've simplified it to button.innerHTML = markup

src/components/message/index.js
grablack: I've updated to return the button.

src/components/message/Message.js
grablack: I've changed when this happens

webpack.config.js
grablack: I split the configs and removed split chunking

src/components/message/Message.js
grablack: updated

src/components/message/Message.js
grablack: updated

webpack.config.dev.js
```diff
@@ -90,14 +90,25 @@ module.exports = (env = {}) => {
...
+        filename: '[name].js',
+        env: env.NODE_ENV,
+        vars: globals({
+            ...env,
+            TARGET: 'messagingComponent'
+        })
+    });
+
+    MESSAGING_DEV_COMPONENTS_CONFIG.entry = {
+        'smart-credit-message': './src/components/message/index.js'
```
Seavenly: We should be able to use the `entry` and `filename` options of `getWebpackConfig()`

webpack.config.js
```diff
@@ -70,11 +82,12 @@ module.exports = (env = {}) => {
     const modules = {
         library: [MESSAGES_CONFIG],
         components: [COMPONENTS_CONFIG],
-        render: [RENDERING_CONFIG]
+        render: [RENDERING_CONFIG],
+        messagingComponent: [MESSAGING_COMPONENTS_CONFIG]
```
Seavenly: For now can we still include the `MESSAGING_COMPONENTS_CONFIG` with the `components` "module" instead of creating a new `messagingComponent`?

src/components/message/Message.js
```diff
@@ -0,0 +1,204 @@
...
+    };
+
+    const paramVars = {
+        current: {
+            metaMessageRequestId: null,
+            parentStyles: null,
+            warnings: null,
+            markup: null
+        }
+    };
```
Seavenly: There is a simple `createState` helper function in our utils that we could use to simplify creating these stateful objects. Is there a reason we're nesting it inside of the `current` object property?

src/components/message/Message.js
```diff
@@ -0,0 +1,204 @@
...
+    button.style.border = 'none';
+    button.style.outline = 'none';
+    button.style.textAlign = style?.text?.align || 'left';
+    button.style.fontFamily = 'inherit';
+    button.style.fontSize = 'inherit';
+    button.innerHTML = markup;
+
+    function iframeRendered() {
+        if (typeof onReady === 'function') {
+            if (paramVars.current.metaMessageRequestId !== meta.messageRequestId) {
```
Seavenly: These 2 hooks should always fire on the first render, so these checks should be safe to remove here.

src/components/message/Message.js
```diff
@@ -0,0 +1,204 @@
...
+                onReady({
+                    meta,
+                    activeTags: getActiveTags(button),
+                    // Utility will create iframe deviceID if it doesn't exist.
+                    deviceID: getOrCreateStorageID()
+                });
+                paramVars.current.metaMessageRequestId = meta.messageRequestId;
```
Seavenly: This should be safe to set at the top with the rest of our state.

src/components/message/Message.js
```diff
@@ -0,0 +1,204 @@
...
+                {
+                    amount,
+                    currency,
+                    buyerCountry,
+                    style: JSON.stringify(window.xprops.style),
+                    offer,
+                    payerId,
+                    clientId,
+                    merchantId
+                }
```
Seavenly: This does not compare the properties of the object. This will always be `true` because it's comparing the object reference, which will be unique each time because we're creating a new object each time.
`{} !== {}`

webpack.config.dev.js
grablack: changed

webpack.config.js
grablack: moved

src/components/message/Message.js
grablack: changed to createState.

src/components/message/Message.js
grablack: removed

src/components/message/Message.js
grablack: moved

src/components/message/Message.js
grablack: forgot the stringify. I also updated to prevent listing it out like that

src/components/message/Message.js
```diff
@@ -0,0 +1,192 @@
...
+    function iframeRendered() {
+        onReady({
+            meta,
+            activeTags: getActiveTags(button),
+            // Utility will create iframe deviceID if it doesn't exist.
+            deviceID: getOrCreateStorageID()
+        });
+
+        onMarkup({ meta, styles: parentStyles, warnings });
+        updateParamVars({ ...paramVars, parentStyles, warnings, markup });
```
Seavenly: Are we able to just assign these values at the initial state declaration above?

src/components/message/Message.js
```diff
@@ -0,0 +1,192 @@
...
+
+    const [propVars, updatePropVars] = createState({
+        amount: window.xprops.amount ?? null,
+        currency: window.xprops.currency ?? null,
+        buyerCountry: window.xprops.buyerCountry ?? null,
+        style: JSON.stringify(style),
+        offer: window.xprops.offer ?? null,
+        payerId: window.xprops.payerId ?? null,
+        clientId: window.xprops.clientId ?? null,
+        merchantId: window.xprops.merchantId ?? null
```
Seavenly: Consider pulling the `window.xprops.xyz` values out above with the `const { ... } = window.xprops` destructuring.

src/components/message/Message.js
Seavenly: Any particular reason we are using `window.xprops` instead of the first argument which contains the xprops? If so can we leave a brief comment?

src/components/message/Message.js
```diff
@@ -0,0 +1,192 @@
...
+            const propObject = {};
+            Object.keys(propVars).forEach(row => {
+                if (row === 'style') {
+                    propObject[row] = JSON.stringify(window.xprops[row]) ?? null;
+                } else {
+                    propObject[row] = window.xprops[row] ?? null;
+                }
+            });
+
+            if (JSON.stringify(propVars) !== JSON.stringify(propObject)) {
```
Seavenly: Consider a more declarative approach:
``` javascript
const shouldRerender = Object.keys(propVars).some((key) =>
    typeof propVars[key] !== 'object'
        ? propVars[key] !== window.xprops[key]
        : JSON.stringify(propVars[key]) !== JSON.stringify(window.xprops[key])
);

if (shouldRerender) {
     ...   
}
```

src/components/message/Message.js
```diff
@@ -0,0 +1,192 @@
...
+                    offer,
+                    payerId,
+                    clientId,
+                    merchantId,
+                    version,
+                    env
+                } = window.xprops;
+
+                updatePropVars({
+                    ...propVars,
```
Seavenly: The update function provided by `createState()` automatically merges the supplied object with the current state object so this shouldn't be necessary.

src/components/message/Message.js
```diff
@@ -0,0 +1,192 @@
...
+        currency: window.xprops.currency ?? null,
+        buyerCountry: window.xprops.buyerCountry ?? null,
+        style: JSON.stringify(style),
+        offer: window.xprops.offer ?? null,
+        payerId: window.xprops.payerId ?? null,
+        clientId: window.xprops.clientId ?? null,
+        merchantId: window.xprops.merchantId ?? null
+    });
+
+    const [paramVars, updateParamVars] = createState({
```
Seavenly: Can we give this a more meaningful name such as `[serverData, setServerData]`?

src/components/message/Message.js
```diff
@@ -0,0 +1,192 @@
+import objectEntries from 'core-js-pure/stable/object/entries';
+import { request, getActiveTags, ppDebug, getOrCreateStorageID, createState } from '../../utils';
+
+const Message = function({ markup, meta, parentStyles, warnings }) {
+    const { onClick, onReady, onHover, onMarkup, onProps, resize, style } = window.xprops;
+    const dimensionsRef = { current: { width: 0, height: 0 } };
+
+    const [propVars, updatePropVars] = createState({
```
Seavenly: `propVars` seems a little redundant, consider just using `[props, setProps]` instead.

src/components/message/Message.js
```diff
@@ -0,0 +1,192 @@
...
+                        }
+
+                        if (typeof onReady === 'function') {
+                            onReady({
+                                meta: data.meta ?? meta,
+                                activeTags: getActiveTags(button),
+                                // Utility will create iframe deviceID if it doesn't exist.
+                                deviceID: getOrCreateStorageID()
+                            });
+                        }
```
Seavenly: In the original code it looks like this callback it looks like this has a dependency on `meta.messageRequestId` changing.

src/components/message/Message.js
```diff
@@ -0,0 +1,192 @@
...
+                                    meta: data.meta ?? meta,
+                                    styles: data.parentStyles ?? parentStyles,
+                                    warnings: data.warnings ?? warnings
+                                });
+                                updateParamVars({
+                                    ...paramVars,
+                                    parentStyles: data.parentStyles ?? parentStyles,
+                                    warnings: data.warnings ?? warnings,
+                                    markup: data.markup ?? markup
+                                });
```
Seavenly: Should this just be called regardless at the end of the promise along with updating `metaMessageRequestId` to simplify things?

src/components/message/Message.js
grablack: I didn't do it because the lint complains that the destructuring in the onProps function recreates a const with the same names.

src/components/message/Message.js
grablack: I moved it to simplify.

src/components/message/Message.js
grablack: Added

src/components/message/Message.js
grablack: updated

src/components/message/Message.js
grablack: updated

src/components/message/Message.js
grablack: removed the spread operator portions

src/components/message/Message.js
grablack: updated

src/components/message/Message.js
grablack: Yes, I assigned at the top

Seavenly (approved):

PR #545, opened 2021-06-25
refactor: modify message update call to use smart/message

danzhaas:
supplements PR of same name in CPNW


tests/unit/spec/src/components/message/Message.test.js
```diff
@@ -12,20 +12,33 @@ jest.mock('src/utils', () => ({
...
-            data: {
+            data: `<!-- {
                 markup: '<div>mock</div>',
                 meta: {
                     messageRequestId: '23456'
                 },
                 parentStyles: 'body { color: blue; }',
                 warnings: []
-            }
+            } -->
```
gillycheesesteak: This probably should be `JSON.stringify`-ed

src/components/message/Message.jsx
```diff
@@ -97,13 +97,14 @@ const Message = () => {
 
         ppDebug('Updating message with new props...', { inZoid: true });
 
-        request('GET', `${window.location.origin}/credit-presentment/renderMessage?${query}`).then(({ data }) => {
+        request('GET', `${window.location.origin}/credit-presentment/smart/message?${query}`).then(({ data }) => {
+            const dataObject = JSON.parse(data.slice(data.indexOf('<!-- ') + 5, data.indexOf(' -->')));
```
rosman21: was there a reason you changed this from the code that looked for the HTML comment? I guess my worry with this is if somehow the comment is modified with like an extra `-` or something would it throw this check off? just being very picky but also curious

demo/standalone.html
```diff
@@ -10,16 +10,16 @@
...
-            //     .Messages({
-            //         style: {
-            //             layout: 'text',
-            //             logo: {
-            //                 type: 'none'
-            //             }
-            //         }
-            //     })
-            //     .render('.messages');
+            paypal
```
rosman21: I think this should be commented out and moved back to its original code

rosman21 (changes requested):

src/components/message/Message.jsx
danzhaas: So I changed it from the code which found the re-render script in the comment by searching the HTML nodes because  that required parsing the response as an HTML document and thus running 4 non-commented scripts, which are for rendering not re-rendering.  Does that make sense?
Your concern is valid; let me think about what can be done.

danzhaas:
Will unskip and adjust Message.test.jsx as necessary upon approval of changes

rosman21 (dismissed):

rosman21:
@danzhaas looks good. just waiting for next steps, go ahead and update that test though. if you need help lmk

danzhaas:
Update: concern mentioned below is now fixed and good to go.

I see I took a step backwards here, so do not merge - fixing this now

gillycheesesteak (approved):

rosman21 (approved):

PR #546, opened 2021-07-06
refactor: including stats attributes in hover, click events

paypal-are:


src/utils/stats.js
```diff
@@ -29,6 +29,54 @@ if (!isZoidComponent()) {
...
+                    trackingDetails
+                }
+            };
+        });
+    };
+}
+
+// Function semantically similar to runStats, but returns payload to be incorporated
+// into the meta attributes of a provided event (served, hovered, click).
+export function buildStatsMeta(eventType, { container, activeTags, index }) {
```
gillycheesesteak: Should this function be used in `runStats` to avoid some repetition?

paypal-are: I'm leery of changing anything about the existing flow, but I can use this function in `runStats` if we want to.

gillycheesesteak: I think since this is basically copy-paste from that function, we can safely use the return value in triggering the `stats` event

src/utils/stats.js
```diff
@@ -29,46 +29,74 @@ if (!isZoidComponent()) {
...
-        visible: isInViewport(container).toString(),
-        // Visible message sections
-        active_tags: activeTags,
-        // Performance measurements
-        first_render_delay: Math.round(firstRenderDelay).toString(),
-        render_duration: Math.round(getCurrentTime() - state.renderStart).toString()
-    };
+    buildStatsPayload('stats', { container, activeTags, index }).then(statsPayload => {
+        // No need for scroll event if banner is above the fold
+        if (statsPayload.visible === 'false') {
```
gillycheesesteak: Because I'm unsure of the ramifications of moving this after the adblock check, lets move it outside this function and just calculate the value it needs (`visible`) again

gillycheesesteak (dismissed):

src/utils/stats.js
```diff
@@ -29,46 +29,74 @@ if (!isZoidComponent()) {
...
+// into the meta attributes of a provided event (served, hovered, click).
+export function buildStatsPayload(eventType, { container, activeTags, index }) {
     // Get outer most container's page location coordinates
     const containerRect = container.getBoundingClientRect();
     const topWindow = getTopWindow();
 
     const sdkMetaAttributes = getScriptAttributes();
 
-    const firstRenderDelay = getPerformanceMeasure('firstRenderDelay');
+    return checkAdblock().then(detected => {
```
Seavenly, line 54: Should we tweak `checkAdblock()` to return a memoized value so that it only ever runs the first time since it always takes around 250ms to run?

src/utils/stats.js
```diff
@@ -29,46 +29,74 @@ if (!isZoidComponent()) {
     });
 }
 
-export function runStats({ container, activeTags, index }) {
-    const { messagesMap } = getGlobalState();
-    const { state } = messagesMap.get(container);
+// Provide stats payload to event meta
+export function formatStatsMeta(account, { messageRequestId, trackingDetails }) {
+    return function formatPayload({ index, ...stats }) {
```
Seavenly: Would making use of the `curry()` util function be beneficial here?
``` javascript
export const addLoggerMeta = curry((account, { messageRequestId, trackingDetails }, { index, ...stats }) => {
    ...
});
```

src/zoid/message/component.js
```diff
@@ -364,6 +397,27 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
...
+            },
+            partnerAttributionId: {
+                type: 'string',
+                queryParam: true,
+                required: false,
+                value: () => (getScriptAttributes() ?? {})[SDK_SETTINGS.PARTNER_ATTRIBUTION_ID] ?? null,
+                debug: ppDebug(
+                    `Partner Attribution ID: ${(getScriptAttributes() ?? {})[SDK_SETTINGS.PARTNER_ATTRIBUTION_ID]}`
+                )
+            },
```
Seavenly, line 383: I believe this value may already be coming up with the `sdkMeta` param

src/zoid/message/component.js
```diff
@@ -138,6 +153,13 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
                             link: 'Banner Wrapper'
                         });
 
+                        // Attach stats attributes to meta payload
+                        buildStatsPayload('click', {
+                            container: getContainer(),
+                            activeTags,
+                            index
+                        }).then(formatStatsMeta(merchantId || account, meta));
```
Seavenly: Since it appears that we're wanting to include the stats as part of the meta for every request, does it make sense to immediately add it to the meta where we run `runStats()`? Would that reduce the need for adding the stats to the meta in the click and hover events?

src/utils/stats.js
```diff
@@ -29,46 +29,74 @@ if (!isZoidComponent()) {
...
+            delete existingMeta[index];
+
+            return {
+                existingMeta,
+                [index]: {
+                    type: 'message',
+                    account,
+                    messageRequestId,
+                    trackingDetails,
+                    ...statsPayload
```
gillycheesesteak: CPNW expects this under a `stats` property

src/utils/stats.js
```diff
@@ -29,46 +29,74 @@ if (!isZoidComponent()) {
...
+                }
+            };
+        });
+
+        logger.track({
+            et: 'CLIENT_IMPRESSION',
+            event_type: 'stats',
+            ...statsPayload
```
gillycheesesteak: Will this be duplicate data (in the event as well as in the meta)?

paypal-are: Yes, but I think removing this as CPNW currently stands will [lose some attributes](https://github.paypal.com/Credit-R/crcpresentmentnodeweb/blob/develop/lib/middleware/beaverLogger/index.js#L75), since they were assumed to be attached to the stats event rather than the `stats` in the meta.

gillycheesesteak: Can we explicitly include only the properties which are not currently being pulled from the meta, to avoid as much duplication as possible?

src/utils/stats.js
```diff
@@ -29,46 +29,71 @@ if (!isZoidComponent()) {
...
+    if (!isInViewport(container)) {
         getViewportIntersectionObserver().then(observer => observer.observe(container));
     }
 
-    checkAdblock().then(detected => {
-        payload.adblock = detected.toString();
-        payload.blocked = isHidden(container).toString();
-        logger.track(payload); // TODO: , container.getAttribute('data-pp-message-hidden') === 'true');
+    buildStatsPayload({ container, activeTags, index }).then(statsPayload => {
+        logger.addMetaBuilder(existingMeta => {
```
gillycheesesteak: Consider abstracting this operation of mutating the existing meta here and other places it is done using this utility function:

```
export function addLoggerMetaMutator(index, metaMutation) {
    logger.addMetaBuilder(existingMeta => {
        const newMetaForIndex = {
            ...existingMeta[index],
            ...metaMutation
        };

        // Remove potential existing meta info
        // Necessary because beaver-logger will not override an existing meta key if these values change
        // eslint-disable-next-line no-param-reassign
        delete existingMeta[index];

        return {
            [index]: newMetaForIndex
        };
    });
}
```

PR #554, opened 2021-07-12
fix: ie11 loading issue

grablack:
Fixes issue on ie11 where function undefined parameters prevent message from rendering.

### DIFF
--- a/src/utils/performance.js
+++ b/src/utils/performance.js
@@ -37,11 +37,17 @@ export function addPerformanceMeasure(name, { startMark, endMark, repeat } = {})
             performance.clearMarks(namespaced(name));
         }
 
-        performance.measure(
-            namespaced(name),
-            startMark ? namespaced(startMark) : undefined,
-            endMark ? namespaced(endMark) : undefined
-        );
+        const isIE11 = !!window.MSInputMethodContext && !!document.documentMode;
+        if (!isIE11 || (!!startMark && !!endMark)) {
+            performance.measure(
+                namespaced(name),
+                startMark ? namespaced(startMark) : undefined,
+                endMark ? namespaced(endMark) : undefined
+            );
+        } else {
+            // undefined breaks ie11
+            performance.measure(namespaced(name));
+        }
 
         performance.mark(namespaced(name));
     }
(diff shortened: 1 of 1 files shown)


src/utils/performance.js
```diff
@@ -37,11 +37,17 @@ export function addPerformanceMeasure(name, { startMark, endMark, repeat } = {})
             performance.clearMarks(namespaced(name));
         }
 
-        performance.measure(
-            namespaced(name),
-            startMark ? namespaced(startMark) : undefined,
-            endMark ? namespaced(endMark) : undefined
-        );
+        const isIE11 = !!window.MSInputMethodContext && !!document.documentMode;
```
paypal-are: We can use Belter's isIE function here:
https://github.com/krakenjs/belter/blob/8cf35443d53cb81fcb7bce0bdb385aeaffc7423c/src/device.js#L128

src/utils/performance.js
```diff
@@ -37,11 +37,17 @@ export function addPerformanceMeasure(name, { startMark, endMark, repeat } = {})
...
+        if (!isIE11 || (!!startMark && !!endMark)) {
```
paypal-are: I think this can be simplified into:

`isIE() ? performance.measure(namespaced(name)) : performance.measure(...)`. We can also remove the double negation as we don't need to cast to a boolean here.

src/utils/performance.js
```diff
@@ -37,11 +37,17 @@ export function addPerformanceMeasure(name, { startMark, endMark, repeat } = {})
...
+            performance.measure(
+                namespaced(name),
+                startMark ? namespaced(startMark) : undefined,
+                endMark ? namespaced(endMark) : undefined
+            );
+        } else {
+            // undefined breaks ie11
+            performance.measure(namespaced(name));
```
paypal-are: We should confirm that this won't introduce confusion to our performance metrics.

grablack: The addPerformanceMeasure function is not called with the startMark or endMark argument anywhere so its currently always undefined. Using performance.measure with undefined for startMark and endMark or using it without the arguments will [measure from "navigation start to now"](https://developer.mozilla.org/en-US/docs/Web/API/Performance/measure#example)
 
The issue seems to have been caused by how IE handles receiving arguments that have not been set using performance.mark(arg) 

The negation I was doing previously was to prevent measurement confusion if startMark or endMark were ever used.

paypal-are (approved):

PR #556, opened 2021-07-13
feat: store experiment treatmentsHash in browser localStorage

yeul:


src/utils/storeTreatments.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * Store hashed treatments received from server to parent window.
+ * Check to see if the treatmentsHash is present in the postMessage event. This information can be found in the
+ * meta of the message response. If present, assign the value to the variable treatmentsHash.
+ * If localStorage does not contain "treatmentsHash" OR the existing treatmentsHash does not match the value coming back
+ * from the server, use localStorage.setItem() to be the incoming treatmentsHash value.
+ */
+export const storeTreatments = () => {
+    window.addEventListener('message', event => {
```
merlinstardust: Is this a custom listener we send? If so, we should probably use a prefixed event string like `pp_message`.

yeul: Any time you use `postMessage`, the event in the browser is called a "message" so it has to be that. Similar to how a click event is called "click".

merlinstardust (dismissed):

src/zoid/message/component.js
```diff
@@ -177,12 +177,23 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
...
 
+                        /**
+                         * If treatmentsHash doesn't exist in localStorage or if the treatmentsHash coming back from
+                         * the server doesn't match what's currently in localStorage, set treatmentsHash to the incoming value.
+                         */
+                        if (
+                            !localStorage.getItem('treatmentsHash') ||
+                            treatmentsHash !== localStorage.getItem('treatmentsHash')
+                        ) {
+                            localStorage.setItem('treatmentsHash', treatmentsHash);
```
Seavenly: Should we make this a little more unique to prevent possible conflicts? Such as `__paypal_messages_treatmentsHash__`?

src/zoid/message/component.js
```diff
@@ -177,12 +177,23 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
...
                         // Write deviceID from iframe localStorage to merchant domain localStorage
                         writeStorageID(deviceID);
 
+                        /**
+                         * If treatmentsHash doesn't exist in localStorage or if the treatmentsHash coming back from
+                         * the server doesn't match what's currently in localStorage, set treatmentsHash to the incoming value.
+                         */
+                        if (
+                            !localStorage.getItem('treatmentsHash') ||
+                            treatmentsHash !== localStorage.getItem('treatmentsHash')
```
Seavenly: Can this be simplified to just `treatmentsHash !== localStorage.getItem('treatmentsHash')`? Would that also account for `!localStorage.getItem('treatmentsHash')` and prevent reading from local storage twice?

PR #564, opened 2021-07-19
feat: Add merchant configuration hash for caching

JustinDoan:
Adds an optional merchant configuration hash that can be passed to query params on message request so CDN caching can utilize it as a caching key.


src/zoid/message/validation.js
```diff
@@ -180,6 +179,17 @@ export default {
             }
         }
 
+        return undefined;
+    },
+    merchantConfig: ({ props: { merchantConfig } }) => {
```
paypal-are, line 179: Since this value is auto-generated by the SDK, I'm not sure there is any recourse for the merchant to take on this prop directly, so the console warning might be confusing.

src/utils/sdk.js
```diff
@@ -31,6 +28,15 @@ export function getEnv() {
     }
 }
 
+export function getMerchantConfig() {
+    if (__MESSAGES__.__TARGET__ === 'SDK') {
+        const fundingEligibility = getFundingEligibility();
+        return fundingEligibility?.paylater?.merchantConfigHash;
```
paypal-are: Can be condensed into `getFundingEligibility()?.paylater?.merchantConfigHash` if desired.

src/zoid/message/component.js
```diff
@@ -305,11 +304,19 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
...
-                    stringStartsWith(props.account, 'client-id:') ? props.account.slice(10) : null,
+                decorate: ({ props }) => {
+                    return stringStartsWith(props.account, 'client-id:') ? props.account.slice(10) : null;
+                },
                 default: () => '',
                 required: false
             },
+            merchantConfigHash: {
+                type: 'string',
+                queryParam: 'merchant_config',
```
paypal-are, line 314: We'll probably need to include a CPNW PR in order to [allow this param](https://github.paypal.com/Credit-R/crcpresentmentnodeweb/blob/develop/lib/validator/schema.smart.js) to be passed to our route if it's ever going to end up hitting that service.

JustinDoan: Added PR

src/zoid/message/validation.js
```diff
@@ -175,6 +174,17 @@ export default {
...
+        if (typeof merchantConfig !== 'undefined') {
```
paypal-are, line 180: We can probably remove this validation entirely since this will not come from user input.

paypal-are (approved):

PR #570, opened 2021-07-28
refactor: client-side generated unique identifier provided as messageRequestId

paypal-are:


src/zoid/message/component.js
```diff
@@ -177,7 +177,13 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
 
                     return ({ meta, activeTags, deviceID }) => {
                         const { account, merchantId, index, modal, getContainer } = props;
-                        const { messageRequestId, trackingDetails, offerType, ppDebugId } = meta;
+                        // Message request ID generated server-side
+                        const {
+                            messageRequestId: serverMessageRequestId,
```
gillycheesesteak: We should consider moving the `serverMessageRequestId` into the encrypted tracking payload so we do not have to access it explicitly on the client

danzhaas (approved):

dj110919871 (approved):

src/zoid/message/component.js
```diff
@@ -176,9 +176,11 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
                 value: ({ props }) => {
                     const { onReady } = props;
 
-                    return ({ meta, activeTags, deviceID }) => {
+                    return ({ meta, activeTags, deviceID, messageRequestId }) => {
```
gillycheesesteak: If this value doesn't change on re-renders, can we pull it from `props` below?

src/components/message/Message.js
```diff
@@ -147,17 +147,13 @@ const Message = function({ markup, meta, parentStyles, warnings }) {
...
-                                    meta: data.meta ?? meta,
-                                    activeTags: getActiveTags(button),
-                                    // Utility will create iframe deviceID if it doesn't exist.
-                                    deviceID: getOrCreateStorageID()
-                                });
-                            }
+                            onReady({
+                                meta: data.meta ?? meta,
+                                activeTags: getActiveTags(button),
+                                messageRequestId,
```
gillycheesesteak: Should we generate a new `messageRequestId` here when the message updates?

gillycheesesteak (approved):

PR #572, opened 2021-07-29
fix: overflow 0px fix 

rosman21:
this isssue is related to merchants having problems with the 
`"PayPal Message attempting fallback. Message must be visible and requires minimum dimensions of 0px x 0px. Current container is 0px x 0px."`

checks to make sure that the Root element does not have a height of 0px. If it does, it will use the next child element. 

this is a possible fix not finalized


Seavenly:
Going to work on this in #573

PR #573, opened 2021-07-30
fix: remove collapsed elements from overflow calculation

Seavenly:


src/utils/elements.js
```diff
@@ -14,12 +14,6 @@ import { ppDebug } from './debug';
  * @param {HTMLElement} el Element to check.
  * @returns boolean
  */
```
yeul: Since you removed the `elementIsOffscreen` function, we can remove this comment that was associated with it as well.

src/utils/elements.js
```diff
@@ -299,6 +293,7 @@ export function getAllBySelector(selector) {
     return [];
 }
 
+// Check if an element is completely within the bounds of another element
```
yeul: Would you mind adding a comment here for what this fn returns? (i.e. `@returns boolean`) since we do that for many of our other functions in this file.

yeul (approved):

PR #577, opened 2021-08-05
fix: address DE GPL style bugs [snapshots]

yeul:


server/message/parts/MutatedText.jsx
```diff
@@ -20,30 +20,35 @@ const MutatedText = ({ tagData, options }) => {
...
+        );
         if (match !== null) {
             match.forEach(foundString => {
                 const filteredString = foundString
                     .replace(/(\.|,)00-/g, '-')
                     .replace(/(\.|,)00$/g, '')
-                    .replace(/(\.|,)00€/g, '€');
+                    .replace(/(\.|,)00€/g, '€')
+                    .replace(/(\.|,)00(\s|\S)EUR/g, '€')
+                    .replace(/(\sEUR)/g, '€');
```
merlinstardust: On the off chance that there are multiple spaces or no spaces, it would probably be good to use the `*` for 0 or more spaces.

Why is it `\s|\S`? Won't that match all characters? So is there a reason to use `\s|\S` instead of `.`?

merlinstardust (approved):

merlinstardust (approved):

PR #582, opened 2021-08-10
feat: Use treatments to use cache param

grablack:
Receive treatments and send CDN cache feature if merchant in experiment.


grablack:
Using headers instead

PR #588, opened 2021-08-12
refactor: use separate storage key for device ID

gillycheesesteak:
After conversation with @bluepnume, we want to update the field in storage that we use so we don't disturb the original storage ID. We will attempt to use this new ID if it is available, and fall back to the original storage ID if not. Also, we want to only use the child ID if it is persistent (ie the state is not fresh)


src/components/message/Message.js
```diff
@@ -59,7 +59,7 @@ const Message = function({ markup, meta, parentStyles, warnings }) {
         meta,
         activeTags: getActiveTags(button),
         // Utility will create iframe deviceID if it doesn't exist.
-        deviceID: getOrCreateStorageID()
+        deviceID: isStorageFresh() ? null : getDeviceID()
```
rosman21: just curious but would it make more sense to just add parentDeviceId here instead of setting it to null? It looks like eventually it will be replaced later in the code but this is more just me being curious rather than a suggestion/code change

rosman21 (approved):

src/utils/sdk.js
```diff
@@ -12,13 +12,13 @@ import {
     getSDKMeta,
     getSDKAttributes,
     getSDKQueryParam,
+    getSDKStorage,
```
rosman21: library is failing because this doesn't seem to be an exported function 
https://github.com/paypal/paypal-sdk-client/blob/189b6f035845d9cd9f4fd5b804ff17c087b614ac/src/session.js#L8

rosman21 (changes requested):

rosman21 (approved):

paypal-are (approved):

PR #593, opened 2021-08-19
refactor: fix accessibility issues from test reports

grablack:
Meet WCAG 2.1 AA Guidelines


merlinstardust (changes requested):
All looks good except for a small style change around the css. There should be a space between the css target and the curly brace.

`&.sticky{` should be `&.sticky {` and so on

merlinstardust (approved):

PR #607, opened 2021-09-17
refactor: modify message update call to use smart/message

danzhaas:
supplements PR of same name in CPNW

### DIFF
--- a/src/components/message/Message.js
+++ b/src/components/message/Message.js
@@ -103,7 +103,6 @@ const Message = function({ markup, meta, parentStyles, warnings }) {
                 });
 
                 const query = objectEntries({
-                    message_request_id: meta.messageRequestId,
                     amount,
                     currency,
                     buyer_country: buyerCountry,
--- a/src/components/message/Message.js
+++ b/src/components/message/Message.js
@@ -129,8 +129,16 @@ const Message = function({ markup, meta, parentStyles, warnings }) {
 
                 request('GET', `${window.location.origin}/credit-presentment/smart/message?${query}`).then(
                     ({ data: resData }) => {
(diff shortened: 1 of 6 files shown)


utils/devServerProxy/proxy.js
```diff
@@ -187,18 +188,46 @@ export default (app, server, compiler) => {
...
+            const initializer = `
+            <script>
+            function fromBinary(binary) {
+                const bytes = new Uint8Array(binary.length);
+                for (let i = 0; i < bytes.length; i++) {
+                    bytes[i] = binary.charCodeAt(i);
+                }
+                // need to use .apply instead of spread operator so IE can understand
+                return String.fromCharCode.apply(null, new Uint16Array(bytes.buffer));
+            }
```
gillycheesesteak: Consider moving this function definition to its own script in the markup within `createMockZoidMarkup` to keep this to just the initialization code, and potentially allow it to be used elsewhere

src/components/message/Message.js
```diff
@@ -128,8 +127,18 @@ const Message = function({ markup, meta, parentStyles, warnings }) {
...
+                request('GET', `${window.location.origin}/credit-presentment/smart/message?${query}`).then(
+                    ({ data: resData }) => {
+                        function fromBinary(binary) {
+                            const bytes = new Uint8Array(binary.length);
+                            for (let i = 0; i < bytes.length; i++) {
+                                bytes[i] = binary.charCodeAt(i);
+                            }
+                            // need to use .apply instead of spread operator so IE can understand
+                            return String.fromCharCode.apply(null, new Uint16Array(bytes.buffer));
+                        }
```
gillycheesesteak: Let's move this function to the `utils` folder (maybe `miscellaneous.js`?)

PR #624, opened 2021-10-01
fix: gb gpl flex mutations [snapshots]

jadutter:
Change mutations to only use one disclaimer


Seavenly:
Changes have been previously merged.

PR #638, opened 2021-10-12
feat: add ni universal modal

jadutter:


content/modals/US/ni_um.json
```diff
@@ -0,0 +1,55 @@
...
+            [
+                "View the PayPal Credit ",
+                ["Terms and Conditions.", "https://www.paypalobjects.com/marketing/web/us/en/ppcterms/ppcterms.html"]
+            ],
+            [
+                "PayPal Credit is subject to credit approval as determined by the lender, Synchrony Bank, and is available to U.S. customers who are of legal age in their state of residence. You must pay with PayPal Credit to get the offers. Offers not valid on previous purchases, returns or exchanges. Minimum purchase required is before shipping and tax. For New Accounts: Variable Purchase APR is 23.99%. The APR is accurate as of 3/1/2021 and will vary with the market based on the Prime Rate (as defined in your credit card agreement). Minimum interest charge is $2.00."
+            ],
+            ["Copyright {fullYear} Bill Me Later, Inc. All rights reserved."]
+        ],
+        "listLink": "See other pay later options"
```
yeul: Consider changing this to ```"linkToProductList": "See other ways to pay later"``` so we can keep it unified across different views.

src/components/modal/v2/parts/ContentWrapper.jsx
```diff
@@ -21,15 +21,21 @@ const ContentWrapper = () => {
 
     // Add views to productView object where the keys are the product name and the values are the view component
     const productView = {
-        PAY_LATER_LONG_TERM: <LongTerm {...useContent(product)} />
+        PAY_LATER_LONG_TERM: <LongTerm {...useContent(product)} />,
+        NI: <NI {...useContent(product)} />
```
yeul: Once the sticky scroll change gets merged in, let's pass in the contentBodyRef.
```suggestion
        NI: <NI {...useContent(product)} contentBodyRef={contentBodyRef} />
```

src/components/modal/v2/parts/views/NI/styles/mixins.scss
```diff
@@ -0,0 +1,34 @@
...
+	'-foreground' => variable containing color for text itself
+	'-background' => variable containing color for background behind text
+	'-color' => variable containing color for an element
+*/
+
+/* Text Colors */
+// header text
+$primary-foreground: #ffffff;
+// instructions text
+$secondary-foreground: #000000;
```
yeul: For instances of colors that are simple (i.e. black or white), it might be best to just call the variables $black and $white if we use variables at all vs. just ```black``` or ```white```. We can also add them to the common ```colors.scss``` file.

src/components/modal/v2/parts/views/NI/styles/mixins.scss
```diff
@@ -0,0 +1,34 @@
...
+// terms text
+$tertiary-foreground: #6b6e76;
+
+/* Background Colors */
+// header background
+$primary-background: url("data:image/svg+xml,%3Csvg xmlns:svg='http://www.w3.org/2000/svg' xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink' width='376' height='233.812' viewBox='0 0 99.483331 61.862759' version='1.1' %3E%3Cpath id='path1037' style='fill:%23003ba4;fill-opacity:1;stroke-width:4.23333' d='M 99.483325,61.862811 0,61.862759 V 0 h 99.483322 z'/%3E%3Cpath id='path1035' style='fill:%2300318b;fill-opacity:1;stroke-width:4.23333' d='M 99.483323,14.632997 0,9.4128142 v 4.6129528 l 99.483323,6.335645 z'/%3E%3Cpath id='rect1024' style='fill:%23003698;fill-opacity:1;stroke-width:4.23333' d='m 0,14.025767 45.290411,2.884342 -22.159757,44.95265 H 0 Z'/%3E%3Cpath id='path1029' style='fill:%23004cbb;fill-opacity:1;stroke-width:4.23333' d='m 86.048221,0 h 13.435101 v 61.862812 l -43.674314,-5.3e-5 z'/%3E%3Cpath id='path1033' style='fill:%230053b8;fill-opacity:1;stroke-width:4.23333' d='m 79.410308,13.579707 20.073014,1.05329 v 5.728415 L 76.801426,18.916904 Z'/%3E%3C/svg%3E%0A");
```
yeul: We will not need this as once the product list modal is complete, as it will be content pulled from PStudio. In the meantime, this is being pulled in via the Icon component into the Header component and is shared across all views so it can be removed.

src/components/modal/v2/parts/views/NI/styles/mixins.scss
```diff
@@ -0,0 +1,34 @@
...
+// content background
+$secondary-background: #ffffff;
+// footer background
+$tertiary-background: #f5f7fb;
```
yeul: This color already exists inside of ```colors.scss``` as ```$instructions-background```.

src/components/modal/v2/parts/views/NI/styles/mixins.scss
```diff
@@ -0,0 +1,34 @@
+$content-padding: 25px;
+$line-height: 20px;
+
+/* 
+	'-foreground' => variable containing color for text itself
+	'-background' => variable containing color for background behind text
+	'-color' => variable containing color for an element
+*/
+
+/* Text Colors */
```
yeul: Consider putting color variables in the common ```colors.scss``` file rather than creating a new file.

src/components/modal/v2/lib/hooks/applyNow.js
```diff
@@ -0,0 +1,30 @@
...
+
+    const [, handleClose] = useTransitionState();
+
+    const urlBase = {
+        local: 'msmaster.qa.paypal',
+        stage: 'msmaster.qa.paypal',
+        sandbox: 'sandbox.paypal',
+        production: 'paypal'
+    }[env];
+    console.info('useApplyNow', { urlBase, payerId, env, refId });
```
yeul:
```suggestion
```
It seems like this is a leftover console.log statement.

src/components/modal/v2/parts/views/NI/Content.jsx
```diff
@@ -0,0 +1,71 @@
...
+import Icon from '../../Icon';
+import ProductListLink from '../../ProductListLink';
+import Instructions from '../../Instructions';
+import styles from './styles/index.scss';
+import { useApplyNow } from '../../../lib';
+
+export const NI = ({ instructions, terms, buttonText, disclaimer, footer, listLink, contentBodyRef }) => {
+    const buttonRef = useRef();
+    const handleApplyNowClick = useApplyNow('Apply Now');
+    const [expandedState] = useState(false);
```
yeul:
```suggestion
```
NI doesn't really make use of ```expandedState```. It was mostly meant for Pay Monthly as a means to account for the different styles when the calculator was populated with offers vs. without. We can remove this from here in this case.

src/components/modal/v2/parts/views/NI/Content.jsx
```diff
@@ -0,0 +1,71 @@
...
+
+    return (
+        <Fragment>
+            <style>{styles._getCss()}</style>
+            <div className="content__container">
+                <main className="main">
+                    <div className="content__body" ref={contentBodyRef}>
+                        <div className="content__row dynamic">
+                            <div className="content__col">
+                                <Instructions instructions={instructions} expandedState={expandedState} />
```
yeul:
```suggestion
                                <Instructions instructions={instructions} />
```
Since NI doesn't make use of ```expandedState``` we can remove this from here.

src/components/modal/v2/parts/views/NI/Content.jsx
```diff
@@ -0,0 +1,71 @@
...
+                                <div className="button__container">
+                                    <Button className="content__row" onClick={handleApplyNowClick} ref={buttonRef}>
+                                        {buttonText}
+                                    </Button>
+                                    <div className="content__row content__disclaimer">{disclaimer}</div>
+                                </div>
+                            </div>
+                            <div className={`content__col ${expandedState ? '' : 'collapsed'}`}>
```
yeul:
```suggestion
                            <div className="content__col">
```

content/modals/US/ni_um.json
```diff
@@ -0,0 +1,55 @@
...
+    }
+}
```
yeul: What are your thoughts about renaming this file ```ni_v2.json```? Eventually it will get changed to just being ```ni.json``` post-ramp, but in the meantime I think it might be more clear than ```_um``` imo.

demo/standalone-modal.html
```diff
@@ -6,12 +6,12 @@
...
         <script src="//localhost.paypal.com:8080/modal.js"></script>
     </head>
-    ​
     <body>
         <button class="learn-more">Learn more</button>
         <script>
             const modal = window.paypal.MessagesModal({
-                account: 'DEV_US_MULTI'
+                account: 'DEV_US_MULTI',
+                // account: 'DEV_US_NI',
```
yeul, line 14: Since the ```DEV_US_MULTI``` account is supposed to be multi-product based on amounts which includes NI, I think we can remove this commented out account from here.

src/components/modal/v2/parts/views/NI/Content.jsx
```diff
@@ -0,0 +1,71 @@
...
+                                            {text}
+                                        </a>
+                                    );
+                                }
+                                return <span>{item}</span>;
+                            });
+                            return <li className="content__footer-item">{line}</li>;
+                        })}
+                        <li className="content__footer-item">
+                            <ProductListLink>{listLink}</ProductListLink>
```
yeul:
```suggestion
                            <ProductListLink>{renderProductListLink()}</ProductListLink>
```
Corresponds to the above suggestion about adding a function to display or hide the link to the product list modal.

utils/devServerProxy/config/devAccountsV2.config.js
```diff
@@ -66,5 +66,32 @@ export default {
...
+        offers: {
+            PAYPAL_CREDIT_NO_INTEREST: [
+                {
+                    apr: 0,
+                    nominalRate: 0,
+                    minAmount: 99,
+                    maxAmount: 9999.99
+                }
+            ]
+        }
```
yeul: Perhaps instead of creating this entirely new dev account, we can add the appropriate information to the ```DEV_US_MULTI``` account instead. That way we avoid the similar problem we have now of too many dev accounts to keep track of.

src/components/modal/v2/parts/views/NI/styles/header.scss
```diff
@@ -0,0 +1,11 @@
...
+
+.header {
+    &__content {
+        padding-bottom: 40px;
+
+        @include mobile {
+            padding-bottom: 0px;
+        }
+    }
+}
```
yeul: Since this header file is so short now, instead of having this be an entirely new file, maybe we can put it inside of your ```main.scss``` file. We can then rename your ```main.scss``` to ```styles.scss``` and adjust the import in your ```NI/Content.jsx``` file just to try to keep a convention if possible.

src/components/modal/v2/parts/views/NI/styles/mixins.scss
```diff
@@ -0,0 +1,34 @@
...
+// header text
+$primary-foreground: #ffffff;
+// instructions text
+$secondary-foreground: #000000;
+// terms text
+$tertiary-foreground: #6b6e76;
```
yeul: This color exists in colors.scss as $legal-disclaimer so we actually won't need this in this file as well.

src/components/modal/v2/parts/views/NI/styles/mixins.scss
```diff
@@ -0,0 +1,34 @@
...
+
+/* Element Colors */
+// 'Apply Now'
+$primary-button-color: #142c8e;
+// Close
+$secondary-button-color: $primary-foreground;
+// Instructions bullet
+$primary-bullet-color: #142c8e;
+// Terms bullet
+$secondary-bullet-color: #c4c4c4;
```
yeul: It looks like only $secondary-bullet-color is a unique color from these that can be added to colors.scss since the button and bullet-related styles are now being pulled in from common.

src/components/modal/v2/parts/views/NI/styles/mixins.scss
```diff
@@ -0,0 +1,34 @@
+$content-padding: 25px;
+$line-height: 20px;
```
yeul:
```suggestion
```
Since these are no longer being used, I think it's safe to remove them

src/components/modal/v2/parts/views/NI/styles/mixins.scss
```diff
@@ -0,0 +1,34 @@
+$content-padding: 25px;
```
yeul: I left some various suggestions in this file and it seems like a couple of the colors can be moved over to the common colors.scss file. The rest of the file seems to not be necessary anymore and we should be safe to remove it after we add those. You can then import the common colors file into your main styles scss file to update the color variables as well.

src/components/modal/v2/parts/views/NI/Content.jsx
```diff
@@ -0,0 +1,71 @@
+/** @jsx h */
+import { h, Fragment } from 'preact';
+import { useState, useRef } from 'preact/hooks';
```
yeul:
```suggestion
import { useRef } from 'preact/hooks';
```
You should be safe to remove ```useState``` after removing ```expandedState```.

src/components/modal/v2/parts/views/NI/Content.jsx
```diff
@@ -0,0 +1,71 @@
...
+import Instructions from '../../Instructions';
+import styles from './styles/index.scss';
+import { useApplyNow } from '../../../lib';
+
+export const NI = ({ instructions, terms, buttonText, disclaimer, footer, listLink, contentBodyRef }) => {
+    const buttonRef = useRef();
+    const handleApplyNowClick = useApplyNow('Apply Now');
+    const [expandedState] = useState(false);
+
+    return (
```
yeul, line 21:
```suggestion
    const renderProductListLink = () => {
        if (useServerData()?.views?.length > 1) {
            return <ProductListLink>{linkToProductList}</ProductListLink>;
        }
        return <Fragment />;
    };

    return (
```
Suggesting this change here since I think it would be best to rename this ```linkToProductList``` to match Pay Monthly and Pi4 as closely as possible for similar things content-schema-wise. 

Also we can add a check here above the return that determines whether or not we should render the Product List Link component. If we have only one view returned (i.e. merchant is only eligible for just Pay in 4 or just NI), they won't see the link to the product list modal since they only get one product back.

src/components/modal/v2/parts/views/NI/Content.jsx
```diff
@@ -0,0 +1,71 @@
+/** @jsx h */
+import { h, Fragment } from 'preact';
+import { useState, useRef } from 'preact/hooks';
+import Button from '../../Button';
+import Icon from '../../Icon';
+import ProductListLink from '../../ProductListLink';
+import Instructions from '../../Instructions';
+import styles from './styles/index.scss';
+import { useApplyNow } from '../../../lib';
```
yeul:
```suggestion
import { useServerData, useApplyNow } from '../../../lib';
```
We can import `useServerData` here so that we can use it in a check for rendering the product list list or not.

PR #647, opened 2021-10-15
feat: performance benchmarking

grablack:
Outputs consumable report detailing all metrics


tools/performance/lighthouse.js
```diff
@@ -0,0 +1,191 @@
...
+ * Turns array of page scores into one averaged score per page
+ * @param {array} desktopLighthouseScores - all scores for each desktop page
+ * @param {array} mobileLighthouseScores - all scores for each mobile page
+ * @returns {object} - desktop and mobile averaged score for each page
+ */
+const getScores = (desktopLighthouseScores, mobileLighthouseScores) => {
+    const desktopAverageScores = {};
+    const mobileAverageScores = {};
+
+    Object.keys(desktopLighthouseScores).forEach(page => {
```
paypal-are: If we use `Object.entries()` here, could we refactor `getSortedAverageScores` to accept an array of scores and a page name without having to lookup the scores by the `page` key?

tools/performance/lighthouse.js
```diff
@@ -0,0 +1,191 @@
...
+        desktopAverageScores[page] = getSortedAverageScores(desktopLighthouseScores, page);
+    });
+
+    Object.keys(mobileLighthouseScores).forEach(page => {
```
paypal-are: With the refactor above, we could simplify this function into: 

```
const computeSortedAverage = scores =>
    Object.entries(scores).reduce(([page, scores]) => {
        acc[key] = getSortedAverageScores(scores, page);
        return acc;
    }, {});

return {
    desktopAverageScores: computeSortedAverage(desktopLighthouseScores),
    mobileAverageScores: computeSortedAverage(mobileLighthouseScores)
}
```

If `computeSortedAverages` is pulled out into a separate function, `getScores` can be reduced to a single line.

tools/performance/metrics.js
```diff
@@ -0,0 +1,232 @@
...
+            console.log('metrics.json failed to save');
+            console.log(err);
+        } else {
+            console.log('metrics.json saved');
+        }
+    });
+};
+
+// start puppeteer
+(async () => {
```
paypal-are: It seems like this file kicks off when it is imported via the IIFE. Is it possible to separate this behavior into a function that is initiated by a driver whenever puppeteer needs to start?

tools/performance/components.js
```diff
@@ -0,0 +1,100 @@
...
+
+    // Modals Sizes
+    const modals = getFileSizeTableRowHtml(componentsReport);
+
+    // Largest file sizes
+    const largestFiles = getFileSizeTableRowHtml(
+        [...componentsReport, ...messagesReport].sort((a, b) => b.gzipSize - a.gzipSize).splice(0, 3)
+    );
+
+    html += `<h2>File Sizes</h2>`;
```
paypal-are: We should experiment whether performance suffers significantly if we convert this html to entirely string interpolation via backticks (`) i.e:

```
`<html>
...
</html>`
```

Doing so may improve the readability of this and other HTML generation throughout the module.

tools/performance/lighthouse.js
```diff
@@ -0,0 +1,191 @@
...
+const createLighthouseHtml = (desktopAverageScores, mobileAverageScores) => {
+    const lighthouseHeadings = `<tr><td>URL</td><td>Performance</td><td>Accessibility</td><td>Best Practices</td><td>SEO</td></tr>`;
+
+    const desktopLighthouse = Object.keys(desktopAverageScores)
+        .map(
+            score =>
+                `<tr>
+                    <td>${score}</td>
+                    <td>${desktopAverageScores[score].performance.toFixed(3)}</td>
+                    <td>${desktopAverageScores[score].accessibility.toFixed(3)}</td>
```
paypal-are: Throughout the module we have virtually identical functions for `desktopAverageScores` and `mobileLighthouseScore`, could we abstract these transformations to accept an arbitrary `score` and remove this duplication?

tools/performance/lighthouse.js
```diff
@@ -0,0 +1,191 @@
+const fs = require('fs');
+
+const basePath = process.cwd();
+
+/**
+ * Get lighthouse files inside of lighthouse directory if it exists
+ * @returns {Promise} - array of lighthouse files names
+ */
+const checkDirectory = async () => {
```
paypal-are: This function is `async` but there doesn't appear to be a corresponding `await` for any asynchronous operation.

tools/performance/lighthouse.js
```diff
@@ -0,0 +1,191 @@
...
+        if (err) {
+            console.log('lighthouseScores.json failed to save');
+            console.log(err);
+        } else {
+            console.log('lighthouseScores.json saved');
+        }
+    });
+};
+
+Promise.resolve(checkDirectory()).then(files => {
```
paypal-are: Since `checkDirectory()` already returns a `Promise`, do we need this additional wrapping?

tools/performance/lighthouse.js
```diff
@@ -0,0 +1,191 @@
...
+const basePath = process.cwd();
+
+/**
+ * Get lighthouse files inside of lighthouse directory if it exists
+ * @returns {Promise} - array of lighthouse files names
+ */
+const checkDirectory = async () => {
+    return new Promise(resolve => {
+        fs.readdir(`${basePath}/lighthouse`, { encoding: 'utf8' }, (err, file) => {
+            if (err) {
```
paypal-are, line 12: We lose the error message here. If this is a critical step, should we instead `reject` the Promise and handle the error gracefully? If `checkDirectory()` returns an error are we able to continue or should the program terminate?

tools/performance/components.js
```diff
@@ -0,0 +1,100 @@
+const fs = require('fs');
+
+const basePath = process.cwd();
+
+/**
+ * Get the json data for the analyzed bundles
+ * @param {string} file - file name
+ * @returns {promise} - array of objects
+ */
+const getComponentFileData = file => {
```
paypal-are:
```suggestion
const getComponentFileData = file =>
  new Promise( ...
```

tools/performance/compile.js
```diff
@@ -0,0 +1,40 @@
...
+                        td {
+                            padding: 0.25rem 0.45rem; 
+                        }
+                    </style>
+                </head>
+                <body>
+                    <div>${new Date().toDateString()}</div>`;
+try {
+    // get html from json files
+    html += JSON.parse(fs.readFileSync(`${basePath}/dist/lighthouseScores.json`)).html;
```
paypal-are: Can these files be read in parallel or is it necessary that they be read here? Additionally, we could separate out this File I/O from by constructing `html` in its own function that accepts an arbitrary list of parsed objects and interpolates them in the body of the output.

tools/performance/lighthouse.js
grablack: added use of Object.entries()

tools/performance/lighthouse.js
grablack: I've been having an issue with this. I'm getting back the original scores combined with the accumulated data because it was an array that didn't use the url as keys. So the keys of the result ends up looking like ```
[
  '0',
  '1',
  '2',
  '3',
  '4',
  'http://..',
  'http://.../cart',
  'http://.../category/jewelry',
   ...
]
```

File: tools/performance/metrics.js (hunk 3 above)
Comment (grablack) [reply tools/performance/metrics.js]: I put it into a function. Now a environment variable is used to start it in Jenkins

File: tools/performance/lighthouse.js (hunk 5 above)
Comment (grablack) [reply tools/performance/lighthouse.js]: made an update here

File: tools/performance/lighthouse.js (hunk 6 above)
Comment (grablack) [reply tools/performance/lighthouse.js]: removed async

File: tools/performance/lighthouse.js (hunk 7 above)
Comment (grablack) [reply tools/performance/lighthouse.js]: Nope, removed

File: tools/performance/lighthouse.js (hunk 8 above)
Comment (grablack) [reply tools/performance/lighthouse.js:12]: It will currently exclude the output in the final report. I can make it throw an error and prevent the final report from being generated.

File: tools/performance/components.js (hunk 9 above)
Comment (grablack) [reply tools/performance/components.js]: updated

File: tools/performance/compile.js (hunk 10 above)
Comment (grablack) [reply tools/performance/compile.js]: I can move it to its own function. All the data is ready by the time this file runs

File: tools/performance/lighthouse.js [hunk 11]
```diff
@@ -0,0 +1,179 @@
...
+    files.forEach(file => {
+        if (file.indexOf('json') !== -1 && file.indexOf('desktop-report') !== -1) {
+            const lighthouseReport = JSON.parse(
+                fs.readFileSync(`${basePath}/lighthouse/${file}`, { encoding: 'utf8' })
+            );
+            if (!desktopScores[lighthouseReport.requestedUrl]) {
+                desktopScores[lighthouseReport.requestedUrl] = [];
+            }
+            desktopScores[lighthouseReport.requestedUrl].push({
+                url: lighthouseReport.requestedUrl,
```
paypal-are: This object the the one for mobile scores seem identical in structure. Can we factor out this code duplication?

tools/performance/lighthouse.js
```diff
@@ -0,0 +1,179 @@
...
+        });
+    });
+};
+
+/**
+ * Read raw lighthouse json files and organize by url and seperate desktop and mobile
+ * @param {array} files - array of file names
+ * @returns {object} - desktop and mobile scores grouped by url
+ */
+const groupScores = files => {
```
paypal-are, line 49: Could we decouple the file IO in this function with the computation of the scores objects? Also, does it make sense to generalize over calculating desktop and mobile scores?

tools/performance/components.js
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * Get the json data for the analyzed bundles
+ * @param {string} file - file name
+ * @returns {promise} - array of objects
+ */
+const getComponentFileData = file =>
+    new Promise(resolve => {
+        fs.readFile(`${basePath}/dist/${file}.json`, { encoding: 'utf-8' }, (err, data) => {
+            if (err) {
+                console.log(`${file} not found`);
```
paypal-are: Nitpick: `fs.readFile` can return a number of different errors even if the file exists. It might be better to log the specific error that the library returns.

tools/performance/metrics.js
```diff
@@ -0,0 +1,254 @@
...
+                    } else {
+                        stats.totalUpload += requests.size;
+                    }
+                });
+                // sort largest requests
+                const largestRequests = [...networkRequests].sort((a, b) => {
+                    return b.size - a.size;
+                });
+                // take only top 3 largest requests
+                stats.largestRequests = [...largestRequests].splice(0, 3);
```
paypal-are: Nitpick: This is equivalent to `largestRequests.slice(0, 3)`

tools/performance/lighthouse.js
grablack: made an update here

tools/performance/metrics.js
grablack: updated

tools/performance/components.js
grablack: updated

paypal-are (approved):

PR #659, opened 2021-10-20
feat: pay in 4 modal view

grablack:
Pay in 4 view


content/messages/US/short_term.json
```diff
@@ -0,0 +1,47 @@
...
+        "instructions": [
+            "Choose PayPal at checkout to pay later with <strong>Pay in 4</strong>. ",
+            "Complete your purchase with a 25% down payment.",
+            "Use autopay for the rest of your payments. It’s easy!"
+        ],
+        "disclosure": "Pay in 4 is currently not available to residents of New Mexico, North Dakota, Missouri, South Dakota, Wisconsin, or any U.S. Territories. Offer availability also depends on the merchant. Late fees may apply for missed payments depending on your state of residency. When applying, a soft credit check may be needed, but will not affect your credit score. You must be of legal age in your U.S. state of residence to use Pay in 4. Loans to California residents are made or arranged pursuant to a California Finance Lenders Law license. PayPal, Inc. is a Georgia Installment Lender Licensee, NMLS #910457.",
+        "linkToProductList": "See other ways to pay later",
+        "buttonText": "Continue with Pay Monthly"
+    }
+}
```
yeul: We actually don't need this file because we will just be using the existing pi4 message json that we already have in the ```US``` folder (i.e. gpl.json, gplq.json, etc.).

content/modals/US/short_term.json
```diff
@@ -0,0 +1,47 @@
...
+        "qualifyingSubheadline": "Split your purchase of $${transaction_amount} into {total_payments} with no impact on credit score and no late fees.",
+
+        "donutTimestamps": ["Today", "2 weeks", "4 weeks", "6 weeks"],
+
+        "termsLabel": {
+            "apr": "Est. APR",
+            "totalInterest": "Interest",
+            "total": "Total"
+        },
+        "disclaimer": "Sample payments shown are based on 15% APR and may not include shipping and taxes.",
```
yeul: These look specific to Pay Monthly right? If so, we can remove these.

content/modals/US/short_term.json
```diff
@@ -0,0 +1,47 @@
...
+        "instructions": [
+            "Choose PayPal at checkout to pay later with <strong>Pay in 4</strong>. ",
+            "Complete your purchase with a 25% down payment.",
+            "Use autopay for the rest of your payments. It’s easy!"
+        ],
+        "disclosure": "Pay in 4 is currently not available to residents of New Mexico, North Dakota, Missouri, South Dakota, Wisconsin, or any U.S. Territories. Offer availability also depends on the merchant. Late fees may apply for missed payments depending on your state of residency. When applying, a soft credit check may be needed, but will not affect your credit score. You must be of legal age in your U.S. state of residence to use Pay in 4. Loans to California residents are made or arranged pursuant to a California Finance Lenders Law license. PayPal, Inc. is a Georgia Installment Lender Licensee, NMLS #910457.",
+        "linkToProductList": "See other ways to pay later",
+        "buttonText": "Continue with Pay Monthly"
```
yeul: Since this is for Pay Monthly we can remove ```buttonText```.

content/offers/US/short_term.json
```diff
@@ -0,0 +1,47 @@
...
+        "disclaimer": "Sample payments shown are based on 15% APR and may not include shipping and taxes.",
+        "instructions": [
+            "Choose PayPal at checkout to pay later with <strong>Pay in 4</strong>. ",
+            "Complete your purchase with a 25% down payment.",
+            "Use autopay for the rest of your payments. It’s easy!"
+        ],
+        "disclosure": "Pay in 4 is currently not available to residents of New Mexico, North Dakota, Missouri, South Dakota, Wisconsin, or any U.S. Territories. Offer availability also depends on the merchant. Late fees may apply for missed payments depending on your state of residency. When applying, a soft credit check may be needed, but will not affect your credit score. You must be of legal age in your U.S. state of residence to use Pay in 4. Loans to California residents are made or arranged pursuant to a California Finance Lenders Law license. PayPal, Inc. is a Georgia Installment Lender Licensee, NMLS #910457.",
+        "linkToProductList": "See other ways to pay later",
+        "buttonText": "Continue with Pay Monthly"
+    }
```
yeul: We can remove the ```content.json``` object and just leave the meta in this file.

src/components/modal/v2/parts/views/PayInFour/styles.scss
```diff
@@ -0,0 +1,320 @@
...
+                line-height: 48px;
+                margin: 0 auto;
+            }
+        }
+
+        h2 {
+            margin-top: 0.25rem;
+            font-size: 1rem;
+            line-height: 24px;
+            font-family: PayPalSansBig, Helvetica, Arial, Sans-Serif;
```
yeul: We should be safe to remove all instances of ```font-family``` in this file since the common styles will pull in the fonts accordingly.

src/components/modal/v2/parts/views/PayInFour/styles.scss
```diff
@@ -0,0 +1,320 @@
...
+    background-color: #003087;
+    margin-top: -2rem;
+    padding-top: 2rem;
+    padding-bottom: 0;
+
+    .content__body {
+        z-index: 0;
+        position: relative;
+        top: 40px;
+        background-color: #F5F7FB;
```
yeul: We can use the ```$instructions-background``` variable here since it's in the common colors.scss file.

src/components/modal/v2/parts/views/PayInFour/styles.scss
```diff
@@ -0,0 +1,320 @@
...
+}
+
+.content__background {
+    @include desktop {
+        max-width: 884px;
+    }
+}
+
+.content__container {
+    background-color: #003087;
```
yeul: We can use the ```$dark-blue``` variable here since it's in the common colors.scss file.

src/components/modal/v2/parts/views/PayInFour/styles.scss
```diff
@@ -0,0 +1,320 @@
...
+                padding: 0 25px;
+                margin: 0 auto;
+
+                @include tablet {
+                    padding: 0 75px;
+                    max-width: initial;
+                }
+
+                @include desktop {
+                    max-width: 270px;
```
yeul: Let's actually change this to be around 340px because it allows the instructions bullets to be a little more in-line with the donuts which is what the mock shows.

src/components/modal/v2/parts/views/index.js
```diff
@@ -1 +1,2 @@
 export * from './LongTerm/Content';
+export * from './PayInFour/Content';
```
yeul: I noticed you named your folder "PayInFour". can we actually change that to be called ```ShortTerm```? Basically because not all products that are going to be using this view in other countries will be Pay in 4. For example, the UK has Pay in 3 but will be using this view.

src/components/modal/v2/parts/views/PayInFour/styles.scss
```diff
@@ -0,0 +1,320 @@
...
+                    max-width: 840px;
+                    margin: 0 auto;
+
+                    .branded-image {
+                        margin-top: 50px;
+                    }
+                }
+    
+                &.instructions {
+                    max-width: 315px;
```
yeul: We should be safe to remove all instances of ```max-width: 315px``` in this file because the common styles will fill that in for us. Removing it also matches the mock a bit more in terms of content being more in-line with itself. On mobile in particular we want the content to be a bit closer to the edge. Removing the instances of 315px helps with this.

src/components/modal/v2/parts/views/PayInFour/header.scss
```diff
@@ -0,0 +1,50 @@
...
+
+        .close {
+            svg {
+                path {
+                    stroke-width: 2.2;
+                }
+            }
+        }
+    }
+}
```
yeul: I think we can actually remove this file because currently the styles are getting pulled in from common.

src/components/modal/v2/parts/views/PayInFour/styles.scss
```diff
@@ -0,0 +1,320 @@
...
+                max-width: 577px;
+                margin: 1rem auto 0;
+            }
+         }
+    }
+}
+
+.content__background {
+    @include desktop {
+        max-width: 884px;
```
yeul: Let's actually remove this ```max-width: 884px``` because the dimensions of the modal are being pulled from common.

src/components/modal/v2/parts/views/PayInFour/styles.scss
```diff
@@ -0,0 +1,320 @@
...
+                        padding: 0 40px;
+                        position: relative;
+                        top: 30px;
+                    }
+
+                    &.donuts__non_qualifying {
+                        @include desktop {
+                            max-width: 330px;
+                        }
+                    }
```
yeul: Removing this max-width style actually makes the donuts a bit more spaced out when non-qualifying. Similar to how they are in the mock.

src/components/modal/v2/parts/views/PayInFour/styles.scss
```diff
@@ -0,0 +1,320 @@
...
+                    margin: 0 auto;
+                    padding: 0 20px;
+
+                    @include tablet {
+                        max-width: 320px;
+                    }
+
+                    @include desktop {
+                        transform: translate(0px, 1rem);
+                        padding: 0 40px;
```
yeul: I think we can actually removing this padding style here to work with some of the other suggestions to adjust the alignment of the instructions to the donuts in both qualifying/non-qualifying states.

src/components/modal/v2/parts/views/PayInFour/styles.scss
```diff
@@ -0,0 +1,320 @@
...
+
+            @include desktop {
+                font-size: 1.75rem;
+                line-height: 36px;
+                max-width: 577px;
+                margin: 1rem auto 0;
+            }
+         }
+    }
+}
```
yeul:
```suggestion
    .header__content {
        h2 {
            @include mobile {
                padding: 20px 25px 25px;
                line-height: 24px;
            }
         }
    }
}
```
The header text on desktop was looking a bit big so after playing around with it I found that removing the styles here and replacing them with the above looks a bit closer to the mock. It relies more on the common styles now with some slight tweaks for short term.

content/messages/US/short_term.json
grablack: removed

content/modals/US/short_term.json
grablack: removed

content/modals/US/short_term.json
grablack: updated

content/offers/US/short_term.json
grablack: updated

src/components/modal/v2/parts/views/index.js
grablack: updated

src/components/modal/v2/parts/views/PayInFour/styles.scss
grablack: removed

src/components/modal/v2/parts/views/PayInFour/styles.scss
grablack: removed

src/components/modal/v2/parts/views/PayInFour/header.scss
grablack: removed

src/components/modal/v2/parts/views/PayInFour/styles.scss
grablack: updated

src/components/modal/v2/parts/views/PayInFour/styles.scss
grablack: updated

src/components/modal/v2/parts/views/PayInFour/styles.scss
grablack: updated

src/components/modal/v2/parts/views/PayInFour/styles.scss
grablack: updated

src/components/modal/v2/parts/views/PayInFour/styles.scss
grablack: updated

src/components/modal/v2/parts/views/PayInFour/styles.scss
grablack: removed

src/components/modal/v2/parts/views/PayInFour/styles.scss
grablack: removed

src/components/modal/v2/parts/views/ShortTerm/Content.jsx
```diff
@@ -0,0 +1,72 @@
...
+                    <div className="content__body" ref={contentBodyRef}>
+                        <div className="dynamic__container">
+                            <div className="content__row dynamic">
+                                <div className="content__col">
+                                    <div className="content__row donuts">
+                                        <div
+                                            className={`donuts__container ${
+                                                qualifying === 'true' ? 'donuts__qualifying' : 'donuts__non_qualifying'
+                                            }`}
+                                        >
```
yeul:
```suggestion
                                            className="donuts__container">
```

I actually just got out of a 3iab with Scott and brought up that the other countries' donuts didn't have percentages inside of them and if that was intentional. A decision was made to remove the percentage from inside of the US qualifying donuts so we can simplify this class now by removing the check.

src/components/modal/v2/parts/views/ShortTerm/Content.jsx
```diff
@@ -0,0 +1,72 @@
...
+                                            {donutTimestamps.map((val, index) => (
+                                                <Donut
+                                                    qualifying={qualifying}
+                                                    periodicPayment={periodicPayment}
+                                                    currentNum={index + 1}
+                                                    timeStamp={donutTimestamps[index]}
+                                                >
+                                                    {qualifying &&
+                                                        `${((1 / donutTimestamps.length) * 100).toFixed(0)}%`}
+                                                </Donut>
```
yeul:
```suggestion
                                            {donutTimestamps.map((_, index) => (
                                                <Donut
                                                    qualifying={qualifying}
                                                    periodicPayment={periodicPayment}
                                                    currentNum={index + 1}
                                                    timeStamp={donutTimestamps[index]}
                                                />
```
A couple things here. Intellisense was telling me that the original first arg of the map (val) was declared but the value is never read. Since index is the second arg in order to make sure the donuts are showing the correct percentage filled up we can pass in a ```_``` to denote that the first arg isn't used. That way intellisense isn't yelling at us about val. 

Also since we no longer will be using the 25% 25% 25% 25%, we can remove the qualifying check here.

src/components/modal/v2/parts/views/ShortTerm/Content.jsx
```diff
@@ -0,0 +1,72 @@
...
+                                            ))}
+                                        </div>
+                                    </div>
+                                    <Instructions instructions={instructions} expandedState={expandedState} />
```
yeul:
```suggestion
                                    <Instructions instructions={instructions} />
```

This check was used in Instructions mostly for Pay Monthly when a calculator is either expanded with offers or in a collapsed state so it would rearrange a bit of the styles. Since Pay in 4 doesn't utilize this, we can remove this check here.

If you would though, can you make a change inside of ```Instructions.jsx``` and on line 4 set a default on ```expandedState``` where that prop is being passed in to be ```expandedState = false```? Just to be safe. I'll add it to the common branch too, but just in the mean time you can add it to your branch just to make sure there are no weird errors.

src/components/modal/v2/parts/ContentWrapper.jsx
```diff
@@ -19,11 +19,13 @@ const ContentWrapper = () => {
         product = useServerData().views[0].meta.product;
     }
 
-    const { headline, subheadline } = useContent(product);
+    const { headline, subheadline, qualifyingSubheadline } = useContent(product);
```
yeul:
```suggestion
    const { headline, subheadline, qualifyingSubheadline = '' } = useContent(product);
```
Let's set a default here just to be safe since other views don't have ```qualifyingSubheadline``` in their content.

src/components/modal/v2/parts/ContentWrapper.jsx
```diff
@@ -35,6 +37,8 @@ const ContentWrapper = () => {
                         logo="logo"
                         headline={headline}
                         subheadline={subheadline}
+                        isQualifying={isQualifying}
```
yeul:
```suggestion
                        isQualifying={isQualifying ?? false}
```
Another safety check here.

src/components/modal/v2/parts/views/ShortTerm/Content.jsx
```diff
@@ -0,0 +1,72 @@
...
+                                </div>
+                                <div className={`content__col ${expandedState ? '' : 'collapsed'}`}>
+                                    <div className="branded-image">
+                                        {/* TODO: update from temp desktop image */}
+                                        <Icon name="pi4-image" />
+                                    </div>
+                                </div>
+                            </div>
+                        </div>
+                        <div className={`content__row disclosure ${expandedState ? '' : 'collapsed'}`}>
```
yeul:
```suggestion
                        <div className="content__row disclosure">
```
Similar to above, we can removed this expandedState check/collapsed class here.

src/components/modal/v2/parts/views/ShortTerm/Content.jsx
```diff
@@ -0,0 +1,72 @@
...
+                                                >
+                                                    {qualifying &&
+                                                        `${((1 / donutTimestamps.length) * 100).toFixed(0)}%`}
+                                                </Donut>
+                                            ))}
+                                        </div>
+                                    </div>
+                                    <Instructions instructions={instructions} expandedState={expandedState} />
+                                </div>
+                                <div className={`content__col ${expandedState ? '' : 'collapsed'}`}>
```
yeul:
```suggestion
                                <div className="content__col">
```
We can removed this expandedState check/collapsed class here since it's not used in Pay in 4/Short term.

src/components/modal/v2/parts/Header.jsx
```diff
@@ -4,7 +4,17 @@ import { useRef, useState } from 'preact/hooks';
...
-const Header = ({ headline, subheadline, className = '', logo, contentWrapper, contentBodyRef, contentBackground }) => {
+const Header = ({
+    headline,
+    subheadline,
+    className = '',
+    logo,
+    contentWrapper,
+    contentBodyRef,
+    contentBackground,
+    isQualifying = 'false',
```
yeul, line 15:
```suggestion
    isQualifying = false,
```
Let's change this from a string false to a boolean false since the check was actually failing when looking for a string.

src/components/modal/v2/parts/Header.jsx
```diff
@@ -68,7 +78,7 @@ const Header = ({ headline, subheadline, className = '', logo, contentWrapper, c
                 </div>
                 <div className="header__content">
                     <h1>{headline}</h1>
-                    <h2>{subheadline}</h2>
+                    {isQualifying === 'true' ? <h2>{qualifyingSubheadline ?? subheadline}</h2> : <h2>{subheadline}</h2>}
```
yeul:
```suggestion
                    {isQualifying === true ? <h2>{qualifyingSubheadline ?? subheadline}</h2> : <h2>{subheadline}</h2>}
```
If we change this to a boolean true instead of a string true, the check works as expected.

src/components/modal/v2/parts/ContentWrapper.jsx
```diff
@@ -19,11 +19,13 @@ const ContentWrapper = () => {
...
 
-    const { headline, subheadline } = useContent(product);
+    const { headline, subheadline, qualifyingSubheadline } = useContent(product);
+    const { qualifying: isQualifying } = useProductMeta(product);
 
     // Add views to productView object where the keys are the product name and the values are the view component
     const productView = {
-        PAY_LATER_LONG_TERM: <LongTerm {...useContent(product)} contentBodyRef={contentBodyRef} />
+        PAY_LATER_LONG_TERM: <LongTerm {...useContent(product)} contentBodyRef={contentBodyRef} />,
+        PAY_LATER_SHORT_TERM: <ShortTerm {...useContent(product)} contentBodyRef={contentBodyRef} />
```
yeul:
```suggestion
        PAY_LATER_SHORT_TERM: <ShortTerm {...useContent(product)} {...useProductMeta(product)} contentBodyRef={contentBodyRef} />
```
If we pass in the product meta information into here, we can actually simplify the check inside of the Pi4's ```Content.jsx``` file.

src/components/modal/v2/parts/views/ShortTerm/Content.jsx
```diff
@@ -0,0 +1,72 @@
+/** @jsx h */
+import { h, Fragment } from 'preact';
+import { useState } from 'preact/hooks';
+import Icon from '../../Icon';
+import Instructions from '../../Instructions';
+import Donut from '../../Donut';
+import ProductListLink from '../../ProductListLink';
+import styles from './styles.scss';
+
+import { useServerData } from '../../../lib/providers';
```
yeul, line 10:
```suggestion
import { useServerData } from '../../../lib/providers';
```
Since this no longer will be used, we can remove it.

src/components/modal/v2/parts/views/ShortTerm/Content.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+export const ShortTerm = ({ instructions, linkToProductList, disclosure, donutTimestamps, contentBodyRef }) => {
```
yeul:
```suggestion
export const ShortTerm = ({ instructions, linkToProductList, disclosure, qualifying, periodicPayment, donutTimestamps, contentBodyRef }) => {
```
Inside of ```ContentWrapper.jsx``` I suggested we add ```{...useProductMeta(product)}``` to the Short Term component so that we can then import ```qualifying``` and  ```periodicPayment``` information here as props.

src/components/modal/v2/parts/views/ShortTerm/Content.jsx
```diff
@@ -0,0 +1,72 @@
...
+    let qualifying;
+    let periodicPayment;
+    if (useServerData()?.views?.length > 0) {
+        qualifying = useServerData().views[0].meta.qualifying;
+        periodicPayment = useServerData().views[0].meta.periodicPayment;
+    }
```
yeul:
```suggestion
```
Because we can grab the qualifying and periodic payment information from the meta (see above), we don't need this check. This also prevents the problem where we can't guarantee that the view in position ```[0]``` will be Pay in 4. Likely it will be the product list modal so we would potentially be getting back incorrect information.

src/components/modal/v2/parts/views/ShortTerm/Content.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+    const [expandedState] = useState(false);
```
yeul:
```suggestion
```
Since this file will no longer use expandedState as it is unneeded in Pi4, we can remove this as well.

src/components/modal/v2/parts/views/ShortTerm/Content.jsx
```diff
@@ -0,0 +1,72 @@
+/** @jsx h */
+import { h, Fragment } from 'preact';
+import { useState } from 'preact/hooks';
```
yeul:
```suggestion
```
Because expandedState will be removed, we can remove this import.

src/components/modal/v2/parts/views/ShortTerm/Content.jsx
```diff
@@ -0,0 +1,72 @@
...
+export const ShortTerm = ({ instructions, linkToProductList, disclosure, donutTimestamps, contentBodyRef }) => {
+    let qualifying;
+    let periodicPayment;
+    if (useServerData()?.views?.length > 0) {
+        qualifying = useServerData().views[0].meta.qualifying;
+        periodicPayment = useServerData().views[0].meta.periodicPayment;
+    }
+
+    const [expandedState] = useState(false);
+    return (
```
yeul, line 27:
```suggestion
      const renderProductListLink = () => {
        if (useServerData()?.views?.length > 1) {
            return <ProductListLink>{linkToProductList}</ProductListLink>;
        }
        return <Fragment />;
    };
    
    return (
```

We can add a check here above the return that determines whether or not we should render the Product List Link component. If we have only one view returned (i.e. merchant is only eligible for Pay in 4), they won't see the link to the product list modal since they only get one product back. You can test this by commenting out the other offers of the DEV_US_MULTI account besides short term and seeing the product list link disappear.

src/components/modal/v2/parts/views/ShortTerm/Content.jsx
```diff
@@ -0,0 +1,72 @@
...
+                                    </div>
+                                </div>
+                            </div>
+                        </div>
+                        <div className={`content__row disclosure ${expandedState ? '' : 'collapsed'}`}>
+                            <p>{disclosure}</p>
+                        </div>
+                        <div className="content__row productLink">
+                            <div className="productLink__container">
+                                <ProductListLink>{linkToProductList}</ProductListLink>
```
yeul:
```suggestion
                                <ProductListLink>{renderProductListLink()}</ProductListLink>
```
From the above comment about adding the ```renderProductListLink``` function.

src/components/modal/v2/parts/ContentWrapper.jsx
```diff
@@ -35,6 +39,8 @@ const ContentWrapper = () => {
                         logo="logo"
                         headline={headline}
                         subheadline={subheadline}
+                        isQualifying={isQualifying ?? false}
```
yeul:
```suggestion
                        isQualifying={isQualifying ?? 'false'}
```

src/components/modal/v2/parts/Header.jsx
```diff
@@ -68,7 +78,7 @@ const Header = ({ headline, subheadline, className = '', logo, contentWrapper, c
                 </div>
                 <div className="header__content">
                     <h1>{headline}</h1>
-                    <h2>{subheadline}</h2>
+                    {isQualifying === true ? <h2>{qualifyingSubheadline ?? subheadline}</h2> : <h2>{subheadline}</h2>}
```
yeul:
```suggestion
                    {isQualifying === 'true' && qualifyingSubheadline !== ''  ? <h2>{qualifyingSubheadline}</h2> : <h2>{subheadline}</h2>}
```
So it turns out it was fine to have `isQualifying` be a string instead of a boolean. It's just `qualifyingSubheadline ?? subheadline` part of the check was initially failing and changing it to a boolean "fixed" this aspect of it. Instead the above suggestion addresses both concerns.

src/components/modal/v2/parts/Header.jsx
```diff
@@ -4,7 +4,17 @@ import { useRef, useState } from 'preact/hooks';
...
-const Header = ({ headline, subheadline, className = '', logo, contentWrapper, contentBodyRef, contentBackground }) => {
+const Header = ({
+    headline,
+    subheadline,
+    className = '',
+    logo,
+    contentWrapper,
+    contentBodyRef,
+    contentBackground,
+    isQualifying = false,
```
yeul:
```suggestion
    isQualifying = 'false',
```

src/components/modal/v2/parts/ContentWrapper.jsx
grablack: updated

src/components/modal/v2/parts/Header.jsx
grablack: updated

src/components/modal/v2/parts/Header.jsx
grablack: updated

PR #660, opened 2021-10-20
feat: add support for channel prop

yeul:


src/library/zoid/message/validation.js
```diff
@@ -190,6 +190,22 @@ export default {
...
+            const options = ['UPSTREAM', 'CHECKOUT'];
+            if (validateType(Types.STRING, channel) && arrayIncludes(options, channel)) {
+                return channel;
+            }
+            if (!arrayIncludes('channel', options, channel)) {
+                logInvalidOption('channel', options, channel);
+            }
+            if (!validateType(Types.STRING, channel)) {
+                logInvalidType('channel', Types.STRING, channel);
+            }
```
Seavenly: Consider reversing the order of these and using `if`/`else` blocks to remove the need for the first `validateType(Types.STRING, channel) && arrayIncludes(options, channel)` and remove some redundant calls for bad input scenarios.

src/library/zoid/message/validation.js
```diff
@@ -190,6 +190,22 @@ export default {
...
 
+        return undefined;
+    },
+    channel: ({ props: { channel } }) => {
+        if (typeof channel !== 'undefined') {
+            const options = ['UPSTREAM', 'CHECKOUT'];
+            if (validateType(Types.STRING, channel) && arrayIncludes(options, channel)) {
+                return channel;
+            }
+            if (!arrayIncludes('channel', options, channel)) {
```
Seavenly: Does `'channel'` need to be removed here?

yeul: yup yup - fixed

Seavenly (approved):

PR #665, opened 2021-10-26
feat: common lander support

yeul:


src/components/modal/v2/parts/Modal.jsx
```diff
@@ -1,15 +1,17 @@
...
+import { TransitionStateProvider, XPropsProvider, ServerDataProvider, IntegrationProvider } from '../lib';
 
 const Modal = ({ styles, serverData, children }) => {
     return (
         <XPropsProvider>
             <ServerDataProvider data={serverData}>
                 <TransitionStateProvider>
-                    <style>{styles}</style>
-                    {children}
+                    <IntegrationProvider data={{ isLander: __MESSAGES__.__TARGET__ === 'LANDER' }}>
```
Seavenly: Introducing this as Context prevents any benefits from tree shaking unused code. Could we consider creating a simple utility that we can use to check.
``` js
// util.js
export const isLander = __MESSAGES__.__TARGET__ === 'LANDER';
```
``` js
// Component.js
import { isLander } from './util.js';
...
{isLander ? (...) : (...)}
```
If webpack can determine there is a static boolean value for a variable that is used in a conditional, it will remove the code that is impossible to get to. By placing the value into Context it is no longer deterministic for webpack.

yeul: Good call, I'll work on making this change.

PR #670, opened 2021-11-04
feat: Pay Monthly view

yeul:


src/components/modal/v2/parts/Calculator.jsx
```diff
@@ -6,9 +6,59 @@ import { useCalculator, useServerData, useXProps } from '../lib';
...
+ * This function takes in the offers returned from the server and loops over each offer
+ * in order to find the absolute min and max amounts from the offers.
+ * From the offer with the lowest min amount, the belowThreshold error content is retrieved.
+ * Similarly with the offer with the highest max amount, the aboveThreshold error content is retrieved.
+ */
+const getComputedVariables = offers =>
+    offers.reduce(
+        (acc, { meta, content }) => {
+            if (meta.minAmount) {
+                acc.minAmount = acc.minAmount
```
grablack, line 23: Is this working as expected? It looks like this will always be true. Same for line 33

yeul: So these ```minAmount``` and ```maxAmount``` reducers are actually our way of getting computed variables from our offers. Meaning that when we get a set of offers back from crcofferserv, they are not currently providing us these computed values so we have to do it ourselves at the moment. It is something we've spoken to them about and they seem open to making some changes to accommodate us, but in the mean time this is the current state. We need that absolute min and absolute max information ($199-$20,000) in order to show the correct errors and disclaimer information. For example, a 6 month offer could return a min of $199 and a max of $10000, 12 months of $1000.01-$15000, and 24 of $15000.01-$20000). So we need to reduce to grab those smallest and largest numbers of all the qualifying offers. 

This was actually a change that used to be in CPNW and the devServerProxy, but Nate and I discussed and we decided to repurpose it for this for now. You can see where it used to be here - https://github.com/Seavenly/paypal-messaging-components/blob/4f3a0495808dd418e4f2b8853d5c322d41109f5d/utils/devServerProxy/modals.js#L6-L6

PR #671, opened 2021-11-09
fix: accessibility issues 

danzhaas:


src/components/modal/lib/utils.js
```diff
@@ -30,3 +30,39 @@ export const getContent = memoize(
...
+
+export function setupTabIndex() {
+    let tabIndex;
+    const modal = document.querySelector('.modal-container');
+    const focusableElementsString =
+        "a[href], button, input, textarea, select, details, [tabindex]:not([tabindex='-1'])";
+    // eslint-disable-next-line unicorn/prefer-spread
+    const focusableElements = Array.from(modal.querySelectorAll(focusableElementsString));
+    // eslint-disable-next-line unicorn/prefer-spread
+    const tabElements = Array.from(modal.querySelectorAll('.tab'));
```
Seavenly: We should use `arrayFrom()` from `core-js-pure/stable/array/from` since `Array.from` is not supported in all browsers.

src/components/modal/lib/utils.js
```diff
@@ -30,3 +30,31 @@ export const getContent = memoize(
         );
     }
 );
+
+export function setupTabIndex() {
```
Seavenly: Would something like `setupTabTrap()` be a little more descriptive of the side-effect that is being created by calling this function?

src/components/modal/lib/utils.js
```diff
@@ -30,3 +30,31 @@ export const getContent = memoize(
...
+    let tabIndex;
```
Seavenly: `tabIndex` makes me think this variable is holding an integer value, but it's holding an array of elements. Consider having the variable name more closely reflect the type of value it holds.

src/components/modal/lib/utils.js
```diff
@@ -30,3 +30,31 @@ export const getContent = memoize(
...
+    const modal = document.querySelector('.modal-container');
+    const focusableElementsString =
+        "a[href], button, input, textarea, select, details, [tabindex]:not([tabindex='-1'])";
+    // eslint-disable-next-line unicorn/prefer-spread
+    const focusableElements = Array.from(modal.querySelectorAll(focusableElementsString));
```
Seavenly: Do we need the `modal` element selected? ince we're locked inside of our IFrame already would `document.querySelectorAll(focusableElementsString)` suffice?

Seavenly: Also, consider using `arrayFrom` from `core-js-pure/stable/array/from` since `Array.from()` is not supported in all browsers.

src/components/modal/lib/utils.js
```diff
@@ -30,3 +30,31 @@ export const getContent = memoize(
...
+
+    function trapTabKey(e) {
+        tabIndex = focusableElements.filter(node => window.getComputedStyle(node).visibility === 'visible');
```
Seavenly: This shouldn't need to happen on every single `keydown` event, it should only need to happen on every `tab` event. Consider moving this within the first `if` block.

src/components/modal/lib/utils.js
```diff
@@ -30,3 +30,31 @@ export const getContent = memoize(
...
+        // Check for TAB key press
+        if (e.keyCode === 9) {
+            // SHIFT + TAB
+            if (e.shiftKey) {
+                if (document.activeElement === tabIndex[0]) {
```
Seavenly: Consider condensing this into a single `if` statement to reduce nesting and improving readability.

src/components/modal/lib/utils.js
```diff
@@ -30,3 +31,27 @@ export const getContent = memoize(
...
+            if (e.shiftKey && document.activeElement === tabArray[0]) {
+                e.preventDefault();
+                tabArray[tabArray.length - 1].focus();
+            } else if (document.activeElement === tabArray[tabArray.length - 1]) {
+                e.preventDefault();
+                tabArray[0].focus();
+            }
+        }
+    }
+    modal.addEventListener('keydown', trapTabKey);
```
Seavenly: Would `window.addEventListener('keydown', trapTabKey);` work similarly without needing to query the DOM for a modal element?

src/components/modal/lib/utils.js
```diff
@@ -30,3 +31,27 @@ export const getContent = memoize(
...
+    let tabArray;
+    const modal = document.querySelector('.modal-container');
+    const focusableElementsString =
+        "a[href], button, input, textarea, select, details, [tabindex]:not([tabindex='-1'])";
+    const focusableElements = arrayFrom(document.querySelectorAll(focusableElementsString));
+
+    function trapTabKey(e) {
+        // Check for TAB key press
+        if (e.keyCode === 9) {
+            tabArray = focusableElements.filter(node => window.getComputedStyle(node).visibility === 'visible');
```
Seavenly: `tabArray` is never used outside of `trapTabKey()`. Consider just instantiating the variable right here instead of creating it outside of the function to lock down the scope.

Seavenly (approved):

PR #675, opened 2021-11-17
feat: universal modal

yeul:


src/components/modal/v2/lib/hooks/calculator.js
```diff
@@ -60,7 +62,8 @@ const localize = (country, amount) => {
 
 export default function useCalculator({ autoSubmit = false } = {}) {
     const calculateRef = useRef();
-    const { terms: initialTerms, country, setServerData } = useServerData();
+    const { views, country, setServerData } = useServerData();
+    const initialViewOfferTerms = views[0];
```
Seavenly: This looks like a potential point of failure with multiple views. We should consider refactoring this hook to be less "global" and more view specific where it's possible to have multiple calculators if we wanted. Perhaps we could pass in the respective view as an argument?

src/components/modal/v2/lib/hooks/calculator.js
```diff
@@ -70,12 +73,16 @@ export default function useCalculator({ autoSubmit = false } = {}) {
...
+        devTouchpoint
     } = useXProps();
+
     const [state, dispatch] = useReducer(reducer, {
-        inputValue: localize(country, initialTerms.amount),
-        prevValue: localize(country, initialTerms.amount),
-        terms: initialTerms,
+        inputValue: localize(country, amount),
+        prevValue: localize(country, amount),
+        terms: initialViewOfferTerms,
```
Seavenly: This seems a little confusing to follow since we're changing the key name from a "view" to "terms". Can we try and keep some continuity around the variable/property names as we pass them around?

src/components/modal/v2/parts/Button.jsx
```diff
@@ -0,0 +1,19 @@
+/** @jsx h */
+/* eslint-disable eslint-comments/disable-enable-pair, react/button-has-type */
+import { h } from 'preact';
+import { forwardRef } from 'preact/compat';
+
+const Button = forwardRef(({ secondary, children, size, className, type = 'button', onClick = () => {} }, ref) => {
+    const classes = ['button'];
+    if (secondary) classes.push('secondary');
+    if (size) classes.push(`${size}`);
```
Seavenly:
```suggestion
    if (size) classes.push(size);
```

src/components/modal/v2/lib/providers/xprops.js
```diff
@@ -1,8 +1,12 @@
 /** @jsx h */
 import { h, createContext } from 'preact';
 import { useState, useEffect, useContext } from 'preact/hooks';
+import polyfillZoid from '../zoid-polyfill';
+
 // Required for the lander where the modal is not loaded inside of a zoid iframe
-import '../zoid-polyfill';
+if (__MESSAGES__.__TARGET__ === 'LANDER') {
```
Seavenly: Consider using the `isLander` helper variable.

src/components/modal/v2/parts/Container.jsx
```diff
@@ -36,7 +46,7 @@ const Container = ({ children, contentWrapper }) => {
         if (typeof onReady === 'function') {
             onReady({
                 type,
-                products: products.map(({ meta: productMeta }) => productMeta.product),
+                products: views.map(({ meta: productMeta }) => productMeta.product),
```
Seavenly: This will eventually include an empty product list view so we should consider filtering against when `productMeta.product` is `undefined`.

src/components/modal/v2/parts/Container.jsx
```diff
@@ -64,7 +75,8 @@ const Container = ({ children, contentWrapper }) => {
 
     return (
         <ScrollProvider containerRef={contentWrapper}>
-            <div className="modal-wrapper">
+            <div className={`modal-wrapper ${isLander ? 'lander' : ''}`}>
+                {isLander ? <Icon name="header-background" /> : <Fragment />}
```
Seavenly: Is there a particular reason we use ternary as opposed to the common pattern of `isLander && <Icon name="header-background" />`? Is it not well supported in Preact or cause an issue with tree shaking?

src/components/modal/v2/parts/ContentWrapper.jsx
```diff
@@ -2,34 +2,49 @@
...
     const contentWrapper = useRef();
+    const contentBackground = useRef();
+    const contentBodyRef = useRef();
 
     let product;
 
-    if (useServerData()?.products?.length > 0) {
-        product = useServerData().products[0].meta.product;
+    if (useServerData()?.views?.length > 0) {
+        product = useServerData().views[0].meta.product;
```
Seavenly: Hooks should not be called conditionally. They are designed to always be called in the same order and the same amount of times. We should also be cautious about this when the multi product view enters the mix and is likely first in the list.

src/components/modal/v2/parts/ContentWrapper.jsx
```diff
@@ -2,34 +2,49 @@
...
-    const { headline, subheadline } = useContent(product);
+    const { headline, subheadline, qualifyingSubheadline = '' } = useContent(product);
+    const { qualifying: isQualifying } = useProductMeta(product);
 
     // Add views to productView object where the keys are the product name and the values are the view component
     const productView = {
-        PAY_LATER_LONG_TERM: <LongTerm {...useContent(product)} />
+        PAY_LATER_LONG_TERM: <LongTerm {...useContent(product)} contentBodyRef={contentBodyRef} />,
+        PAY_LATER_SHORT_TERM: (
+            <ShortTerm {...useContent(product)} {...useProductMeta(product)} contentBodyRef={contentBodyRef} />
```
Seavenly: Consider calling each hook once and passing the data from the singular calls into each view component to reduce redundant calls. We should also consider using common props amongst all views such as `content={...} meta={...}` instead of spreading a bunch of props onto each component.

src/components/modal/v2/parts/Donut.jsx
```diff
@@ -0,0 +1,78 @@
...
+    style = { fontSize: '0.375rem' },
+    children,
+    segmentStrokeWidth = 4,
+    currentNum = 0,
+    numOfPayments = 4,
+    timeStamp,
+    periodicPayment,
+    qualifying
+}) => {
+    const { percentage } = generateDonutData(currentNum, numOfPayments);
```
Seavenly: Consider simplifying this since the function is called once and contains minimal logic which leads to unnecessary jumping around the code while reading through it.
```suggestion
    const percentage = (currentNum / numOfPayments) * 100;
```

src/components/modal/v2/parts/Donut.jsx
```diff
@@ -0,0 +1,78 @@
...
+    let segStrokeWidth = segmentStrokeWidth;
+    if (!segmentStrokeWidth) {
+        segStrokeWidth = strokeWidth;
+    }
```
Seavenly:
```suggestion
    const segStrokeWidth = segmentStrokeWidth ?? strokeWidth;
```

src/components/modal/v2/parts/Header.jsx
```diff
@@ -1,33 +1,99 @@
...
+            stickyScroll(22);
+        });
+    }
+
+    // Sticky header on scroll for desktop and tablet
+    if (contentBackground.current) {
+        contentBackground.current.addEventListener('scroll', () => {
+            stickyScroll(150);
+        });
+    }
```
Seavenly: Attaching event listeners is a side effect which should likely be run inside of a `useEffect()` hook, otherwise we are relying on an unintended rerender from outside of this component since refs that hold DOM elements like `contentWrapper.current` will always be empty for the first render.

src/components/modal/v2/parts/Icon.jsx
```diff
@@ -52,14 +52,111 @@ const Icon = ({ name, color }) => {
                     />
                 </svg>
             );
-        case 'pay-monthly-image':
+        case 'pay-monthly-temp-image':
```
Seavenly: Is this change needed?

src/components/modal/v2/parts/Icon.jsx
```diff
@@ -52,14 +52,111 @@ const Icon = ({ name, color }) => {
...
             return (
                 <img
-                    alt="temporary"
+                    alt="pp-lt"
                     style={{ width: '100%' }}
-                    src="https://preview.redd.it/7c2gaheiowu31.png?auto=webp&s=5581d6fd2c0769db69c5e71e1d9938866aabb7fc"
+                    src="https://www.paypalobjects.com/marketing/web/us/en/credit-line-and-card-services/new/pp-credit-and-cards-hero.png"
```
Seavenly: These images should be pulled from the downstream content response

src/components/modal/v2/parts/Icon.jsx
```diff
@@ -52,14 +52,111 @@ const Icon = ({ name, color }) => {
...
                 />
             );
+        case 'pi4-image':
+            return <img alt="temporary" style={{ width: '100%' }} src="https://placehold.jp/450x550.png" />;
```
Seavenly: Similar as above

src/components/modal/v2/parts/LoadingShimmer.jsx
```diff
@@ -1,13 +1,17 @@
...
+     */
     return (
         <Fragment>
-            {/* TODO: revist map over offers */}
-            {[0, 1, 2].map(() => {
+            {/* eslint-disable-next-line unicorn/prefer-spread */}
+            {Array.from(Array(numOffers).keys()).map(() => {
                 return (
-                    <div className={`offer__container shimmer ${hasError ? 'has-error' : ''}`}>
+                    <div className="offer__container shimmer">
```
Seavenly: We should import `arrayFrom` from `core-js-pure/stable/array/from` since it's not supported in all browsers. `Array.from` also supports creating an array of certain length. We should also consider using the `key` attribute when creating elements in a loop to reduce unnecessary rerendering.
``` js
{arrayFrom({ length: numOffers }).map((_, index) => {
    return (
        <div className="offer__container shimmer" key={index}>
```

src/components/modal/v2/parts/OfferCard.jsx
```diff
@@ -1,27 +1,29 @@
 /** @jsx h */
 import { h } from 'preact';
 
-const OfferCard = ({ termsLabel, offer, index }) => {
+const OfferCard = ({ offer: { content, meta }, index }) => {
+    const { termsLabel } = content;
+    const aprRemoveTrailingZeros = meta?.apr.replace(/([^\d]00)(?=[^\d]|$)/, '');
```
Seavenly: Will `meta.apr` ever not end with a digit? Can this regex be simplified to just `/\D00$/`?

src/components/modal/v2/parts/OfferCard.jsx
```diff
@@ -1,27 +1,29 @@
...
+
     return (
         <div className={`offer__container ${index}`}>
```
Seavenly: Do we need the index as a class name? If this is for CSS are we able to use the `:nth-of-type(x)` selector?

src/components/modal/v2/parts/ProductListLink.jsx
```diff
@@ -0,0 +1,33 @@
...
+import { forwardRef } from 'preact/compat';
+
+const openProductList = () => {
+    // eslint-disable-next-line no-console
+    console.warn('TODO: ProductListLink to change view to Product List');
+};
+
+const openProductListViaKeyboard = () => {
+    openProductList();
+};
```
Seavenly: Do we need this function or can `onClick` and `onKeyUp` share the same function?

src/components/modal/v2/parts/TermsTable.jsx
```diff
@@ -1,29 +1,37 @@
...
+
     if (isLoading || hasError) {
         return (
             <Fragment>
                 <div className="offer__wrapper">
-                    <LoadingShimmer hasError={hasError} />
+                    <LoadingShimmer numOffers={numOffers} />
                 </div>
-                <div className="finance-terms__disclaimer">{disclaimer}</div>
             </Fragment>
```
Seavenly: Is it safe to remove the `<Fragment>` now?

src/components/modal/v2/parts/TermsTable.jsx
```diff
@@ -1,29 +1,37 @@
...
         );
     }
-    const sortedOffers = offers.slice().sort((a, b) => b.term - a.term);
+
+    const sortedOffers = offers
+        .slice()
+        .sort((a, b) => b.term - a.term)
```
Seavenly: Is sorting necessary? If so I don't think we should be doing it upstream. We should be getting the correct order from downstream so that the order is the same in checkout as well.

src/components/modal/v2/parts/TermsTable.jsx
```diff
@@ -1,29 +1,37 @@
...
+        .filter(offer => offer.meta.qualifying === 'true')
```
Seavenly, line 23: If we `filter` as the first part of this chain it will create a new array for us with needing to `slice` and reduce loop iterations for all later loops

src/components/modal/v2/parts/views/LongTerm/Content.jsx
```diff
@@ -1,40 +1,76 @@
...
+                <Fragment>
+                    <div className="button__container">
+                        <Button onClick={() => onClose({ linkName: 'Pay Monthly Continue' })} className="cta">
+                            {cta.buttonText}
+                        </Button>
+                    </div>
+                </Fragment>
+            );
+        }
+        if (useServerData()?.views?.length > 1 && typeof cta === 'undefined') {
```
Seavenly: Hooks should not be called conditionally and can lead to unexpected result. Consider calling all hooks at the top of the component similar to defining dependencies. Also `typeof cta === 'undefined'` should be safe to remove because getting passed the first `if` block guarantees that.

src/components/modal/v2/parts/views/LongTerm/Content.jsx
```diff
@@ -1,40 +1,76 @@
...
-                                    termsLabel={termsLabel}
                                     disclaimer={disclaimer}
+                                    buttonText={buttonText}
                                 />
                             </div>
                             <div className={`content__col ${expandedState ? '' : 'collapsed'}`}>
                                 <div className="branded-image">
                                     {/* TODO: update from temp desktop image */}
-                                    <Icon name="pay-monthly-image" />
+                                    <Icon name="pay-monthly-temp-image" />
```
Seavenly: Is this safe to remove?

src/components/modal/v2/parts/views/ShortTerm/Content.jsx
```diff
@@ -0,0 +1,69 @@
...
+    instructions,
+    linkToProductList,
+    disclosure,
+    donutTimestamps,
+    qualifying,
+    periodicPayment,
+    contentBodyRef
+}) => {
+    const renderProductListLink = () => {
+        if (useServerData()?.views?.length > 1) {
```
Seavenly: Similar comment as above. Consider calling and storing hook results at the top of the component for every render as a dependency to increase readability and follow best practices.

src/components/modal/v2/parts/views/ShortTerm/Content.jsx
```diff
@@ -0,0 +1,69 @@
...
+                                    <div className="content__row donuts">
+                                        <div className="donuts__container">
+                                            {donutTimestamps.map((val, index) => (
+                                                <Donut
+                                                    qualifying={qualifying}
+                                                    periodicPayment={periodicPayment}
+                                                    currentNum={index + 1}
+                                                    timeStamp={donutTimestamps[index]}
+                                                />
+                                            ))}
```
Seavenly: Consider adding the `key` attribute to follow best practices when creating components in a loop.

src/components/modal/v2/styles/globals/_mixins.scss
```diff
@@ -1,15 +1,28 @@
...
+//     @media (min-width: 1366px) {
+//         @content;
+//     }
+// }
+
+// @mixin tablet {
+//     @media (min-width: 640px) and (max-width: 1365px) {
+//         @content;
+//     }
+// }
```
Seavenly: Is this safe to remove?

src/components/modal/v2/lib/hooks/calculator.js
```diff
@@ -186,10 +196,17 @@ export default function useCalculator({ autoSubmit = false } = {}) {
...
+    if (state.inputValue === '0') {
+        isLoading = false;
+    }
+
     return {
         terms: state.terms,
-        value: state.inputValue,
-        isLoading: state.isLoading,
+        // Replace start of value string that isn't a digit (i.e. if someone tries to enter a period or a comma first) with an empty string.
+        value: state.inputValue.replace(/^([^$\d]|\.|,)/g, ''),
```
Seavenly: Can this regex be simplified to `/^\D/`?

utils/devServerProxy/modals.js
```diff
@@ -0,0 +1,140 @@
...
+            populateTemplate(template, {
+                ...morsVars,
+                ...otherVars
+            })
+        );
+
+        if (offers) {
+            viewData.offers = offers.map(offer => JSON.parse(populateTemplate(offer.template, offer.morsVars)));
+            // viewData.offers is required
+            viewData.meta.computed = getComputedVariables(viewData);
```
Seavenly: Is this safe to remove now since we pivoted to doing this client side?

src/components/modal/v2/parts/Header.jsx
```diff
@@ -1,33 +1,101 @@
...
+            });
+        }
+
+        // Sticky header on scroll for desktop and tablet
+        if (contentBackground.current) {
+            contentBackground.current.addEventListener('scroll', () => {
+                stickyScroll(150);
+            });
+        }
+    }, [sticky]);
```
Seavenly: Are we able to use our current `ScrollProvider` and `useScroll` hook to achieve the same thing here? This looks like it has the potential to keep adding event listeners without cleaning them up which could get expensive. The provider and hook should nicely handle setting/cleaning up scroll event listeners in a more declarative way.

yeul: Improvements to sticky header/scroll behavior to be addressed via [#681](https://github.com/paypal/paypal-messaging-components/pull/681).

PR #678, opened 2021-11-22
refactor: offer parameter universal modal

grablack:
offer type parameters


src/utils/miscellaneous.js
```diff
@@ -215,22 +219,27 @@ export function getProductForOffer(offer) {
...
+            return OFFER.PAY_LATER_PAY_IN_1;
         case 'EZP':
         case 'EZP:ANY:EQZ':
         case 'EZP:ANY:GTZ':
         case 'PALA:MULTI:EQZ':
         case 'PALA:MULTI:GTZ':
         case 'PALA:SINGLE:EQZ':
         case 'PALA:SINGLE:GTZ':
-            return 'EZP';
+            return OFFER.PAYPAL_CREDIT_INSTALLMENTS;
```
Seavenly: This line looks like it can be removed since it will match with the next group which has the same return value;

src/components/modal/v2/parts/ContentWrapper.jsx
```diff
@@ -12,12 +13,18 @@ const ContentWrapper = () => {
...
+    const { offer, integrationType } = useXProps();
     const [product, setProduct] = useState();
     const content = useContent(product);
     const productMeta = useProductMeta(product);
 
     if (views?.length > 0) {
-        setProduct(views[0].meta.product);
+        const viewsContainsOffer = views.some(viewOffers => viewOffers.meta.product === offer);
+        if (!viewsContainsOffer && integrationType === 'STANDALONE_MODAL') {
+            validate.offer({ props: { offer } });
```
Seavenly: Why do we need to validate the `offer` xprop? It should have been already been pre-validated from zoid and we aren't injecting this value into the DOM.

src/components/modal/v2/parts/ContentWrapper.jsx
```diff
@@ -12,12 +13,18 @@ const ContentWrapper = () => {
...
+        }
+        setProduct(viewsContainsOffer && offer ? offer : views[0].meta.product);
```
Seavenly: This will lead to double renders initially because the initial `useState()` value is `undefined` and then we immediate update the state. Does `product` need to be a state variable? It appears that it's computed off of `offer` so can we just set it as a regular variable and then pass that directly into our `useContent()` and `useProductMeta()` hooks?

``` js
const product = views.find(view => view.meta.product === offer)?.meta.product ?? views[0].meta.product;
const content = useContent(product);
const productMeta = useProductMeta(product);
```

We should also note that `views[0].meta.product` may need to be updated when the product list view is introduced.

src/library/zoid/message/component.js
```diff
@@ -126,7 +136,7 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
                             amount,
                             buyerCountry,
                             onApply,
-                            offer: offerType,
+                            offer: offer ?? offerType,
```
Seavenly: We want to ensure that the value being passed into `offer` is one of our new supported values. Consider wrapping this in `getProductForOffer()`.

src/library/zoid/message/component.js
```diff
@@ -223,7 +234,7 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
                             activeTags,
                             index
                         });
-                        modal.updateProps({ refIndex: index, offer: offerType });
+                        modal.updateProps({ refIndex: index, offer: offer ?? offerType });
```
Seavenly: Similar comment as above

src/library/zoid/message/validation.js
```diff
@@ -108,10 +108,20 @@ export default {
...
+            'PAYPAL_CREDIT_INSTALLMENTS',
+            'GPL',
+            'NI'
+        ];
+
         if (typeof offer !== 'undefined') {
             if (!validateType(Types.STRING, offer)) {
                 logInvalidType('offer', Types.STRING, offer);
-            } else if (offer !== 'NI') {
+            } else if (!arrayIncludes(options, offer)) {
```
Seavenly: This validation currently only runs with a message render which currently does not support these newer values yet. We may want to leave this as is for now until we add support for these values through the message flow.

src/components/modal/parts/Tabs.jsx
```diff
@@ -9,7 +9,7 @@ import { getProductForOffer } from '../../../utils';
 // Determine pre-selected tab based on the offer type of the banner.
 const getInitialTabIndex = (offer, tabs) =>
     Math.max(
-        arrayFindIndex(tabs, ({ product }) => product === getProductForOffer(offer)),
+        arrayFindIndex(tabs, ({ product }) => getProductForOffer(product) === getProductForOffer(offer)),
```
Seavenly: If we ensure the `offer` input into the modal is one our new values, would we be able to simplify this to `getProductForOffer(product) === offer`?

src/library/zoid/modal/component.js
```diff
@@ -265,6 +267,35 @@ export default createGlobalVariableGetter('__paypal_credit_modal__', () =>
...
+                                supportedProduct => supportedProduct === requestedProduct
+                            )
+                        ) {
+                            logger.warn('invalid_option_value', {
+                                location: 'offer',
+                                description: `Expected one of ["${[...products, ...newAttributes].join(
+                                    '", "'
+                                )}"] but received "${props.offer}".`
+                            });
+                        }
```
Seavenly: Can you provide some context why this validation was moved here from `src/library/controllers/modal/interface.js`?

grablack: This was an artifact from the previous story. I'll restore

src/utils/miscellaneous.js
grablack: updated

src/components/modal/v2/parts/ContentWrapper.jsx
grablack: updated

src/components/modal/v2/parts/ContentWrapper.jsx
grablack: updated

src/library/zoid/message/component.js
grablack: updated

src/library/zoid/message/component.js
grablack: updated

src/components/modal/parts/Tabs.jsx
grablack: Yes we can simplify. I'll need to update Tabs.test.js since the offer provided is no longer updated.

src/components/modal/content/US/parts/Content.jsx
```diff
@@ -22,9 +22,11 @@ const Content = ({ headerRef, contentWrapper }) => {
     const handleApplyNowClick = useApplyNow('Apply Now');
     const [showApplyNow, setApplyNow] = useState(false);
     // Offer may be undefined when modal is rendered via standalone modal integration
-    const product = getProductForOffer(offer);
+    const product = getStandardProductOffer(offer);
```
Seavenly: If we're standardizing the `offer` param now outside the IFrame right before it gets passed in, is this now safe to remove? Can we just use the raw `offer` value instead of creating the `product` variable?

src/components/modal/content/US/parts/Content.jsx
```diff
@@ -22,9 +22,11 @@ const Content = ({ headerRef, contentWrapper }) => {
...
     // Product can be NONE when standalone modal so default to first product
-    const initialProduct = arrayFind(products, prod => prod.meta.product === product) || products[0];
+    const initialProduct =
+        arrayFind(products, prod => (offer ? prod.meta.product === offer : prod.meta.product === product)) ||
```
Seavenly: `offer` and `product` should be the exact same value now since they are both being sent through `getStandardProductOffer()`. `prod.meta.product` could be an older value though so when comparing to `offer` we should probably use `getStandardProductOffer(prod.meta.product) === offer`.

src/utils/miscellaneous.js
```diff
@@ -215,22 +219,26 @@ export function getProductForOffer(offer) {
...
         case 'INST:ANY:GTZ':
         case 'PALAQ:ANY:EQZ':
         case 'PALAQ:ANY:GTZ':
-            return 'INST';
+            return OFFER.PAYPAL_CREDIT_INSTALLMENTS;
+        case OFFER.PAYPAL_CREDIT_NO_INTEREST:
+        case 'NI':
         default:
-            return 'NI';
+            return OFFER.PAYPAL_CREDIT_NO_INTEREST;
```
Seavenly, line 241: Let's consider changing the `default` case to return `undefined` since that would mean the supplied value is not one of our supported cases and would help reduce confusion in the future with our multi-product world. It historically defaulted to NI since that was part of our fallback logic, but I believe the modals now have other fallback logic to more appropriately account for this.

src/components/modal/content/US/parts/Content.jsx
grablack: There was another spot using product, but I updated it similarly.

src/components/modal/content/US/parts/Content.jsx
grablack: updated

src/utils/miscellaneous.js
grablack: updated

PR #679, opened 2021-11-23
fix: fix-iframe-renders

jadutter:
Bug: when message banners
1. are in an iframe 
1. the iframe they are in is currently outside viewport
1. the message would be 100% visible if the page was reloaded at the correct scroll height 
then currently, the observer hides the message. 

The goal of this PR is to fix it so that those iframes outside the viewport will be visible when the user scrolls.


Seavenly:
Closing due to staleness since the issue has not been prioritized

PR #682, opened 2021-12-02
refactor: remove renderMessage (re-adding from dtcgemi-32)

rosman21:


src/components/message/Message.js
```diff
@@ -144,8 +152,10 @@ const Message = function({ markup, meta, parentStyles, warnings }) {
 
                 ppDebug('Updating message with new props...', { inZoid: true });
 
-                request('GET', `${window.location.origin}/credit-presentment/renderMessage?${query}`).then(
-                    ({ data }) => {
+                request('GET', `${window.location.origin}/credit-presentment/smart/message?${query}`).then(
+                    ({ data: resData }) => {
+                        const encodedData = resData.slice(resData.indexOf('<!--') + 4, resData.indexOf('-->'));
```
gillycheesesteak, line 157: I think in the original PR we updated this to use a builtin DOM accessor (something like `document.firstChild`)

rosman21: Let me take a look one more time to see if I missed that.

gillycheesesteak: Regardless, let's see if that is a viable option to be more explicit

rosman21: The `resData` value is coming back as a string so the document obj won't exist there. I think what you are referring to is [this](https://github.paypal.com/Credit-R/crcpresentmentnodeweb/blob/ded55783c6b26408901ba5e85bd75d09050c07f5/lib/scriptsLoader.js#L104) . I will see if there is a better string manipulation function we can use to do this in a different way possibly.

gillycheesesteak (approved):

PR #684, opened 2021-12-06
feat: channel and integration identifer logging

grablack:
Log when a consumer views or interacts with a modal.


src/library/zoid/message/component.js
```diff
@@ -214,7 +214,8 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
                                     type: 'message',
                                     messageRequestId,
                                     account: merchantId || account,
-                                    trackingDetails
+                                    trackingDetails,
+                                    channel: channel ?? 'UPSTREAM'
```
Seavenly: Can we add the channel prop to the props list like [the modal](https://github.com/paypal/paypal-messaging-components/blob/58df1de8d4c133bcd2590f57c351157c25b55c34/src/library/zoid/modal/component.js#L111)? And we should consider adding the default value by adding `default: () => 'UPSTREAM'` to the prop descriptors similar to [this](https://github.com/paypal/paypal-messaging-components/blob/58df1de8d4c133bcd2590f57c351157c25b55c34/src/library/zoid/modal/component.js#L302)?

src/library/zoid/modal/component.js
```diff
@@ -243,7 +251,9 @@ export default createGlobalVariableGetter('__paypal_credit_modal__', () =>
                                     type: 'modal',
                                     messageRequestId,
                                     account: merchantId || account,
-                                    trackingDetails
+                                    trackingDetails,
+                                    channel: channel ?? 'UPSTREAM',
```
Seavenly: Similar to the above comment. Consider using the default prop descriptor to handle this.

src/library/zoid/message/component.js
grablack: updated

src/library/zoid/modal/component.js
grablack: updated

Seavenly (approved):

PR #687, opened 2021-12-21
feat: measure cdn stats (DTCRCGEMI-194)

jadutter:
Measure the time it takes for an iframe to load its first script,


src/utils/performance.js
```diff
@@ -2,7 +2,48 @@ import { getPerformance, isIE } from 'belter/src';
 
 const namespaced = name => `__paypal_messaging_performance__${name}`;
 
-const performance = getPerformance();
+export const performance = getPerformance();
+
+export const PERFORMANCE_MEASURE_KEYS = {
```
jadutter, line 7: I wanted to make it more clear on what we are tracking by ensuring the items are enumerated here. 

While checking what we are tracking, I noticed some things were listed as deprecated and made a note in the comments. 

This map does not include the renderStart metric we are collecting. I would have liked to have refactored it to be included in this map, but I needed to begin work on the PR for cpnw.

src/utils/performance.js
```diff
@@ -2,7 +2,50 @@ import { getPerformance, isIE } from 'belter/src';
...
+            ({ name, entryType }) =>
+                (entryType === 'navigation' && `${name}`.indexOf('/credit-presentment/smart/message') > -1) ||
+                (entryType === 'resource' && `${name}`.indexOf('/credit-presentment/renderMessage') > -1)
+        );
+
+    const [{ connectStart, responseStart }] = requests.slice(-1);
+
+    if (typeof connectStart !== 'undefined') {
+        // This measures the "Waiting (Time To First Byte)" for the request;
+        // how long we've spent waiting for a response after sending the request
```
jadutter: The value recorded here will match the value in
> Dev Tools > Network > Select the request > Timing > Waiting (TTFB)

src/utils/performance.js
```diff
@@ -2,7 +2,50 @@ import { getPerformance, isIE } from 'belter/src';
...
+    // eslint-disable-next-line compat/compat
+    const requests = window.performance
+        .getEntries()
+        .filter(
+            ({ name, entryType }) =>
+                (entryType === 'navigation' && `${name}`.indexOf('/credit-presentment/smart/message') > -1) ||
+                (entryType === 'resource' && `${name}`.indexOf('/credit-presentment/renderMessage') > -1)
+        );
+
+    const [{ connectStart, responseStart }] = requests.slice(-1);
```
gillycheesesteak: This could cause a destructuring error if `requests` is somehow empty

src/utils/performance.js
```diff
@@ -11,12 +54,12 @@ export function getPerformanceMeasure(name) {
 export function getNavigationTiming(name) {
     const entry = performance?.getEntriesByType('navigation')[0];
 
-    return entry?.[name] ?? -1;
+    return Math.round(entry?.[name] ?? -1).toString();
```
gillycheesesteak: should this use the new `formatStat` function?

src/utils/performance.js
gillycheesesteak: Is `responseEnd` available here? That would encompass the time to download the content as well

gillycheesesteak (approved):

PR #690, opened 2021-12-27
feat: expand modal for api iframes

merlinstardust:


src/components/modal/v2/styles/components/_header.scss
```diff
@@ -77,6 +82,10 @@
         svg {
             width: 900px;
             transform: translate(0, -30%);
+
+            @include apiIframe {
+                width: 100%;
```
grablack: Looks like this width 100% is causing that transparent space on the lander iframe. If we can't use 900px we probably need to also have transform: translate(0, -35%); to cover the transparent space.

src/components/modal/v2/parts/Container.jsx
```diff
@@ -77,9 +77,16 @@ const Container = ({ children }) => {
         });
     }, [currency, amount, payerId, clientId, merchantId, buyerCountry]);
 
+    const modalWrapperClassName = [
+        'modal-wrapper',
+        isLander && !isIframe ? 'lander' : '',
+        loading ? 'loading' : '',
+        isLander && isIframe ? 'api-iframe' : ''
+    ].join(' ');
```
grablack, line 90: There are some extra spaces in the class. We can filter the empty responses.
`].filter(Boolean).join(' ');`

grablack (changes requested):

src/components/modal/v2/parts/Container.jsx
merlinstardust: There's a couple other places in the library (like `v2/parts/Button`) where we use a `.join` without a filter since extra spaces don't make a difference, so I'd like to leave it as is

grablack (approved):

merlinstardust:
US modal tests are failing because "Special Financing" bullet was removed.

src/components/modal/v2/parts/BodyContent.jsx
```diff
@@ -49,7 +49,10 @@ const BodyContent = () => {
 
     useDidUpdateEffect(() => {
         scrollTo(0); // Reset scroll position to top when view changes
-        window.document.querySelector('#close-btn').focus();
+        const closeButton = window.document.querySelector('#close-btn');
+        if (closeButton) {
+            closeButton.focus();
+        }
```
Seavenly: Thoughts on `window.document.querySelector('#close-btn')?.focus();`?

src/components/modal/v2/parts/views/NI/styles.scss
```diff
@@ -91,6 +89,11 @@
                     @include fixedWrapper;
                     bottom: 0;
                 }
+
+                @include apiIframe {
+                    max-width: $content-wrapper-max-width-api-iframe;
```
Seavenly: Can we instead include this in the `fixedWrapper` mixin with the other `max-width` definitions?

src/components/modal/v2/parts/views/NI/styles.scss
```diff
@@ -106,6 +109,10 @@
                     .button {
                         width: calc(80%);
                     }
+
+                    @include apiIframe {
+                        max-width: $content-wrapper-max-width-api-iframe;
+                    }
```
Seavenly: Can we update the 2 `max-width: 640px` values in the tablet and mobile section to be `$content-wrapper-max-width-tablet`? That was likely an oversight before. And then should this be up one level and underneath both the `@include tablet` and `@include mobile` so that it follows the pattern of our other areas where we choose between the 3 `max-width` variables?

src/components/modal/v2/styles/components/_body-content.scss
```diff
@@ -18,6 +18,12 @@
     @include desktop {
         max-width: $content-wrapper-max-width-desktop;
     }
+
+    @include apiIframe {
+        max-width: $content-wrapper-max-width-api-iframe;
+        margin: 0;
```
Seavenly: Is this `margin: 0` necessary since it looks to be `auto` above and the full width would prevent that from being any sort of issue.

src/components/modal/v2/styles/components/_header.scss
```diff
@@ -27,6 +27,11 @@
         @include mobile {
             border-radius: 0;
         }
+
+        @include apiIframe {
+            max-width: $content-wrapper-max-width-api-iframe;
```
Seavenly: Similar comment above where I think this style should ideally live inside the `fixedWrapper` mixin with the others.

src/components/modal/v2/styles/components/_header.scss
```diff
@@ -117,6 +127,14 @@
                     justify-content: center;
                 }
 
+                @include apiIframe {
+                    transform: translate(calc(#{$content-wrapper-max-width-api-iframe} / 2 - 50%), 2rem);
+
+                    &--scroll {
+                        transform: translate(0, 0);
+                    }
```
Seavenly, line 135: Can we leave a brief comment above this that we're duplicating this style here because the `apiIframe` mixin has a higher priority than the original scroll below?

PR #694, opened 2022-01-03
fix: accessibility improved in header

danzhaas:


danzhaas:
Will suggest this be added to newly opened accessibility issues story GEMI-234 instead of GEMI-188 which has already been released

PR #701, opened 2022-01-11
fix: accessibility issues including reverse tabbing and nav link focus

danzhaas:


src/components/modal/content/US/parts/Content.jsx
```diff
@@ -74,11 +75,21 @@ const Content = ({ headerRef, contentWrapper }) => {
         }
     };
 
+    function setFocusAfterLinkClick() {
+        const focusableElementsString =
+            "a[href], button, input, textarea, select, details, [tabindex]:not([tabindex='-1'])";
+        const tabArray = arrayFrom(document.querySelectorAll(focusableElementsString)).filter(
+            node => window.getComputedStyle(node).visibility === 'visible'
+        );
```
Seavenly: Since we're using this portion of code across multiple files now would it be helpful to turn this into a simple helper function such as `getFocusableElements()`? We could put it inside of `src/utils/elements.js`.

src/components/modal/content/US/parts/Content.jsx
```diff
@@ -74,11 +75,21 @@ const Content = ({ headerRef, contentWrapper }) => {
...
+        tabArray[tabArray.length - 1].focus();
+    }
+
     // Tabs component will track tab switches by default
     // for "fake" tabs that show as links, we must track it manually
     const tabLinkClick = newProduct => {
         onClick({ linkName: newProduct, src: 'link_click' });
         selectProduct(newProduct);
+        setTimeout(() => setFocusAfterLinkClick(), 0);
```
Seavenly: Is the `setTimeout()` to account for delaying the call so that `preact` can switch the selected product and the new focusable elements are in view? If so we should consider using the built in `preact` hooks to account for lifecycle changes. We can't guarantee that `setTimeout(..., 0)` will always fire after `preact` re-renders. Instead we should consider watching for the change to `selectedProduct`.
``` jsx
useDidUpdateEffect(() => {
    setFocusAfterLinkClick()
}, [selectedProduct]);
```

Seavenly (approved):

PR #706, opened 2022-01-16
ci: simplify test paths

merlinstardust:


.github/scripts/getTestPathPatterns.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# lists all functional test paths and then removes text leaving only {locale}/{type}/{subtype}
+validTestPaths=$(ls ./tests/functional/spec/*/*/*.test.js | sed -E 's/(.*spec\/|.test.js)//g')
```
Seavenly: Could we make this a little more path generic with something like this, while also tightening up the `sed` a bit?
``` bash
find ./tests/functional/spec -name '*.test.js' | grep -v non-snapshot-tests | sed -E 's/(^.*spec\/|.test.js$)//g'
```

.github/scripts/getTestPathPatterns.sh
```diff
@@ -0,0 +1,8 @@
...
+# transforms the list into a comma separated json array
+testPathPatterns="[ \"$(echo $validTestPaths | tr '\n' ',' | sed 's/ /,/g;s/,$//;s/,/","/g' )\" ]"
```
Seavenly, line 8: Is the first `s/ /,/g;` able to be removed? It looks like it may not be matching anything.

merlinstardust: If you run this on a Mac `tr '\n' ','` handles the newlines and turns them into commas. When I tested it in the CI, `tr` did nothing because it was already a single line. So `sed 's/ /,/g` handles that case. It's safer to have both and the only downside is a little extra code

Seavenly: Would you mind adding that as a brief comment in the script as well?

Seavenly (approved):

PR #728, opened 2022-02-04
fix: prevent false show events

Seavenly:


src/components/modal/v2/styles/globals/_common.scss
```diff
@@ -7,6 +7,9 @@ body {
 
 html {
     min-width: inherit !important;
+    // Overscroll background color since we use an SVG pattern
+    // for the core background which moves with the overscroll
+    background: #1040c1;
```
grablack: The background color is showing up on standalone modal overlay. I think we need to add it to modal-wrapper instead.

Seavenly: Good catch. The background color needs to be on the root `html` element, so I refactored where the classes get applied so that it's applied conditionally.

grablack (approved):

PR #733, opened 2022-02-16
feat: ES modal, instrumentation, and func tests

danzhaas:


src/components/modal/content/ES/index.js
```diff
@@ -1,15 +0,0 @@
...
-import ContentWrapper from './parts/ContentWrapper';
-
-export function setupModal(props) {
-    render(
-        <Modal serverData={props} styles={styles._getCss()}>
-            <ContentWrapper contentMaxWidth={750} contentMaxHeight={537} />
-        </Modal>,
-        document.body
-    );
-}
```
danzhaas, line 15: @gillycheesesteak This looks harmless but unusual to me. I found the library breaks if this file or its directory (ES) is deleted, but the library runs fine when the file is empty.  I couldn't find exactly what made it crash, so I left it empty and it works.  Any concerns?

gillycheesesteak (approved):

PR #734, opened 2022-02-16
fix: corrects product button icon position

danzhaas:


danzhaas:
fix included in hotfix #735

PR #739, opened 2022-02-18
fix: ensure callback functions work

jadutter:
Changes: 
* Attaching `data-pp-onapply` will execute the defined callback function
* Adding `onApply` to `paypal.Message` config will execute the defined callback function
* Added functional tests to confirm callback events are triggered accordingly

### DIFF
--- a/demo/blank.html
+++ b/demo/blank.html
@@ -4,7 +4,6 @@
         <title>Messages Dev Sandbox</title>
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no" />
-        <!-- <script src="//localhost.paypal.com:8080/sdk/js?client-id=DEV00000000NI&components=messages"></script> -->
         <script src="//localhost.paypal.com:8080/messaging.js" data-pp-account="DEV00000000NI"></script>
     </head>
     <body></body>
--- a/demo/blank.html
+++ b/demo/blank.html
@@ -1,0 +1,11 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <title>Messages Dev Sandbox</title>
+        <meta charset="UTF-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no" />
+        <!-- <script src="//localhost.paypal.com:8080/sdk/js?client-id=DEV00000000NI&components=messages"></script> -->
+        <script src="//localhost.paypal.com:8080/messaging.js" data-pp-account="DEV00000000NI"></script>
+    </head>
+    <body></body>
+</html>
--- a/tests/functional/spec/non-snapshot-tests/onEvents.test.js
+++ b/tests/functional/spec/non-snapshot-tests/onEvents.test.js
@@ -29,72 +29,70 @@ const getContext = () => {
 
 // Add a message banner with attached callbacks
 const renderMessage = async method => {
-    switch (method) {
-        case 'inline attribute':
-            return page.evaluate(async () => {
-                const {
-                    onEvents,
-                    config: {
-                        account,
-                        amount,
-                        style: { layout }
-                    }
-                } = await getContext();
+    if (method === 'inline attribute') {
+        return page.evaluate(async () => {
+            const {
+                onEvents,
+                config: {
+                    account,
+                    amount,
+                    style: { layout }
+                }
+            } = await getContext();
 
-                const div = document.createElement('div');
-                div.setAttribute('data-pp-account', account);
-                div.setAttribute('data-pp-amount', amount);
-                div.setAttribute('data-pp-style-layout', layout);
+            const div = document.createElement('div');
+            div.setAttribute('data-pp-account', account);
+            div.setAttribute('data-pp-amount', amount);
+            div.setAttribute('data-pp-style-layout', layout);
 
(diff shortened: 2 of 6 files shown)


tests/functional/spec/non-snapshot-tests/onEvents.test.js
```diff
@@ -0,0 +1,186 @@
...
+                    inValue: `console.log(\`${eventMessage}\`)`
+                }
+            };
+        }, {})
+    };
+};
+
+// Add a message banner with attached callbacks
+const renderMessage = async method => {
+    switch (method) {
```
gillycheesesteak: should we use `if..else` here?

tests/functional/spec/non-snapshot-tests/onEvents.test.js
```diff
@@ -0,0 +1,186 @@
...
+                        account,
+                        amount,
+                        style: { layout }
+                    }
+                } = await getContext();
+
+                const div = document.createElement('div');
+                div.setAttribute('data-pp-account', account);
+                div.setAttribute('data-pp-amount', amount);
+                div.setAttribute('data-pp-style-layout', layout);
```
gillycheesesteak: Do we need to use inline attributes for these, or should there be a single method for setting up the basic config of the message?

tests/functional/spec/non-snapshot-tests/onEvents.test.js
```diff
@@ -0,0 +1,186 @@
...
+                        [name]: inValue
+                    };
+                }, {});
+
+                // Create the interior of the script function so that we create
+                // a paypal message with attached callback events
+                script.innerHTML = `
+                paypal.Message(
+                ${writeMsgConfig({ ...config, ...eventCallbacks })},
+                ).render('#message')`;
```
gillycheesesteak: can `exposeFunction` be used here to be able to call `paypal.Message()` directly in this script rather than creating another?

tests/functional/spec/non-snapshot-tests/onEvents.test.js
```diff
@@ -0,0 +1,186 @@
...
+    });
+
+    // Check that we can write a callback function on the message html tag
+    test.each([
+        // methods for attaching the callback function
+        'inline attribute',
+        'script'
+    ])('%s', async method => {
+        await page.exposeFunction(`getContext`, getContext);
+        await page.exposeFunction(`delay`, delay);
```
gillycheesesteak: can this be done once in `beforeAll`?

demo/blank.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <title>Messages Dev Sandbox</title>
+        <meta charset="UTF-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no" />
+        <!-- <script src="//localhost.paypal.com:8080/sdk/js?client-id=DEV00000000NI&components=messages"></script> -->
```
gillycheesesteak: should this comment be removed?

tests/functional/spec/non-snapshot-tests/onEvents.test.js
jadutter: I'm indifferent to whether it's a `switch` or an `if`-`else`. 
Want me to change it?

tests/functional/spec/non-snapshot-tests/onEvents.test.js
jadutter: I would rather not mix the two methods of creating a message.

tests/functional/spec/non-snapshot-tests/onEvents.test.js
jadutter: `exposeFunction` may work, but I want to make sure the page loaded by puppeteer is executing the function that a merchant would be loading and executing.

demo/blank.html
jadutter: Probably. 

Follow up question though; should we trust that if functional tests pass for `message.js` that it will also pass for the sdk?

tests/functional/spec/non-snapshot-tests/onEvents.test.js
jadutter: My goal was to create functional tests that would simulate exactly what a merchant would have at their disposal; I wanted to be able to say with certainty (and automation), that the methods of attaching a callback function as described in our documentation would work.

tests/functional/spec/non-snapshot-tests/onEvents.test.js
gillycheesesteak: yes, lets change

tests/unit/spec/src/controllers/modal/interface.test.js
```diff
@@ -31,6 +31,12 @@ jest.mock('src/utils/logger', () => ({
     }
 }));
 
+const onEvents = {
```
gillycheesesteak: Are there supposed to be new tests that use this?

jadutter: Nope. These are scraps I meant to remove before committing.

demo/blank.html
gillycheesesteak: Let's remove one of the methods of including messaging (either should be fine)

PR #743, opened 2022-02-21
aditya kapadiya patch 1

Aditya-Kapadiya:
Changed HTTP to HTTPS for secure connection. The link redirected me to the HTTP as HTTP is unsecure it can lead to a privacy issues or the user can become a victim of MITM attack and can be vulnerable to many other security risks.


JustinDoan:
Hi @Aditya-Kapadiya, The URL is taken directly from the Apache license template. It may be better to reach out to the Apache org about updating this.

PR #753, opened 2022-03-01
chore: update packages to krakenjs scope

westeezy:
Updating the krakenjs github packages to be under the npm org `@krakenjs`. There are no changes to the packages themselves just the organization move.


gregjopa:
Great to see CI is passing! @westeezy can you update the PR description and mark this as ready for review?

I added @gillycheesesteak and @Seavenly to help review.

Seavenly:
@westeezy Should we update `grumbler-scripts` in `devDependencies`to use `@krakenjs` as well?

westeezy:
@Seavenly haven't moved `grumbler-scripts` to krakenjs scope just yet, but that being said might as well bump it today just so we can take care of it in one shot. So long story short I'll do that sometime later today

westeezy:
@Seavenly given this repo is on grumbler-scripts v3 there will be some bigger updates to make to bump it to latest. Apologies I did not notice it was behind at first. This is something I can do but I propose we do it after moving to `@krakenjs` for runtime dependencies. This will provide us a more iterative path and allow us to get this unit of work out rather than letting it idle.

westeezy:
Had some extra time so added the move to grumbler-scripts v6 as well. required updating a number of eslint related packages and turning off a few new rules which yall can chose to turn back on if you like :D 

edit: looks like dev builds have issues with nullish coalescing but all other builds seem fine.. ill investigate this week
edit2: needed node 14 for the register which is using @babel/preset-env. pushing a commit to use node14 in CI, nvmrc, etc

package.json
```diff
@@ -42,45 +42,45 @@
...
+        "@krakenjs/beaver-logger": "^5.0.0",
+        "@krakenjs/belter": "^2.0.0",
+        "@krakenjs/jsx-pragmatic": "^3.0.0",
+        "@krakenjs/post-robot": "^11.0.0",
+        "@krakenjs/zalgo-promise": "^2.0.0",
+        "@krakenjs/zoid": "^10.0.0",
+        "@paypal/common-components": "^1.0.24-alpha.1",
+        "@paypal/sdk-client": "^4.0.166-alpha.2",
+        "@paypal/sdk-constants": "^1.0.11-6.alpha-1",
+        "@paypal/sdk-logos": "^1.0.43-alpha.1",
```
Seavenly: Are these `alpha` versions the latest?

package.json
```diff
@@ -42,45 +42,45 @@
...
-        "jsx-pragmatic": "^2.0.7",
-        "post-robot": "^10.0.29",
-        "zalgo-promise": "^1.0.38",
-        "zoid": "^9.0.46"
+        "@krakenjs/beaver-logger": "^5.0.0",
+        "@krakenjs/belter": "^2.0.0",
+        "@krakenjs/jsx-pragmatic": "^3.0.0",
+        "@krakenjs/post-robot": "^11.0.0",
+        "@krakenjs/zalgo-promise": "^2.0.0",
+        "@krakenjs/zoid": "^10.0.0",
```
Seavenly, line 50: Are these major version bumps just for switching the dependency location, and not any actually breaking changes we should be cautious about?

westeezy: correct just a change to be under the `krakenjs` npm scope

package.json
westeezy: these will be replaced before the merge with the published versions. alpha was purely for testing

westeezy: pushed an update to most recent versions

Seavenly (approved):

PR #761, opened 2022-03-07
refactor: lander index for instrumentation

grablack:


src/utils/logger.js
```diff
@@ -18,11 +18,13 @@ export const logger = Logger({
     flushInterval: 10 * 1000,
     // Override transport so we can use withCredentials
     transport: ({ url, method, json, headers }) => {
+        // Lander does not have index providing one
+        const landerPayload = __MESSAGES__.__TARGET__ === 'LANDER' ? '1' : undefined;
```
Seavenly: Is this necessary since `eventsIndexes` gets combined with `trackingIndexes` here? https://github.com/paypal/paypal-messaging-components/blob/2965531faf00d077af601bc48616817906f07ccf/src/utils/logger.js#L35

grablack: removed

Seavenly (approved):

PR #773, opened 2022-03-18
refactor: enhance offer param to support more offers

erikacolette29:


src/library/zoid/message/validation.js
```diff
@@ -107,10 +108,18 @@ export default {
...
     },
     offer: ({ props: { offer } }) => {
+        const offerTypes = [
+            OFFER.PAY_LATER_SHORT_TERM,
+            OFFER.PAY_LATER_LONG_TERM,
+            OFFER.PAY_LATER_PAY_IN_1,
+            OFFER.PAYPAL_CREDIT_NO_INTEREST,
+            OFFER.PAYPAL_CREDIT_INSTALLMENTS,
+            'NI'
+        ];
```
Seavenly: Consider iterating over `OFFER`'s values so that if a new value is ever added in the future it gets automatically included.
``` javascript
const offerType = [...Object.values(OFFER), 'NI'];
```

Seavenly (approved):

Seavenly (approved):

PR #776, opened 2022-03-22
fix: duplicate stats events

paypal-rosman:


src/components/message/Message.js
```diff
@@ -100,11 +100,20 @@ const Message = function({ markup, meta, parentStyles, warnings }) {
 
     if (typeof onProps === 'function') {
         onProps(xprops => {
-            const shouldRerender = Object.keys(props).some(key =>
-                typeof props[key] !== 'object'
-                    ? props[key] !== xprops[key]
-                    : JSON.stringify(props[key]) !== JSON.stringify(xprops[key])
-            );
+            const shouldRerender = Object.keys(props)
+                // lets check what is null and undefined and make sure we do not attempt to update
```
gillycheesesteak: Let's expand on this comment a bit about exactly the issue we saw

paypal-rosman: updated

src/components/message/Message.js
```diff
@@ -100,11 +100,23 @@ const Message = function({ markup, meta, parentStyles, warnings }) {
...
     if (typeof onProps === 'function') {
         onProps(xprops => {
-            const shouldRerender = Object.keys(props).some(key =>
-                typeof props[key] !== 'object'
-                    ? props[key] !== xprops[key]
-                    : JSON.stringify(props[key]) !== JSON.stringify(xprops[key])
-            );
+            const shouldRerender = Object.keys(props)
+                // specific situation where this variable would compare null to undefined casuing it to render as true.
+                // to migtigae that we are filtering all xprops and props values that are equal to null or undefined
```
gillycheesesteak: `migtigae` => `mitigate`

gillycheesesteak (approved):

src/components/message/Message.js
```diff
@@ -100,11 +100,23 @@ const Message = function({ markup, meta, parentStyles, warnings }) {
...
+                // to mitigate that we are filtering all xprops and props values that are equal to null or undefined
+                // thus only checking values that are not null or undefined and determing if we need to rerender or not. 
+                // Rene 03/23/22
+                .filter(
+                    key =>
+                        props[key] !== undefined &&
+                        props[key] !== null &&
+                        xprops[key] !== undefined &&
+                        xprops[key] !== null
+                )
```
Seavenly: Will this prevent a rerender if a value purposely changes from something to nothing? Such as explicitly changing an amount value from `50` to `null` if a merchant wants to revert back to a generic message. Maybe there is a banner rendered on an SPA that spans multiple pages for example that gets updated. Should we instead improve the predicate function of `some()` instead of filtering out comparing a prop all together?

Also consider using `typeof value === 'undefined'` instead of `value === undefined` when checking for `undefined` as it is considered best practice since it will catch a handful of edge-cases that can slip through with `value === undefined`.

src/components/message/Message.js
```diff
@@ -100,11 +100,23 @@ const Message = function({ markup, meta, parentStyles, warnings }) {
...
-            const shouldRerender = Object.keys(props).some(key =>
-                typeof props[key] !== 'object'
-                    ? props[key] !== xprops[key]
-                    : JSON.stringify(props[key]) !== JSON.stringify(xprops[key])
-            );
+            const shouldRerender = Object.keys(props)
+                // specific situation where this variable would compare null to undefined casuing it to render as true.
+                // to mitigate that we are filtering all xprops and props values that are equal to null or undefined
+                // thus only checking values that are not null or undefined and determing if we need to rerender or not. 
+                // Rene 03/23/22
```
Seavenly: Can this be removed since the authorship is baked into the git commit itself?

src/components/message/Message.js
paypal-rosman: good catch, i just tested the case of `amount: 60` then did not pass amount in a `setTimeout` to simulate a prop update and it looks like with our old code the message did re-render but did not update the amount ( so with the rerender message it showed `amount: 60` still when it should be undefined) and with this new code i put in. So need to take a step back and come up with something better it seems like

src/components/message/Message.js
```diff
@@ -98,13 +98,28 @@ const Message = function({ markup, meta, parentStyles, warnings }) {
...
-            const shouldRerender = Object.keys(props).some(key =>
-                typeof props[key] !== 'object'
-                    ? props[key] !== xprops[key]
-                    : JSON.stringify(props[key]) !== JSON.stringify(xprops[key])
-            );
+            const shouldRerender = Object.keys(props).some(key => {
+                // check to see if either props or xpros values have changed
+                // if any of the props or xprops value have changed
+                // let's take an extra step to determine which values do not equal each other
+                if (didPropsChange(props[key]) || didPropsChange(xprops[key])) {
```
paypal-rosman: Taking a closer look at this, do we only need to check the xprops values? because at the time of this check `props[key]`  has not changed since the original `createState` function. So there might be no need to even check if `props[key]` is undefined or null.

src/components/message/Message.js
```diff
@@ -98,13 +98,28 @@ const Message = function({ markup, meta, parentStyles, warnings }) {
         button.focus();
     });
 
+    // return boolean if propsValue is not equal to undefined or null
+    // propsValue can be either props or xprops
+    // if returns true it means propsValue changed
+    // if returns false it means propsValue did not change
+    const didPropsChange = propsValue => {
```
gillycheesesteak: I think the function name is confusing here - this is only detecting whether we want to consider this prop for a change detection. ex: It is possible that the prop started as `50` and ended as `50`, but this function will return `true`

src/components/message/Message.js
```diff
@@ -98,13 +98,28 @@ const Message = function({ markup, meta, parentStyles, warnings }) {
...
         onProps(xprops => {
-            const shouldRerender = Object.keys(props).some(key =>
-                typeof props[key] !== 'object'
-                    ? props[key] !== xprops[key]
-                    : JSON.stringify(props[key]) !== JSON.stringify(xprops[key])
-            );
+            const shouldRerender = Object.keys(props).some(key => {
+                // check to see if either props or xpros values have changed
+                // if any of the props or xprops value have changed
+                // let's take an extra step to determine which values do not equal each other
```
gillycheesesteak: This comment is similarly misleading, because the first thing we are doing is checking whether we want to _exclude_ a key because it started and ended as `null|undefined`

src/components/message/Message.js
gillycheesesteak: This check should be using `&&`, we want to make sure that both before and after values are not `null|undefined`

src/components/message/Message.js
paypal-rosman: let me change the name of the function, now that i am looking at it, it is confusing. The function does return true in that case but the 2nd check in the ternary conditional will make sure it does not rerender because 

`props[key] !== xprops[key]`

gillycheesesteak (approved):

PR #778, opened 2022-03-23
refactor: clean up FR and GB content and mutation files

yeul:
Cleans up FR and GB local content files after backwards compatibility fix for the non-qualifying purchase range release. Also cleans up mutations.


yeul:
replaced by #792

PR #783, opened 2022-03-31
chore: change EUR to €

merlinstardust:


src/components/modal/v2/parts/views/ShortTerm/Content.jsx
```diff
@@ -59,7 +60,7 @@ export const ShortTerm = ({
                                                 <Donut
                                                     key={index}
                                                     qualifying={qualifying}
-                                                    periodicPayment={periodicPayment}
+                                                    periodicPayment={formattedPeriodicPayment}
```
yeul:
```suggestion
                                                    periodicPayment={periodicPayment.replace(/(\s?EUR)/g, ' €')}
```
Can we change this to the above? it might get confusing if we call it `formattedPeriodicPayment` since that's an actual variable that exists as far as content coming back is concerned. Also, instead of wrapping it in `currencyFormat` which also lops off 0s, we probably just want to have the above for now just to keep the 0s since we don't like removing those when it's a payment breakdown. This regex above will only update EUR to the symbol and everything else stays the same.

src/components/modal/v2/parts/views/ShortTerm/Content.jsx
```diff
@@ -38,8 +38,9 @@ export const ShortTerm = ({
     };
 
     const donutScreenReaderString = donutTimestamps
-        .map(timestamp => `${periodicPayment.replace('.00', '')} for ${timestamp}`)
+        .map(timestamp => `${currencyFormat(periodicPayment)} for ${timestamp}`)
         .join(', ');
+    const formattedPeriodicPayment = currencyFormat(periodicPayment);
```
yeul:
```suggestion
```

src/components/modal/v2/lib/hooks/currency.js
```diff
@@ -10,6 +10,7 @@ export const currencyFormat = string => {
             // eslint-disable-next-line security/detect-unsafe-regex
             /((\$|£)?(\d{1,5}(\.|,)){1,3}00)/g
```
yeul:
```suggestion
            /((\$|£)?(\d{1,5}(\.|,)){1,3}\d\d)(\s?EUR)?/g
```
If we update this regex here it should cover our usecase.

src/components/modal/v2/lib/hooks/currency.js
```diff
@@ -10,6 +10,7 @@ export const currencyFormat = string => {
...
         ) ?? null;
+
     if (match !== null) {
         match.forEach(foundString => {
             const filteredString = foundString
```
yeul, line 16:
```suggestion
            const filteredString = foundString
                .replace(/(\.|,)00-/g, '-')
                .replace(/(\.|,)00$/g, '')
                .replace(/(\.|,)00(.|\s*)EUR/g, ' €');
            formattedStr = formattedStr.replace(foundString, filteredString);
```
We should be good to simplify this regex to the above since we aren't dealing with messages here.

src/components/modal/v2/lib/hooks/currency.js
```diff
@@ -21,5 +22,5 @@ export const currencyFormat = string => {
             formattedStr = formattedStr.replace(foundString, filteredString);
         });
     }
-    return formattedStr;
+    return formattedStr.replace(/(EUR)/g, '€');
```
yeul:
```suggestion
    return formattedStr;
```

merlinstardust:
closing in favor of #788

PR #784, opened 2022-04-01
fix: prevent universal modal view switching on window resize

yeul:


src/components/modal/v2/lib/providers/transition.js
```diff
@@ -32,6 +32,7 @@ export const TransitionStateProvider = ({ children }) => {
                         if (entry.isIntersecting) {
                             setState(STATUS.OPEN);
                             onShow();
+                            observer.unobserve(document.body);
```
Seavenly: Does unobserving prevent our resets from firing as the modal opens and closes multiple times?
https://github.com/paypal/paypal-messaging-components/blob/211217610459a30a628b2baa442e3bcbe46ad3d5/src/components/modal/v2/parts/Container.jsx#L41
https://github.com/paypal/paypal-messaging-components/blob/211217610459a30a628b2baa442e3bcbe46ad3d5/src/components/modal/v2/parts/BodyContent.jsx#L63

merlinstardust (approved):

PR #790, opened 2022-04-11
fix: adds missing clientId and partnerClientId missing on stats event

danzhaas:


src/library/zoid/message/component.js
```diff
@@ -237,6 +237,7 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
                                     type: 'message',
                                     messageRequestId,
                                     account: merchantId || account,
+                                    partnerClientId: merchantId && account.slice(10), // slice is to remove the characters 'client-id:' from account name
```
gillycheesesteak: We should add these to `buildStatsPayload` in `utils/stats.js` so they will be automatically included downstream

danzhaas: @gillycheesesteak do you have a preferred way in mind for how can we get `merchantId` and `account` (or `partnerClientId`) into `stats.js`?

danzhaas:
returning to previous flow per this comment after last commit from Ryan (on Slack):

> actually, it might be better to just add it as a top-level field to meta which is setup right above the call to runStats
> where account is added

src/library/zoid/message/component.js
```diff
@@ -209,6 +209,7 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
                     return ({ meta, activeTags, deviceID, requestDuration, messageRequestId }) => {
                         const { account, merchantId, index, modal, getContainer } = props;
                         const { trackingDetails, offerType, ppDebugId } = meta;
+                        const partnerClientId = merchantId && account.slice(10); // slice is to remove the characters 'client-id:' from account name
```
gillycheesesteak, line 212: Should we also use `account.slice(10)` for the `account` field below when `merchantId` is not present?

danzhaas: just addressed this in latest commit: standardizes client id string format

Seavenly: We can't slice off `client-id:` from the `account` value without making appropriate changes to the logging endpoint. We specifically look for `client-id:` to determine payer ID vs client ID currently.

danzhaas: Good callout @Seavenly .  Does removing the `client-id:` bring value making it worth it to change how the endpoint works currently?  AKA if it ain't broke, don't fix it?  If you guys decide yes, we could use string length to differentiate `payerId` vs `clientId`

danzhaas: also paging @paypal-are

Seavenly: @danzhaas Let's keep this simple for now and focus on just the `partnerClientId` portion and not worry about refactoring our endpoint.

src/library/zoid/message/component.js
```diff
@@ -236,7 +238,8 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
                                 [index]: {
                                     type: 'message',
                                     messageRequestId,
-                                    account: merchantId || account,
+                                    account: merchantId || clientId,
+                                    partnerClientId,
```
Seavenly, line 241: Should we also include this for the modal zoid component?

danzhaas: @Seavenly just documenting what you said in our related discussion from slack:

> I’m pretty confident the answer is yes
> When there’s a standalone modal integration, we would not have the partnerClientId that comes down with the message meta, so it would currently be missing based on the changes

src/library/zoid/message/component.js
danzhaas: ok, it will be undone in this next commit

danzhaas:
new stage tag w/ partner_client_id added to modal zoid component  : dan_partner_client_id_thr
https://UIDeploy--StaticContent--dan_partner_client_id_thr--ghe.preview.dev.paypalinc.com/upstream/bizcomponents/stage?cdn:list

danzhaas:
standalone modal logging `partner_client_id` validated and logged in the jira story.  ready for review

Seavenly (approved):

PR #795, opened 2022-04-15
refactor: use estimated installments

grablack:


src/components/modal/v2/parts/views/ShortTerm/Content.jsx
```diff
@@ -37,9 +37,12 @@ export const ShortTerm = ({
         );
     };
 
-    const donutScreenReaderString = donutTimestamps
-        .map(timestamp => `${currencyFormat(periodicPayment)} for ${timestamp}`)
-        .join(', ');
+    const donutScreenReaderString = Object.keys(estimatedInstallments?.items).length
```
merlinstardust: Since `Object.keys(estimatedInstallments?.items).length` gets used so frequently, it would be good to move it into its own variable like `hasInstallments`

merlinstardust (changes requested):

merlinstardust (approved):

PR #799, opened 2022-04-21
fix: edge caching TS cookie

paypal-rosman:
in order to fix this we are grabbing the `ts_c` cookie from the zoid iframe, writing it to local storage, then appending it to the `meta.global` property in the log endpoint.


src/library/zoid/message/component.js
```diff
@@ -210,10 +211,25 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
                         const { account, merchantId, index, modal, getContainer } = props;
                         const { trackingDetails, offerType, ppDebugId } = meta;
 
+                        const tsCookie =
+                            getCookieByName('ts_c') !== null
```
gillycheesesteak: Should we put `getCookieByName('ts_c')` into a variable to reuse below?

src/library/zoid/message/component.js
```diff
@@ -210,10 +211,25 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
...
+                                  decodeURIComponent(getCookieByName('ts_c').slice(5, getCookieByName('ts_c').length))
+                                      .split('&')
+                                      .reduce((acc, currValue) => {
+                                          // for example vr=2e1ab8701800ae7732f49876ffffffd7
+                                          // we split at = which returns
+                                          // ['vr', '2e1ab8701800ae7732f49876ffffffd7']
+                                          const [key, value] = currValue.split('=');
+                                          return { ...acc, [key]: value };
+                                      }, {})
+                                : getCookieByName('ts_c');
```
gillycheesesteak: In this case, `getCookieByName('ts_c')` will always be `null`, right?

paypal-rosman: yes i should probably just assign it to null or just do a [null check ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) actually i think we may not want to do that

src/library/zoid/message/component.js
paypal-rosman: yea good idea let me do that. it will be cleaner

src/utils/miscellaneous.js
```diff
@@ -268,3 +269,43 @@ export function getStandardProductOffer(offer) {
...
+export function getCookieByName(name) {
+    const cookieString = document.cookie.match(`${name}=[^;]+`);
+    // cookieString will return null if no match
+    return cookieString ? formatTsCookie(cookieString[0]) : cookieString;
+}
+
+// get the ts cookie from local storage
+export function getTsCookieFromStorage() {
+    return getStorage().getState(storage => storage.ts) !== null ||
+        typeof getStorage().getState(storage => storage.ts) !== 'undefined'
```
JustinDoan: Should be able to use something like

getStorage().getState(storage => storage?.ts) ? getCookieByName : getStorage().getState(storage => storage.ts)

JustinDoan: Actually could shorten this further with

getStorage().getState(storage => storage?.ts) ?? getCookieByName('ts_c')

JustinDoan: Depends if there is a possibility of "" being in .ts

src/utils/miscellaneous.js
```diff
@@ -268,3 +269,43 @@ export function getStandardProductOffer(offer) {
...
+
+/**
+ * Get value of cookie name
+ * @param name - name of cookie
+ * @returns string of url encoded string of vt and vr
+ */
+export function getCookieByName(name) {
+    const cookieString = document.cookie.match(`${name}=[^;]+`);
+    // cookieString will return null if no match
+    return cookieString ? formatTsCookie(cookieString[0]) : cookieString;
```
JustinDoan: Nullish coalescing operator could be used here

src/utils/sdk.js
```diff
@@ -156,14 +156,14 @@ function getRawStorage() {
...
         try {
             window.localStorage?.setItem(
                 `__${getNamespace()}_storage__`,
                 JSON.stringify({
                     ...getRawStorage(),
-                    id: storageID
+                    ...values
                 })
             );
         } catch (e) {
```
JustinDoan: Since we are relying on isLocalStorageEnabled, do we want to log anything/err anything, maybe with ppDebug or something if this fails?

src/utils/miscellaneous.js
paypal-rosman: no it will always be null or have a value so this 
`getStorage().getState(storage => storage?.ts) ?? getCookieByName('ts_c')`

should work, I will test it

src/utils/miscellaneous.js
paypal-rosman: this returns a true or false value so i am pretty Nullish coalescing wouldn't work here correct?

JustinDoan: What do you mean this returns true or false? looks like it will return null or a formatted string?

paypal-rosman: whoops you are right, not sure what i was thinking about, clearly the comment is right in front of my face.... Let me modify

src/utils/miscellaneous.js
```diff
@@ -268,3 +269,40 @@ export function getStandardProductOffer(offer) {
...
+        : null;
+}
+
+/**
+ * Get value of cookie name
+ * @param name - name of cookie
+ * @returns string of url encoded string of vt and vr
+ */
+export function getCookieByName(name) {
+    const cookieString = document.cookie.match(`${name}=[^;]+`);
```
jadutter: Can we revise this to not use regex?

src/utils/miscellaneous.js
```diff
@@ -280,19 +280,19 @@ export function getCookieByName(name) {
...
         // get all cookies
         document.cookie
-            // separate the string into an array of cookies
             .split(';')
+            // separate the string into an array of cookies
             // find the cookie by name
             .find(cookieStr => cookieStr.startsWith(`${name}=`))
-            // use only the value of the cookie
-            .slice(5)
+            ?.slice(5)
```
jadutter:
```suggestion
            ?.slice(5) ?? ''
```
otherwise, 
```javascript 
cookieVal === decodeURIComponent(undefined) === 'undefined'
```
and the function will return `{undefined: ''}` instead of of `null`

jadutter (dismissed):

jadutter (approved):

PR #804, opened 2022-04-25
refactor: update US file and directory names

Seavenly:


merlinstardust:
Since this is all renaming stuff, can we rename `src/components/modal/v2/parts/views/NI` to `src/components/modal/v2/parts/views/NoInterest`?

yeul (approved):

PR #821, opened 2022-06-07
feat: add DE universal modal views

yeul:


src/components/modal/v2/parts/BodyContent.jsx
```diff
@@ -11,7 +11,7 @@ import {
     useTransitionState
 } from '../lib';
 import Header from './Header';
-import { LongTerm, ShortTerm, NI, ProductList } from './views';
+import { LongTerm, ShortTerm, NI, ProductList, PayIn1 } from './views';
```
merlinstardust: Thoughts on changing `PayIn1` to `PayInOne` and `NI` to `NoInterest` for consistency?

src/components/modal/v2/parts/BodyContent.jsx
```diff
@@ -72,7 +72,8 @@ const BodyContent = () => {
             <ShortTerm content={content} productMeta={productMeta} openProductList={openProductList} />
         ),
         PAYPAL_CREDIT_NO_INTEREST: <NI content={content} openProductList={openProductList} />,
-        PRODUCT_LIST: <ProductList content={content} setViewName={setViewName} />
+        PRODUCT_LIST: <ProductList content={content} setViewName={setViewName} />,
+        PAY_LATER_PAY_IN_1: <PayIn1 content={content} openProductList={openProductList} />
```
merlinstardust: Should we keep these ordered alphabetically?

src/components/modal/v2/parts/Container.jsx
```diff
@@ -85,12 +85,12 @@ const Container = ({ children }) => {
 
     return (
         <ScrollProvider containerRef={contentWrapperRef}>
-            <div className={`modal-wrapper${loading ? ' loading' : ''}`}>
+            <div className="modal-wrapper">
```
merlinstardust, line 88: Why are we removing `loading`?

src/components/modal/v2/parts/views/NI/styles.scss
```diff
@@ -2,6 +2,10 @@
 @import '../../../styles/globals/mixins';
 @import '../../../styles/globals/colors';
 
+.spinner {
+    display: none;
```
merlinstardust: Since `.spinner` is `display: none` in so many places, maybe the common css should have `display: none` and we just use `display: block` when it needs to be shown. Thoughts?

src/components/modal/v2/parts/Container.jsx
yeul: This is related to the blur style that occurs. Similar to how it worked in the older modal versions, if our state is loading, the loading class is added and from there so is a blur and opacity style to make the shift to the new terms visually less confusing/sudden. Once we are no longer loading the class is removed and so is the style.

src/components/modal/v2/parts/BodyContent.jsx
yeul: Yup no problem, I can totally do that.

src/components/modal/v2/parts/BodyContent.jsx
yeul: Once I merge in develop, `NI` should flip over to `NoInterest`. As for `PayIn1`, I figured it could be called `PayIn1` rather than `PayInOne` to match the credit product identifier naming scheme: `PAY_LATER_PAY_IN_1`.

src/components/modal/v2/parts/views/NI/styles.scss
yeul: Yeah let me take a look into this.

merlinstardust (approved):

PR #842, opened 2022-08-09
Update modals for legal disclosure not showing in the mobile viewport (DTCRCMERC-1565)

erikacolette29:


src/components/modal/v2/styles/globals/_mixins.scss
```diff
@@ -23,6 +23,12 @@
     }
 }
 
+@mixin verySmallMobile {
+    @media (max-width: 321px) {
+        @content;
+    }
+}
+
```
merlinstardust, line 31: I think this will work a little better in terms of multiple widths but the style changes look good.
```
@mixin smallHeightMobile {
    @media (max-height: 700px) {
        @content;
    }
}
```

merlinstardust (changes requested):

merlinstardust (approved):

PR #844, opened 2022-08-12
ci: fix unzip security

Seavenly:


.github/workflows/snapshotCommit.yml
```diff
@@ -41,16 +41,20 @@ jobs:
...
 
                           console.log(`Checking artifact: ${artifact.name}`)
 
+                          execSync(`unzip -o ${tmpDir}.zip -d ${tmpDir}`);
+
                           if (artifact.name === 'metadata') {
-                            execSync(`unzip -o ${{ github.workspace }}/${artifact.name}.zip`);
-                          } else if (artifact.name.includes('v2')) {
-                            execSync(`unzip -o ${{ github.workspace }}/${artifact.name}.zip -d tests/functional/v2/snapshots`);
+                            execSync(`rsync -zavpmr --no-links --include='pull_number.txt' --exclude='*' ${tmpDir}/ ${{ github.workspace }}`);
```
jadutter, line 53: is `--no-links` a valid option for `rsync`?

1. [actions/github-script](https://github.com/actions/github-script/blob/main/action.yml#L30) seems to use node16
1. [node16](https://github.com/nodejs/docker-node/blob/f82af606acd44dc6be7fbb2a069922afa32657f3/16/buster/Dockerfile#L1) seems to build upon buster (a debian version).
1. [rsync package](https://packages.debian.org/bullseye/rsync) for debian links to a homepage, which contains the documentation
1. I don't see the option in the [rsync documentation](https://download.samba.org/pub/rsync/rsync.1)

I also tried it locally
```bash 
docker container run --rm -it node:16 bash
apt update && (yes | apt-get install rsync) && rsync --help 
```

Instead of `-zavpmr --no-links`, I think we want `-zvmrptgoD`, by replacing `-a` with its equivalents, but removing the `-l` option.

.github/workflows/snapshotCommit.yml
```diff
@@ -41,16 +41,20 @@ jobs:
                             archive_format: 'zip'
                           });
                           
-                          fs.writeFileSync(`${{ github.workspace }}/${artifact.name}.zip`, Buffer.from(download.data));
+                          const tmpDir = `/tmp/${artifact.name.replace(/[^a-zA-Z0-9-_]/g, '')}`;
```
merlinstardust, line 44: All looks good. Only change I would make is to at least have this be `tempDir` with an `e` just to be a little more clear. And `/tmp` could be `/temp` for the same reason

merlinstardust (changes requested):

.github/workflows/snapshotCommit.yml
Seavenly: For Linux base systems, `/tmp` is the official root directory where temporary files are written. For that reason the variable is named `tmpDir`. We should continue to use `/tmp`, but I'm happy to rename the variable to `tempDir` if we'd like.

merlinstardust: I suppose it's fine then. I guess I'll just take it up with the creators of Linux. Lol

merlinstardust (approved):

.github/workflows/snapshotCommit.yml
merlinstardust: Can a comment be added as to what this does? What is `-zavpmr` or `-zvmrptgoD`?

jadutter (approved):

PR #846, opened 2022-08-24
feat: Updates to DE no amount message (DTCRCMERC-1600)

erikacolette29:


src/server/locale/DE/GPL/mutations/index.js
```diff
@@ -24,7 +25,10 @@ export default function getMutations(id, type) {
         case 'GPLQ:GTZ:NON-DE':
             return gplqGtzNonDe[type];
         case 'GPL:EQZ':
-        default:
             return gplEqz[type];
+        case 'GENERIC':
+        case 'GENERIC:NON-DE':
+        default:
+            return generic[type];
```
yeul:
```suggestion
        case 'GENERIC:NON-DE':
            return genericNonDe[type]
        case 'GENERIC':
        default:
            return generic[type];
```

yeul (approved):

PR #850, opened 2022-09-06
fix: universal modal lander header stickiness on mobile

yeul:


src/components/modal/v2/parts/Container.jsx
```diff
@@ -90,7 +90,11 @@ const Container = ({ children }) => {
                 <div className="spinner" style={{ opacity: loading ? '1' : '0' }} />
                 <Overlay />
                 {/* Presentational div to clip scrollbars with a rounded border */}
-                <div className={`content__wrapper-overflow ${loading ? 'loading' : ''}`}>
+                {/* Lander variant uses the div with className content__wrapper-overflow as the contentWrapperRef */}
+                <div
+                    className={`content__wrapper-overflow ${loading ? 'loading' : ''}`}
+                    ref={!!(isLander && !isIframe) && contentWrapperRef}
```
merlinstardust, line 96: Can a ref point to two different things? Why not make a separate ref for this?

src/components/modal/v2/parts/Header.jsx
```diff
@@ -24,20 +24,18 @@ const Header = ({ headline, subheadline, logo, isQualifying = 'false', qualifyin
...
-                        </div>
+                </div>
+                <div aria-hidden="true" className="header__fixed-wrapper">
+                    <div className="header__background-wrapper">
+                        <Icon name="header-background" />
                     </div>
-                </Fragment>
-            )}
+                </div>
+            </Fragment>
```
merlinstardust: This inner Fragment shouldn't be necessary

src/components/modal/v2/styles/components/_header.scss
```diff
@@ -128,7 +139,15 @@
...
+                    @include desktop {
+                        transform: translate(calc(#{$content-wrapper-max-width-api-iframe} / 2 - 50%), 2rem);
+                    }
+
+                    @include tablet {
+                        transform: translate(calc(#{$content-wrapper-max-width-api-iframe} / 2 - 50%), 2rem);
+                    }
 
                     // copied here because apiIframe mixin has higher priority than the scroll below
                     &--scroll {
```
merlinstardust, line 153: For the check annotations, aria attributes should be `dash-case`.
So we should be using `aria-selected`, `aria-controls`, `aria-labelledby`

https://reactjs.org/docs/accessibility.html

merlinstardust (changes requested):

merlinstardust (approved):

PR #861, opened 2022-09-30
feat: PPC 3.0 rebrand 

erikacolette29:


src/components/modal/v2/styles/components/_instructions.scss
```diff
@@ -18,6 +18,8 @@
 
         @include mobile {
             padding: 0px;
+            margin-top: -11px;
+            margin-bottom: 0px;
         }
```
yeul: Instead of having this here, can we move these two styles out of the `@include mobile` media query and put it right below in its own class?

```suggestion
      &.ppc {
            @include mobile {
                margin-top: -11px;
                margin-bottom: 0px;
            }
        }
```

yeul (approved):

PR #885, opened 2022-11-17
chore: upgrade to grumbler scripts 8

dtjones404:
[DTPPSDK-934](https://engineering.paypalcorp.com/jira/browse/DTPPSDK-934)

* Migrated to grumbler scripts v8.  
* Updated Webpack CLI commands to be compatible with v3.


babel.config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-    extends: '@krakenjs/grumbler-scripts/config/.babelrc-browser',
+    extends: '@krakenjs/babel-config-grumbler/babelrc-browser',
```
Seavenly, line 2: Does this need to be included with the `package.json` changes?

package.json
```diff
@@ -58,7 +58,8 @@
         "@babel/eslint-parser": "^7.17.0",
         "@babel/register": "^7.10.4",
         "@babel/runtime": "7.12.18",
-        "@krakenjs/grumbler-scripts": "^7.0.0",
+        "@krakenjs/grumbler-scripts": "^8.1.1",
```
Seavenly: Is this required if we're recommended to use the individual config packages like `@krakenjs/webpack-config-grumbler`?

Seavenly (commented):
We have one other spot where we reference `@kraken/grumbler-scripts`: https://github.com/paypal/paypal-messaging-components/blob/d404550a75a4ba555b972f16fab4e488a29031f2/utils/devServerProxy/index.js#L2

package.json
dtjones404: Good point!  I've separated out the dependencies

src/.eslintrc
```diff
@@ -1,5 +1,5 @@
 {
-    "extends": ["../node_modules/@krakenjs/grumbler-scripts/config/.eslintrc-browser.js", "../.eslintrc"],
+    "extends": ["../node_modules/@krakenjs/eslint-config-grumbler/eslintrc-browser.js", "../.eslintrc"],
```
westeezy, line 2: You don't need to path into node_modules as that happens automatically but just an FYI. We don't need to change that now

package.json
```diff
@@ -58,7 +58,9 @@
         "@babel/eslint-parser": "^7.17.0",
         "@babel/register": "^7.10.4",
         "@babel/runtime": "7.12.18",
-        "@krakenjs/grumbler-scripts": "^7.0.0",
+        "@krakenjs/babel-config-grumbler": "^8.1.1",
+        "@krakenjs/eslint-config-grumbler": "^8.1.1",
+        "@krakenjs/webpack-config-grumbler": "^8.1.1",
```
westeezy, line 63: Look at that! Ya'll no longer need to download karma when you don't need it ;)

westeezy (approved):
lgtm from our side!

Seavenly (approved):

PR #891, opened 2022-12-14
fix: 3.0 modal font bolding

erikacolette29:


src/components/modal/v2/styles/components/_product-list-link.scss
```diff
@@ -5,7 +5,7 @@
     min-width: 315px;
     text-decoration: underline;
     color: $light-blue;
-    font-family: $paypal-open-bold;
+    font-family: $paypal-open;
```
yeul, line 8:
```suggestion
    font-family: $paypal-open;
    font-weight: 600;
```
Your font family change makes sense, let's just add back in the font-weight so that the link to the product list modal is still bold.

yeul (approved):

PR #898, opened 2023-01-27
fix: prevent DE LT view from displaying product list tiles

Seavenly:
Summary of the issue: The initial `useState` value of `''` is causing the modal to error out. When this occurs:
https://github.com/paypal/paypal-messaging-components/blob/b2e38445a97912bcb76488cfb6c70e5420e24fde/src/components/modal/v2/parts/views/LongTerm/Content.jsx#L156
resolves to `''` which is falsey and resolves to the incorrect branch: 
https://github.com/paypal/paypal-messaging-components/blob/b2e38445a97912bcb76488cfb6c70e5420e24fde/src/components/modal/v2/parts/views/LongTerm/Content.jsx#L159
which results in an object being passed in instead of a supported string or array of strings which causes the error to be thrown and halt the rendering. The `setState` of the "empty" `useState` is immediately called inside the `Calculator` forces an unnecessary rerender, so the dependency inside `Calculator` has been pulled up into the LT view and then passes what's needed into `Calculator` instead so that the LT view has all the info it needs on the first render.

### DIFF
--- a/src/components/modal/v2/parts/views/LongTerm/Content.jsx
+++ b/src/components/modal/v2/parts/views/LongTerm/Content.jsx
@@ -1,7 +1,7 @@
 /** @jsx h */
 import { h, Fragment } from 'preact';
 import { useState } from 'preact/hooks';
-import { useXProps, useServerData, useCalculator, getComputedVariables } from '../../../lib';
+import { useXProps, useServerData, getComputedVariables } from '../../../lib';
 import Calculator from '../../Calculator';
 import ProductListLink from '../../ProductListLink';
 import Instructions from '../../Instructions';
@@ -12,7 +12,6 @@ import styles from './styles.scss';
 /**
  * Checks qualifying offer APRs in order to determine which APR disclaimer to render.
  */
-
 const getAPRDetails = ({ offers, disclaimer: { zeroAPR, mixedAPR, nonZeroAPR } }) => {
     const qualifyingOffers = offers.filter(offer => offer?.meta?.qualifying === 'true');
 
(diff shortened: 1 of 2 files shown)


src/components/modal/v2/parts/views/LongTerm/Content.jsx
```diff
@@ -9,18 +9,72 @@ import InlineLinks from '../../InlineLinks';
...
+
+    return {
+        /**
+         * Specifically, this impacts US Long Term and which legal disclaimer shows underneath the offer cards.
+         * If no initial amount is passed in or there is an error, we default to the zeroAPR disclaimer as it is more
+         * generic. i.e. Terms may vary based on purchase amount.
+         */
+        aprDisclaimer: zeroAPR,
+        /**
+         * setAPRType is used by DE Long Term to determine which legal disclosure shows at the bottom of the modal.
```
yeul: The rest of the comment is fine, but since we're not using `setAPRType` anymore I think we can remove that from the beginning of the sentence. Just for clarity so no one gets confused.

src/components/modal/v2/parts/views/LongTerm/Content.jsx
```diff
@@ -9,18 +9,72 @@ import InlineLinks from '../../InlineLinks';
...
+
+        if (qualifyingOffers.length === totalZero) {
+            return {
+                aprDisclaimer: zeroAPR,
+                aprType: 'zeroAPR'
+            };
+        }
+
+        return {
+            aprDisclaimer: mixedAPR,
```
danzhaaspaypal, line 45: is a mixed apr possible in DE?  there is no `mixedAPR` in [DE LT's disclaimer](https://github.com/paypal/paypal-messaging-components/blob/develop/content/modals/DE/long_term.json#L30)

yeul: Currently that's not possible in DE, so the condition will resolve before this point.

Seavenly: Also just want to ensure that this little helper function has a consistent return structure where all keys are present to make it more predictable since this view is used for both US and DE.

yeul (approved):

PR #939, opened 2023-05-18
refactor: chunk multi-message rendering to prevent singular long task

grablack:
Break up the long running tasks


src/library/controllers/message/interface.js
```diff
@@ -58,146 +58,177 @@ export default (options = {}) => ({
...
+                            }
+                        };
 
-                        throw err;
-                    }
+                        if (mapIndex) {
+                            checkIdleCallback(resolvePromise);
+                        } else {
+                            resolvePromise();
+                        }
```
Seavenly: Since the `checkIdelCallback` function is only referenced once consider keeping the code inline with where it will execute to keep the mental stack smaller. We can also reduce a level of `else` nesting by short-circuiting when `mapIndex === 0`. `window` will always exist as well so we can remove the optional chaining checks.

```javascript
if (mapIndex === 0) {
    return resolvePromise();
}

if (window.requestIdleCallback) {
    // check that requestIdleCallback is available
    window.requestIdleCallback(resolvePromise);
} else {
    // basic fallback for Safari
    setTimeout(resolvePromise, mapIndex * 20);
}
```

src/library/controllers/message/interface.js
```diff
@@ -58,146 +58,177 @@ export default (options = {}) => ({
...
-                        if (!container.hasAttribute('data-pp-id')) {
-                            container.setAttribute('data-pp-id', nextIndex());
-                        }
+            /* eslint-disable-next-line  promise/no-native, compat/compat */
+            Promise.all(
+                validContainers.map((container, mapIndex) => {
+                    /* eslint-disable-next-line  promise/no-native, compat/compat */
+                    return new Promise((resolve, reject) => {
+                        // return resolved render and updateProps
+                        const resolvePromise = () => {
```
Seavenly: Can we rename this to something more descriptive of the work that will happen inside of this function? Maybe something like `renderOrUpdateMessage`?

src/library/controllers/message/interface.js
```diff
@@ -58,146 +58,177 @@ export default (options = {}) => ({
...
-                        );
-
-                        if (!container.hasAttribute('data-pp-id')) {
-                            container.setAttribute('data-pp-id', nextIndex());
-                        }
+            /* eslint-disable-next-line  promise/no-native, compat/compat */
+            Promise.all(
+                validContainers.map((container, mapIndex) => {
+                    /* eslint-disable-next-line  promise/no-native, compat/compat */
+                    return new Promise((resolve, reject) => {
```
Seavenly: We may want to consider swapping these back to `ZalgoPromise` instances to keep consistency through the code.

src/library/controllers/message/interface.js
```diff
@@ -58,146 +58,177 @@ export default (options = {}) => ({
...
+                            renderStart: ${new Date(renderStart).toLocaleString()},
+                            renderMessageTime: ${new Date().toLocaleString()}
+                            }`
+                                    );
+
+                                    return render(container)
+                                        .then(() => globalEvent.trigger('render'))
+                                        .then(() => {
+                                            resolve();
+                                        });
```
Seavenly:
```suggestion
                                        .then(resolve);
```

src/library/controllers/message/interface.js
```diff
@@ -58,146 +58,177 @@ export default (options = {}) => ({
...
+                                    (acc, [key, val]) =>
+                                        typeof val === 'undefined' ? acc : Object.assign(acc, { [key]: val }),
+                                    {}
+                                );
+
+                                return updateProps(updatedMessageProps)
+                                    .then(() => globalEvent.trigger('render'))
+                                    .then(() => {
+                                        resolve();
+                                    });
```
Seavenly:
```suggestion
                                    .then(resolve);
```

src/library/controllers/message/interface.js
```diff
@@ -58,146 +58,177 @@ export default (options = {}) => ({
...
+                            } catch (err) {
+                                // We only want console.warn to be called once
+                                // check for the known error offer_validation_error and if it’s not that error then we should throw the error again so it gets surfaced
+                                //  catch that known/safe error, but allow any other errors to escape through
+                                if (err.message === 'offer_validation_error') {
+                                    return undefined;
```
Seavenly:
```suggestion
                                    return resolve();
```

src/library/controllers/message/interface.js
Seavenly: Also put a `setTimeout` step for the Safari fallback in the suggestion above which seems to smooth things out a bit for that use case.

src/library/controllers/message/interface.js
grablack: I couldn't get ZalgoPromise to work nicely because it needs to return from inside of the timeouts.

src/library/controllers/message/interface.js
```diff
@@ -58,146 +58,168 @@ export default (options = {}) => ({
...
-
-                        return updateProps(updatedMessageProps).then(() => globalEvent.trigger('render'));
-                    } catch (err) {
-                        // We only want console.warn to be called once
-                        // check for the known error offer_validation_error and if it’s not that error then we should throw the error again so it gets surfaced
-                        //  catch that known/safe error, but allow any other errors to escape through
-                        if (err.message === 'offer_validation_error') {
-                            return undefined;
+                        if (mapIndex === 0) {
+                            return renderOrUpdateMessage();
```
Seavenly: We can remove the `consistent-return` eslint warning by updating this slightly
```suggestion
                            renderOrUpdateMessage();
                            return;
```

src/library/controllers/message/interface.js
```diff
@@ -58,146 +58,168 @@ export default (options = {}) => ({
...
-                        );
-
-                        if (!container.hasAttribute('data-pp-id')) {
-                            container.setAttribute('data-pp-id', nextIndex());
-                        }
+            /* eslint-disable-next-line  promise/no-native, compat/compat */
+            Promise.all(
+                validContainers.map((container, mapIndex) => {
+                    /* eslint-disable-next-line  promise/no-native, compat/compat, consistent-return */
+                    return new Promise((resolve, reject) => {
```
Seavenly: Based on the [previous comment](https://github.com/paypal/paypal-messaging-components/pull/939#discussion_r1218588949) we can't make this change?
```suggestion
            ZalgoPromise.all(
                validContainers.map((container, mapIndex) => {
                    return new ZalgoPromise((resolve, reject) => {
```

grablack: When I removed the promises for zalgopromise I forgot to add back the new ZalgoPromise((resolve, reject) portion. Thats why I couldn't get it to return

Seavenly (approved):

PR #941, opened 2023-05-30
Screen reader for the list of ineligible states reads some state abbreviations as words instead of letters (DTCRCMERC-1884)

erikacolette29:


erikacolette29:
moved to https://github.com/paypal/paypal-messaging-components/pull/942

PR #948, opened 2023-06-08
refactor: update ttl

merlinpaypal:


src/library/zoid/treatments/component.js
```diff
@@ -60,8 +60,6 @@ export default createGlobalVariableGetter('__paypal_credit_treatments__', () =>
                         });
 
                         globalEvent.trigger('treatments');
-
-                        close();
```
Seavenly, line 64: Looks like a test needs to be updated from this change

src/utils/experiments.js
```diff
@@ -24,7 +24,8 @@ export function getLocalTreatments() {
     if (experiments.expiration < Date.now()) {
         // use existing value, but update treatments in the background
         if (document.readyState === 'loading') {
-            window.addEventListener('DOMContentLoaded', fetchTreatments);
+            // wrapping fetchTreatments in a function prevents memoization
```
Seavenly: This doesn't prevent memoization. `fetchTreatments` is already wrapped by the `memoize` helper. `() => fetchTreatments()` ensures that unique event objects are not passed by the event listener into `fetchTreatments` which avoids the benefit of memoization, because we only want the logic inside `fetchTreatments` to execute once, which only occurs when `fetchTreatments` is always called with `()`

Seavenly (approved):

PR #957, opened 2023-07-05
feat: JS SDK Cookie Remediation - Phase 3 (DTCRCMERC-2336)

erikacolette29:


src/library/zoid/message/component.js
```diff
@@ -50,6 +51,16 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
...
+            disableSetCookie: {
+                type: 'boolean',
+                queryParam: 'disable-set-cookie',
+                value: getDisableSetCookie
+            },
+            features: {
+                type: 'string',
+                queryParam: 'features',
+                value: getFeatures
+            },
```
Seavenly: Can we move these below line 351 so that these are grouped with the computed props? These are props that are not controlled by user input such as the account or style options.

src/library/zoid/modal/component.js
```diff
@@ -49,6 +50,16 @@ export default createGlobalVariableGetter('__paypal_credit_modal__', () =>
...
+            disableSetCookie: {
+                type: 'boolean',
+                queryParam: 'disable-set-cookie',
+                value: getDisableSetCookie
+            },
+            features: {
+                type: 'string',
+                queryParam: 'features',
+                value: getFeatures
+            },
```
Seavenly: Can we move these down with the other computed props similar to the above comment?

src/utils/logger.js
```diff
@@ -42,7 +42,10 @@ export const logger = Logger({
                 {}
             );
 
-        return request(method, url, {
+        const urlNew =
+            getDisableSetCookie() === true ? `${url}?disable-set-cookie=true&features=disable-set-cookie` : url;
```
Seavenly: We can omit the `=== true` as that is already implied. Also consider giving this a more meaningful name to improve clarity such as `urlWithCookieParams`
```suggestion
        const urlWithCookieParams =
            getDisableSetCookie() ? `${url}?disable-set-cookie=true&features=disable-set-cookie` : url;
```

src/library/zoid/message/component.js
```diff
@@ -50,6 +51,18 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
...
+                queryParam: 'disable-set-cookie',
+                required: false,
+                value: getDisableSetCookie
+            },
+            features: {
+                type: 'string',
+                queryParam: 'features',
+                required: false,
+                value: getFeatures
+            },
```
Seavenly: Can we move these below line 353 so that they are grouped with the computed props? Those are props not dependent on user input such as the account or style that come directly from a user.

src/library/zoid/modal/component.js
```diff
@@ -49,6 +50,18 @@ export default createGlobalVariableGetter('__paypal_credit_modal__', () =>
...
+                queryParam: 'disable-set-cookie',
+                required: false,
+                value: getDisableSetCookie
+            },
+            features: {
+                type: 'string',
+                queryParam: 'features',
+                required: false,
+                value: getFeatures
+            },
```
Seavenly: Can we move these with the computed props similar to the comment above?

tests/unit/spec/src/utils/experiments.test.js
```diff
@@ -105,4 +105,15 @@ describe('experiments utils', () => {
...
+    test('Handles disableSetCookie to return truthy', () => {
+        expect(getDisableSetCookie).toBeTruthy();
+
+        ensureTreatments();
+
+        expect(document.querySelector('iframe')).not.toBeNull();
+
+        // treatments should not be marked ready
+        expect(globalEvent.trigger).not.toHaveBeenCalled();
+    });
```
Seavenly: Is this test verifying much more than the test directly above it? Since we're mocking the value of `getDisableSetCookie` above the extra line essentially asserts that our mock value is what we set it to. We could consider removing this if that is the case.

Seavenly (approved):

PR #960, opened 2023-07-21
feat: disableSetCookie Query Param Fix (DTCRCGEMI-1239)

erikacolette29:


src/library/zoid/message/component.js
```diff
@@ -468,7 +468,7 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
             },
             disableSetCookie: {
                 type: 'boolean',
-                queryParam: 'disable-set-cookie',
+                queryParam: 'disableSetCookie',
```
Seavenly: Since the prop name matches the `queryParam` value, we can simplify this by just using `queryParam: true` for all 3 instances.

src/utils/sdk.js
```diff
@@ -234,7 +234,7 @@ export function getDevTouchpoint() {
 }
 export function getFeatures() {
     if (getDisableSetCookie()) {
-        return 'disable-set-cookie';
+        return 'disableSetCookie';
```
Seavenly: We can leave the `features` value the same (`disable-set-cookie`) otherwise it will fail our validation

Seavenly (approved):

PR #968, opened 2023-08-11
refactor: remove stage references

yeul:


package.json
```diff
@@ -51,7 +51,8 @@
         "@paypal/sdk-client": "^4.0.166",
         "@paypal/sdk-constants": "^1.0.118",
         "@paypal/sdk-logos": "^2.0.0",
-        "core-js-pure": "3.31.1"
+        "core-js-pure": "3.31.1",
+        "dotenv": "^16.3.1"
```
Seavenly: Are we able to accomplish this without adding a `dependency`? This will cause it to be included as part of the SDK bundle.

globals.js
```diff
@@ -31,7 +32,7 @@ module.exports = (env = { TARGET: 'sdk' }) => ({
         __DEV_TOUCHPOINT__: env.DEV_TOUCHPOINT,
         __DOMAIN__: {
             __LOCAL__: `https://localhost.paypal.com:${PORT}`,
-            __STAGE__: 'https://www.msmaster.qa.paypal.com',
+            __STAGE__: `https://www.${process.env.STAGE_URL}.com`,
```
Seavenly: Have we tried this with the stage SDK bundler? I feel like this could possibly result in an `undefined` value.

src/components/modal/v2/lib/hooks/applyNow.js
```diff
@@ -7,17 +7,17 @@ export default (clickTitle, src = 'link_click') => {
...
+        sandbox: 'https://www.sandbox.paypal.com',
+        production: 'https://www.paypal.com'
     }[env];
 
     return () => {
         onClick({ linkName: clickTitle, src });
         // TODO: Get finalized query param keys
         const win = window.open(
-            `https://www.${urlBase}.com/ppcreditapply/da/us?cats_id=DA_AD_UPSTREAM&actor=merchant&mktgrefid=${refId}&payer_id=${payerId}`
+            `${urlBase}/ppcreditapply/da/us?cats_id=DA_AD_UPSTREAM&actor=merchant&mktgrefid=${refId}&payer_id=${payerId}`
```
Seavenly: Is it possible to incorporate this into our `getGlobalUrl` helper so that it follows the same pattern as our other URLs? I see that the local and stage URLs are the same in this scenario, but I don't think that's a big deal and something we could change for the sake of simplicity. If we need to test this URL in the future while running locally we can create some proxy route in the dev server, but that seems unlikely.

README.md
```diff
@@ -18,6 +18,11 @@ Please feel free to follow the [Contribution Guidelines](./CONTRIBUTING.md) to c
 
 ### Quick Setup
 
+#### For PayPal Internal Testing Only:
+
+-   Make a copy of `.env.boilerplate` and rename the copy to `.env`.
+-   Set `STAGE_URL` to be the base url name for internal stage testing (exclude https://, www., and .com)
```
Seavenly: We should consider including the `www.` and `.com` so that the value is a full host value instead of a partial.

src/components/modal/lib/hooks/applyNow.js
```diff
@@ -7,17 +7,17 @@ export default clickTitle => {
...
 
     const urlBase = {
-        local: 'msmaster.qa.paypal',
-        stage: 'msmaster.qa.paypal',
-        sandbox: 'sandbox.paypal',
-        production: 'paypal'
+        local: __MESSAGES__.__DOMAIN__.__STAGE__,
+        stage: __MESSAGES__.__DOMAIN__.__STAGE__,
+        sandbox: 'https://www.sandbox.paypal.com',
+        production: 'https://www.paypal.com'
```
Seavenly: Can we make the same change here to use `getGlobalUrl`?

src/utils/sdk.js
```diff
@@ -210,6 +210,9 @@ export function getPayPalDomain() {
     } else if (__MESSAGES__.__TARGET__ === 'SDK') {
         return getSDKPayPalDomain();
     } else {
+        if (getEnv() === 'stage') {
+            return window?.__TEST_ENV__;
+        }
```
Seavenly: Will this scenario ever occur? Wouldn't this check above always catch it?
``` js
const testEnviroment = window.__TEST_ENV__ ?? __MESSAGES__.__TEST_ENV__;
if (testEnviroment && getEnv() !== 'production' && getEnv() !== 'sandbox') {
```

tests/unit/spec/src/components/modal/lib/hooks/applyNow.test.js
```diff
@@ -28,24 +29,28 @@ describe('Apply Now URL hook', () => {
 
     describe('Should return the staging URL', () => {
         test('env = local', () => {
+            __ENV__ = 'local';
             applyNow()();
 
             expect(global.open).toBeCalledWith(
-                'https://www.msmaster.qa.paypal.com/ppcreditapply/da/us?cats_id=DA_AD_UPSTREAM&actor=merchant&mktgrefid=1&payer_id=1'
+                `${getGlobalUrl('CREDIT_APPLY')}?cats_id=DA_AD_UPSTREAM&actor=merchant&mktgrefid=1&payer_id=1`
```
Seavenly: Any particular reason we use `getGlobalUrl('CREDIT_APPLY')` here, but hard code the URL in the other tests? Should we just hard code the expected value for both?

Seavenly (approved):

PR #975, opened 2023-08-21
fix: simplify handling of deviceID

gillycheesesteak:
## Description

- Centralizes the handling of deviceID to prevent mismatch between deviceID and treatmentsHash, which is dependent on deviceID
- Enables the treatments component on non-SDK integrations, so deviceID can be populated for those integrations in a common way

## Screenshots

N/A

## Testing instructions

Need to consider all possible states of deviceID in localStorage and different integrations that may be currently relying on the way that has been populated previously

### DIFF
--- a/src/components/message/Message.js
+++ b/src/components/message/Message.js
@@ -8,7 +8,8 @@ import {
     createState,
     parseObjFromEncoding,
     getRequestDuration,
-    getTsCookieFromStorage
+    getTsCookieFromStorage,
+    getOrCreateDeviceID
 } from '../../utils';
 
 const Message = function ({ markup, meta, parentStyles, warnings }) {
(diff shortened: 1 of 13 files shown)


src/components/message/Message.js
```diff
@@ -85,8 +75,8 @@ const Message = function ({ markup, meta, parentStyles, warnings }) {
         activeTags: getActiveTags(button),
         messageRequestId,
         // Utility will create iframe deviceID/ts_cookie values if it doesn't exist.
-        deviceID: isStorageFresh() ? parentDeviceID : getDeviceID(),
         ts: getTsCookieFromStorage(),
+        deviceID: getOrCreateStorageID(),
```
Seavenly: If we're no longer pulling the `deviceID` in the `zoid` component `onReady` hooks for the message or modal does this still serve any purpose? If not can this and other instances for the message and modal code just be completely removed if the treatments component is supposed to be the only frame making that communication?

paypal-rosman: good call out

Seavenly (approved):

paypal-rosman (approved):

PR #977, opened 2023-08-23
feat: track merchants displaying standalone modals

surekhaw:
## Description

logging an event_type: "MORS" to update to trigger dynamic segment update to include merchants who integrate standalone modals via SDK


Seavenly:
Closing for now while the solution is still being determined

PR #985, opened 2023-09-14
feat: updated input on calculator for better accessibility

perco12:
## Description

The screen readers read the correct input types and doesn't read the calculator's input as type "telephone" but still ensures the numerical keyboard comes up on mobile 

## Screenshots

enteramount

Screenshot 2023-09-14 at 11 06 18 AM


erikacolette29:
Looks good, just one spot we might want to fix, this screen shot shows the screen reader text with and without your changes. For some reason i'm seeing "invalid data" with your changes. I think we won't want that experience. 

Screenshot 2023-09-18 at 10 50 01 AM

erikacolette29 (approved):

PR #986, opened 2023-09-14
fix: components render inside external zoid iframes

Seavenly:
## Description

We currently have a util function `isZoidComponent()` that we use to prevent certain logic from running inside or IFrames, such as scanning the page for elements with the `data-pp-message` attribute to auto-render messages. This check looks for `__zoid__` windows to prevent this. This has a side-effect of preventing other merchants from using `zoid` to render our components inside an IFrame. The following change has been made to help address this:

- Update the `isZoidComponent()` util to explicitly look for the tags of our `zoid` components

## Screenshots

N/A

## Testing instructions

Attached is an example that can be placed into the `demo/` directory to reproduce the scenario.
[paypal-zoid.zip](https://github.com/paypal/paypal-messaging-components/files/12611495/paypal-zoid.zip)

### DIFF
--- a/src/library/zoid/treatments/component.js
+++ b/src/library/zoid/treatments/component.js
@@ -4,6 +4,7 @@ import { node, dom } from '@krakenjs/jsx-pragmatic/src';
 import { getCurrentScriptUID } from '@krakenjs/belter/src';
 
 // Direct imports to avoid import cycle by importing from ../../../utils
+import { TAG } from '../../../utils';
 import {
     getMeta,
     getEnv,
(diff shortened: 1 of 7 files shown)


src/library/zoid/treatments/component.js
```diff
@@ -4,6 +4,7 @@ import { node, dom } from '@krakenjs/jsx-pragmatic/src';
 import { getCurrentScriptUID } from '@krakenjs/belter/src';
 
 // Direct imports to avoid import cycle by importing from ../../../utils
+import { TAG } from '../../../utils';
```
gillycheesesteak: Is the above comment not relevant anymore?

gillycheesesteak (approved):

PR #992, opened 2023-10-02
refactor: enable passing multiple merchant ids

surekhaw:
## Description
enable passing multiple merchant ids to cpnw.
 
## Screenshots
N/A

## Testing instructions
valid merchant ids of 10 or 13 characters should pass when inputted as merchant-id=* and data-merchant-id='id1, id2'. otherwise an appropriate error message will appear in the console.log.

### DIFF
--- a/src/library/zoid/message/validation.js
+++ b/src/library/zoid/message/validation.js
@@ -68,15 +68,22 @@ export default {
     },
     merchantId: ({ props: { merchantId } }) => {
         if (typeof merchantId !== 'undefined') {
+            let invalidId;
             if (!validateType(Types.STRING, merchantId)) {
+                console.log('not a string');
                 logInvalidType('merchantId', Types.STRING, merchantId);
-            } else if (merchantId.length !== 13 && merchantId.length !== 10) {
-                logInvalid('merchantId', 'Ensure the correct Merchant ID has been entered.');
-            } else {
-                return merchantId;
+                invalidId = merchantId;
             }
+            const ids = merchantId.toString().split(',');
+            ids.forEach(id => {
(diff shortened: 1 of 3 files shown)


src/library/zoid/message/validation.js
```diff
@@ -68,15 +68,20 @@ export default {
...
+                invalidId = merchantId;
             }
+            const ids = merchantId.toString().split(',');
+            ids.forEach(id => {
+                if (id.length !== 13 && id.length !== 10) {
+                    logInvalid('merchantId', 'Ensure the correct Merchant ID has been entered.');
+                    invalidId = id;
+                }
+            });
+            return invalidId ? undefined : merchantId;
```
Seavenly: Consider the following which contains a few modifications:
1. Tightens the scope where `invalidId` (renamed to `isInvalid` since we're using it as a boolean) is used to reduce mental overhead
2. The `forEach` will potentially log multiple times with the same message if there are multiple bad IDs but we only need it to log once. We can make use of `Array.prototype.some` to short circuit after a single invalid match. It will return a `boolean` value.
3. More closely follows the pattern of other validators by returning the value in an `else` block after all checks have passed.
``` js
            if (!validateType(Types.STRING, merchantId)) {
                logInvalidType('merchantId', Types.STRING, merchantId);
            } else {
                const isInvalid = merchantId.split(',').some(id => id.length !== 13 && id.length !== 10);
                
                if (isInvalid) {
                    logInvalid('merchantId', 'Ensure the correct Merchant ID has been entered.');
                } else {
                    return merchantId;
                }
            }
```

Seavenly (approved):

PR #994, opened 2023-10-04
fix: prevent undefined device ID when namespace

Seavenly:
## Description

These changes include 2 fixes
1. When using the SDK namespace feature, currently that results in `"undefined"` being passed up to the experiments endpoint because the SDK instance running inside the IFrame does not get passed the custom namespace and will always use the default namespace. To fix this issue we pass the default namespace into the IFrame to be used instead.
2. @danzhaas noticed that under certain circumstances that different device IDs could be sent to different endpoints for the same message instance. We were unable to reproduce this issue consistently so we have theoretical changes that enforce all local storage reading and writing to go through the common `belter` storage helper to ensure that the `belter` storage instance used by the SDK does not get out of sync with our manual mutations of local storage directly.

## Screenshots

N/A

## Testing instructions

N/A

### DIFF
--- a/src/utils/sdk.js
+++ b/src/utils/sdk.js
@@ -1,7 +1,7 @@
 /* eslint-disable eslint-comments/disable-enable-pair, no-else-return */
 import arrayFrom from 'core-js-pure/stable/array/from';
 
-import { isLocalStorageEnabled, getStorage as getBelterStorage } from '@krakenjs/belter/src';
+import { getStorage as getBelterStorage } from '@krakenjs/belter/src';
 import { SDK_QUERY_KEYS, SDK_SETTINGS } from '@paypal/sdk-constants/src';
 import {
     getClientID,
@@ -14,8 +14,10 @@ import {
     getSDKQueryParam,
     getCSPNonce,
     getNamespace as getSDKNamespace,
+    getDefaultNamespace as getDefaultSDKNamespace,
     getSessionID as getSDKSessionID,
     getStorageID as getSDKStorageID,
+    getStorageState as getSDKStorageState,
     getPayPalDomain as getSDKPayPalDomain,
     getDisableSetCookie as getSDKDisableCookie
 } from '@paypal/sdk-client/src';
@@ -116,11 +118,19 @@ export function getScriptAttributes() {
     }
 }
 
+export function getDefaultNamespace() {
+    if (__MESSAGES__.__TARGET__ === 'SDK') {
+        return getDefaultSDKNamespace();
+    } else {
+        return 'paypal';
+    }
+}
+
 export function getNamespace() {
     if (__MESSAGES__.__TARGET__ === 'SDK') {
         return getSDKNamespace();
     } else {
-        return getScript()?.getAttribute('data-pp-namespace') || 'paypal';
+        return getScript()?.getAttribute('data-pp-namespace') || getDefaultNamespace();
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/utils/sdk.js
```diff
@@ -116,11 +118,19 @@ export function getScriptAttributes() {
...
+        return 'paypal';
+    }
+}
+
 export function getNamespace() {
     if (__MESSAGES__.__TARGET__ === 'SDK') {
         return getSDKNamespace();
     } else {
-        return getScript()?.getAttribute('data-pp-namespace') || 'paypal';
+        return getScript()?.getAttribute('data-pp-namespace') || getDefaultNamespace();
```
danzhaaspaypal, line 133: i believe we have been seeing this attribute labelled as `data-namespace` not `data-pp-namespace`, though perhaps we should cover both bases

danzhaaspaypal: further investigation: a confluence search returns only 1 example of `data-pp-namespace` (from our documentation no less) vs 55 examples of `data-namespace`

Seavenly: `data-namespace` will be covered via the `if` branch directly above https://github.com/paypal/paypal-messaging-components/blob/4ef68240772a58a4a178a4d6b1093d0901111a70/src/utils/sdk.js#L131 
The `else` is for the standalone variant that is commonly used when developing locally. The SDK script uses `data-*` prefixes, but our library uses `data-pp-*`.

src/utils/sdk.js
```diff
@@ -157,23 +167,12 @@ export function getOrCreateDeviceID() {
...
-              `__${getNamespace()}_storage__`,
-              JSON.stringify({
-                  ...getRawStorage(),
-                  ...values
-              }) ?? '{}'
-          )
-        : {};
+export function updateStorage(values) {
+    if (__MESSAGES__.__TARGET__ === 'SDK') {
+        return getSDKStorageState(storage => Object.assign(storage, values));
```
danzhaaspaypal, line 172: it looks like `getSDKStorageState` is going to always return the value defined as the SDK's `data-namespace` within its localStorage name (as long as it's defined).  Found that by diving into @paypal/sdk-client and @paypal/sdk-constants from here:
Screenshot 2023-10-09 at 4 06 21 PM

danzhaaspaypal: Screenshot 2023-10-09 at 4 13 15 PM
Screenshot 2023-10-09 at 4 13 25 PM

Seavenly: That's what we want and expect though, right?

danzhaaspaypal: no: we found that the sdk, even when it has a namespace specified, will write its deviceID to __paypal_storage__ . So if we want consistent naming of our storage in both the paypal domain and merchant domain local storage, we would have to commit to using __paypal_storage__ for all our local storage names and never the data-namespace.

danzhaaspaypal: If I understand right, the desired outcome is that the merchant and the paypal localStorage use the same storage name.

Here's a short summary of the current behavior, and the behavior with this PR as-is:

This url will show the namespace situation as-is: given an SDK namespace of PayPalSDK, messaging.js will be incosistent with its localStorage use. 
http://localhost:8000/_personal/dan/localStorage/?env=stage&namespace=PayPalSDK
	• Localhost (Merchant) localStorage shows 
		○ `PayPalSDK_storage` complete with id and experiments.
	• Msmaster (paypal) localStorage shows 
		○ `paypal_storage` has id but, no experiments 
		○ `PayPalSDK_storage` has experiments, but no id.

This url will show this PR's namespace fix working: 
http://localhost:8000/_personal/dan/localStorage/?env=stage&namespace=PayPalSDK&stageTag=nate_device_id_0&testEnv=te-nschott-1122
	• Localhost (Merchant) localStorage shows 
		○ `PayPalSDK_storage` complete with id and experiments.
	• Msmaster (paypal) localStorage shows 
		○ `paypal_storage` with the same id and treatments

danzhaaspaypal: update: after discussing in slack, i change my answer and would say this PR is doing what we want and expect.

Without the PR, here is the behavior of the localStorage:
	• Localhost (Merchant) localStorage shows 
		○ `PayPalSDK_storage` complete with id and experiments.
	• Msmaster (paypal) localStorage shows 
		**○ `paypal_storage` has id but, no experiments 
		○ `PayPalSDK_storage` has experiments, but no id.**

With this PR's namespace fix working: 
	• Localhost (Merchant) localStorage shows 
		○ `PayPalSDK_storage` complete with id and experiments.
	• Msmaster (paypal) localStorage shows 
		**○ `paypal_storage` with the same id and treatments**

danzhaaspaypal (approved):

jadutter (approved):

PR #997, opened 2023-10-09
fix: safari focus

jadutter:
## Description
- Improve accessibility
    - adjusted how focus is handled when the modal opens and closes
    - added some aria attributes to increase accessibility
    - Updated the prerender modal close button to receive focus when opened
- Updated the prerender modal 
    - close button to look identical to the modal close button
    - cleaned up some lint issues
- Removed unused code `useAutoFocus`
- ~~Improve debugging~~
    - ~~added debug levels to control logging verbosity~~
    - ~~added debug messages and listeners to view events and their order~~
    - split into a [separate PR](https://github.com/paypal/paypal-messaging-components/pull/998)


src/components/modal/v2/lib/providers/transition.js
```diff
@@ -27,8 +27,10 @@ export const TransitionStateProvider = ({ children }) => {
      * Particularly useful for those using screen readers and other accessibility functions.
      */
     const focusCloseBtnOnModalOpen = () => {
-        const btn = document.querySelector('.close');
-        btn?.focus();
+        // give the document time to update before trying to focus the close button
+        setTimeout(() => {
```
Seavenly: Is there an event we should be looking for like the document `DOMContentLoaded` or window `load` event? If so would we be able to leverage these either of these helpers? https://github.com/paypal/paypal-messaging-components/blob/develop/src/utils/events.js#L11-L21. If we aren't waiting for a specific DOM event and just need to ensure the browser has had a render cycle, could we use `requestAnimationFrame` instead? The primary concern with `setTimeout` is it is an arbitrary time that may or may not fire the same based on device performance.

src/components/modal/v2/parts/BodyContent.jsx
```diff
@@ -13,6 +13,15 @@ import {
...
 import { LongTerm, ShortTerm, NoInterest, ProductList, PayIn1 } from './views';
 
+const VIEW_IDS = {
+    // TODO: add an error view in case we receive an invalid view?
+    PAYPAL_CREDIT_NO_INTEREST: 'PAYPAL_CREDIT_NO_INTEREST',
+    PAY_LATER_LONG_TERM: 'PAY_LATER_LONG_TERM',
+    PAY_LATER_PAY_IN_1: 'PAY_LATER_PAY_IN_1',
+    PAY_LATER_SHORT_TERM: 'PAY_LATER_SHORT_TERM',
+    PRODUCT_LIST: 'PRODUCT_LIST'
+};
```
Seavenly, line 23: Are we able to leverage the `OFFER` constant at all in `utils/constants` to prevent potential duplicate hard-coded string references?

src/components/modal/v2/parts/BodyContent.jsx
```diff
@@ -95,7 +104,10 @@ const BodyContent = () => {
                 viewName={viewName}
             />
             <div className="content__container">
-                <main className="main">
+                <main
+                    className="main"
+                    aria-label={`More info on the Pay Later offer${viewName === VIEW_IDS.PRODUCT_LIST ? 's' : ''}.`}
```
Seavenly: We should not be hard-coding content directly with the code since this will be used for non-US countries. If we need this content then we should consider formally incorporating it into our content.

src/library/zoid/modal/containerTemplate.jsx
```diff
@@ -47,6 +47,7 @@ export default ({ uid, frame, prerenderFrame, doc, event, state, props: { cspNon
             replaceViewport();
             setTimeout(() => {
                 wrapper.classList.add(CLASS.HIDDEN);
+                wrapper.querySelector('iframe')?.blur();
```
Seavenly: Are we able to use the `frame` reference passed in with the original top level function instead of querying for the iframe?

src/library/zoid/modal/containerTemplate.jsx
```diff
@@ -60,8 +61,10 @@ export default ({ uid, frame, prerenderFrame, doc, event, state, props: { cspNon
             ZalgoPromise.delay(TRANSITION_DELAY)
                 .then(() => overlay.classList.add(CLASS.TRANSITION))
                 .then(() => ZalgoPromise.delay(TRANSITION_DELAY))
-                .then(() => destroyElement(prerenderFrame));
+                .then(() => destroyElement(prerenderFrame))
+                .then(() => wrapper.querySelector('iframe')?.focus());
```
Seavenly: Does this compete with focusing on the close button? Curious if a potential race condition is possible here that might result in unfocusing the close button?

src/components/modal/v2/lib/providers/transition.js
jadutter: While manually testing in Safari, I found that
1. The the close button for the modal needed the `autofocus` attribute
1. The `iframe` for the modal in the parent page needed to receive focus
1. The close button for the modal needed to receive focus

The events appeared to be happening in the correct order, but from what I could tell, the iframe was going to be given focus, but had not yet received focus (I found something similar happening with the [tabTrap](https://github.com/paypal/paypal-messaging-components/pull/998/files#diff-68f8cf9ab239f8fcb4755f7547baad90fc6528f6762eb343690ff66857b7917fR75-R80)). 

I thought about setting up listeners to ensure the button focused happened after the `iframe`, but I realized I would need
1.  on the parent page, I would need to listen for when the iframe received focus, then pass the info onto 
1. the zoid event emitter in the parent page, the pass the info onto 
1. a listener in the child iframe, that would finally add focus to the button

I thought about it some more, and I figured the direct pipeline I would setup would be unideal. We are already using `MODAL_EVENT.MODAL_SHOW` to communicate almost the same thing. Creating a pipelline for `MODAL_EVENT.FOCUS` seemed like a lot of extra code to add onto messaging just to focus the close button. 

I found that bumping it to a different place in the event cycle was enough for it to behave correctly though, not necessarily needing to wait literally 10ms. 

I'll change it to `requestAnimationFrame` since it appears to work too.

jadutter: After further testing, I've realized my PR only partially fixed the issue:
- modal open
- modal close 
- message has focus
- modal open 
- modal close
- modal still has focus


Going to keep working on it

src/components/modal/v2/parts/BodyContent.jsx
jadutter: `OFFER` does not contain `PRODUCT_LIST`, since `PRODUCT_LIST` is not an offer. I decided to lean towards minimal changes, replacing hard coding in the one file with an enum.

src/components/modal/v2/parts/BodyContent.jsx
jadutter: While navigating with VoiceOver, I noticed it would read `main`. I thought adding a label here would be helpful, but you're right. I hadn't considered this would code would be used for other languages. I'll remove it.

src/library/zoid/modal/containerTemplate.jsx
jadutter: Focus is different depending on the window it's called within. 

Given:
In the parent page
```
document.activeElement.title === 'PayPal Message 1' 
``` 
In the modal iframe page
```
document.activeElement.tagName === 'BODY' 
``` 

Then executing this in the modal iframe page
```
document.querySelector('#close-btn').focus()
``` 

results in no change since the iframe for the modal does not have focus in the parent page.

src/components/modal/v2/lib/providers/transition.js
jadutter: I've gotten it to a decent state, I think. The two caveats I'm thinking of are: 
1. We'll need to revisit making sure the prerender modal is accessible. I tried to clean it up a bit, but 
    1. I couldn't get its close button to receive focus after opening, closing, and opening again. 
    1. I wasn't at all certain how to effectively communicate via accessibility tags to a non-visual user "you are on a modal that is currently loading content" or announce "the modal has failed to load" 
1. The modal button receives focus in Chrome, but only the modal receives focus in Safari. Apparently, Apple intends this as a "feature"; the user must directly interact with the iframe before any elements in the iframe can be given focus. After the user interacts with the elements within the iframe for the modal, then we can call `focus()` on the close button. Since the modal can still be navigated after the iframe for the modal receives focus, I was going to let this one go

grablack (approved):

src/components/modal/v2/lib/providers/transition.js
```diff
@@ -41,6 +43,7 @@ export const TransitionStateProvider = ({ children }) => {
                         if (entry.isIntersecting) {
                             // Removes .modal-closed class from modal iframe body when modal is open.
                             document.body.classList.remove('modal-closed');
+                            document.body.classList.add('modal-open');
```
Seavenly: It looks like this is always the reverse of `.modal-closed`. Does this add any extra functionality or would it be safe to remove?

src/library/zoid/message/component.js
```diff
@@ -162,7 +162,9 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
                             refIndex: index,
                             src: 'message_click',
                             onClose: () => {
-                                getContainer().querySelector('iframe').focus();
+                                window.requestAnimationFrame(() => {
```
Seavenly: Can we add a comment why this is beneficial or would it be safe to remove? The message `iframe` element will have already been available in the DOM for awhile at this point, so I just want to ensure it doesn't accidentally get removed in the future and cause an unintended side-effect.

src/library/zoid/modal/containerTemplate.jsx
```diff
@@ -47,21 +56,36 @@ export default ({ uid, frame, prerenderFrame, doc, event, state, props: { cspNon
...
+                    } else {
+                        previousFocus.focus();
+                    }
+                });
             }, TRANSITION_DELAY);
         };
 
         const handleEscape = evt => {
-            if (state.open && (evt.key === 'Escape' || evt.key === 'Esc' || evt.charCode === 27)) {
+            if (state.open && (`${evt?.key}`.toLowerCase().startsWith('esc') || evt.charCode === 27)) {
```
Seavenly: Were there keys that weren't being being accounted for that this change now covers?

src/library/zoid/modal/containerTemplate.jsx
```diff
@@ -30,13 +31,21 @@ export default ({ uid, frame, prerenderFrame, doc, event, state, props: { cspNon
...
             state.open = true;
+            previousFocus = document.activeElement;
             wrapper.classList.remove(CLASS.HIDDEN);
             hijackViewport();
             // Browser needs to repaint otherwise the transition happens immediately
             // Firefox requires 2 RAFs due to where they are called in the event loop
             requestAnimationFrame(() => {
                 requestAnimationFrame(() => {
                     overlay.classList.add(CLASS.MODAL_SHOW);
+                    requestAnimationFrame(() => {
```
Seavenly: This is already nested within 2 RAF called. Does this 3rd one provide any extra benefit?

src/library/zoid/modal/containerTemplate.jsx
```diff
@@ -30,13 +31,21 @@ export default ({ uid, frame, prerenderFrame, doc, event, state, props: { cspNon
...
+                        if (renderedModal) {
+                            frame.focus();
+                        } else if (window.document.activeElement !== prerenderFrame) {
```
Seavenly: Is there any downside to just simplify this as an `else` and just always call `prerenderFrame.focus()`?

src/library/zoid/modal/containerTemplate.jsx
```diff
@@ -47,21 +56,36 @@ export default ({ uid, frame, prerenderFrame, doc, event, state, props: { cspNon
             replaceViewport();
             setTimeout(() => {
                 wrapper.classList.add(CLASS.HIDDEN);
+
+                requestAnimationFrame(() => {
```
Seavenly: Since this is occurring after a substantial delay already, does the RAF provide any extra benefit?

src/library/zoid/modal/containerTemplate.jsx
```diff
@@ -47,21 +56,36 @@ export default ({ uid, frame, prerenderFrame, doc, event, state, props: { cspNon
...
+                    if (renderedModal) {
+                        frame.blur();
+                    } else {
+                        previousFocus.focus();
+                    }
```
Seavenly: Do we not want both of these things to occur when the modal has hidden? Remove focus from the iframe and pass focus back to the previously focused element?

src/library/zoid/modal/containerTemplate.jsx
```diff
@@ -47,21 +56,36 @@ export default ({ uid, frame, prerenderFrame, doc, event, state, props: { cspNon
...
 
         const handleTransition = () => {
+            renderedModal = true;
             ZalgoPromise.delay(TRANSITION_DELAY)
                 .then(() => overlay.classList.add(CLASS.TRANSITION))
                 .then(() => ZalgoPromise.delay(TRANSITION_DELAY))
-                .then(() => destroyElement(prerenderFrame));
+                .then(() => destroyElement(prerenderFrame))
+                .then(() => {
+                    if (state.open && document.activeElement !== frame) {
```
Seavenly, line 73: Does this prevent stealing focus from inside the IFrame when an element inside has focus, or is it meant to be an extra cautionary guard?

src/library/zoid/modal/prerenderTemplate.jsx
```diff
@@ -119,36 +119,109 @@ export default ({ doc, props, event }) => {
...
+            });
+        }
+    };
+
+    const handleClose = () => {
+        event.trigger('modal-hide');
+    };
+
+    const handleEscape = evt => {
+        if (!renderedModal && state.open && (`${evt?.key}`.toLowerCase().startsWith('esc') || evt?.charCode === 27)) {
```
Seavenly: Similar comment as above.

src/library/zoid/modal/prerenderTemplate.jsx
```diff
@@ -119,36 +119,109 @@ export default ({ doc, props, event }) => {
...
+
+    const handleRender = element => {
+        closeBtn = element.querySelector('#prerender-close-btn');
+        focusCloseButton();
+        ZalgoPromise.delay(ERROR_DELAY).then(() => {
+            return checkForErrors(element);
+        });
+    };
+
+    event.on('zoid-rendered', () => {
```
Seavenly: Is this an official `zoid` event that could be referenced?
``` js
import { EVENT } from '@krakenjs/zoid/src';
```

src/components/modal/v2/lib/providers/transition.js
jadutter: Yes. It helps to distinguish between:
- premodal is open -> swtiches to modal 
    - modal will not have any classes
- modal is closed -> switches to modal open 

IIRC, this was needed to get the modal close button to focus after switching from premodal to modal.

src/library/zoid/modal/containerTemplate.jsx
jadutter: No, just a general distrust of strings. Reading `evt.key === 'Escape' || evt.key === 'Esc' || evt.charCode === 27` lead me to believe we don't trust all browsers and keyboards to behave the same way, otherwise we'd only be matching `evt.charCode === 27`. I took it a step further to guard against any other shenanigans. 

I'll undo if it if you think the other way is better.

src/library/zoid/modal/containerTemplate.jsx
jadutter: IIRC, if the iframe for the prerender modal has focus in the parent page, and the close button within the prerender modal has focus in the child page, then calling focus on the iframe in the parent page updates the focus on the child page from the button to the body.

src/library/zoid/message/component.js
jadutter: I'll double check each of the RAFs and provide an update.

src/library/zoid/modal/containerTemplate.jsx
jadutter: `renderedModal === true` after we get the `zoid-rendered` event for the modal.

While the modal is not present, we need the prerender modal to return focus back to the message when it closes. However, the modal is already setup to focus the message after closing over [here](https://github.com/paypal/paypal-messaging-components/pull/997/files/97368f82f6fb0b4379dfb6281694a7d7a948e97b#diff-101b681d25c91ec487b69b2c0a172cf98d685c8027be25e9381e37d771eacb28R166).

src/library/zoid/modal/containerTemplate.jsx
jadutter: > Does this prevent stealing focus from inside the IFrame when an element inside has focus

Yes, similar to [here](https://github.com/paypal/paypal-messaging-components/pull/997#discussion_r1358446923)

src/components/modal/v2/lib/providers/transition.js
Seavenly: Can we confirm this? When testing locally I am able to remove the class without impacting focusing but there could be a case I'm not familiar with. If we're able to remove the class I think we should consider it to reduce noise. It appears that `.modal-closed` is already being used for the same thing `.modal-open` would be used for, but `.modal-open` doesn't appear to be tied to any CSS. It looks like we're just using it to declare state on the element which is likely unnecessary.

src/library/zoid/modal/containerTemplate.jsx
Seavenly: Since these values are predefined I think it's better to use the explicit values which acts a loose self documentation, otherwise it can lead to a general lack of clarity when reading this at a future date. The multiple `||` checks is definitely related to supporting older browsers. `evt.key === 'Escape'` is the officially preferred way moving forward. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values) it looks like `evt.key === 'Esc'` is from a really old version of Firefox that we can likely just drop, and the [charCode](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/charCode) is to support older browsers as well.

src/library/zoid/modal/containerTemplate.jsx
Seavenly: If this can steal focus from inside the IFrame, would it be better to not have any logic on the parent page that focuses the modal IFrames at all, and just let the code inside the IFrames handle the focus?

src/library/zoid/modal/containerTemplate.jsx
Seavenly: Would it potentially be better if we tried to handle shifting focus back to the message the same way for both the prerender and regular modal? So if we removed the manual `onClose` we're doing and instead here just always call `previousFocux.focus()` and remove the if/else?

src/library/zoid/modal/containerTemplate.jsx
```diff
@@ -17,6 +17,8 @@ export default ({ uid, frame, prerenderFrame, doc, event, state, props: { cspNon
     // cannot overlay across the entire screen
     if (context === 'popup') return undefined;
 
+    let renderedModal = false;
+    let previousFocus = document.activeElement;
```
Seavenly: Should we consider including these on `state` so that they are easily accessible from both the prerender template and container template, and so we don't have to try and keep them in sync between the two different files? One file should be able to control setting the values while the other just reads from them. It may improve intent as well since `state.*` implies variables that will update throughout the zoid components lifecycle.

src/components/modal/v2/lib/providers/transition.js
jadutter: I wasn't able to consistently reproduce the issue locally anymore, so I'm ripping out `.modal-open`.

src/library/zoid/modal/containerTemplate.jsx
jadutter: No. See https://github.com/paypal/paypal-messaging-components/pull/997#discussion_r1352845500

src/library/zoid/modal/containerTemplate.jsx
```diff
@@ -47,21 +54,29 @@ export default ({ uid, frame, prerenderFrame, doc, event, state, props: { cspNon
             replaceViewport();
             setTimeout(() => {
                 wrapper.classList.add(CLASS.HIDDEN);
+                state.previousFocus.focus();
             }, TRANSITION_DELAY);
         };
 
         const handleEscape = evt => {
-            if (state.open && (evt.key === 'Escape' || evt.key === 'Esc' || evt.charCode === 27)) {
+            if (state.open && (evt?.key === 'Escape' || evt?.key === 'Esc' || evt.charCode === 27)) {
```
Seavenly: Is there a specific scenario we are aware of where `evt` is `undefined` that would benefit from `evt?.*`? If not my general concern with overly eager usage of `?` can lead to lack of certainty in the existence of objects when reading through the code and lead to a practice of just including `?` for any object property accessing which leads to unnecessary overhead.

src/library/zoid/modal/prerenderTemplate.jsx
```diff
@@ -119,36 +119,127 @@ export default ({ doc, props, event }) => {
...
+
+    const handleClose = () => {
+        event.trigger('modal-hide');
+    };
+
+    const handleEscape = evt => {
+        if (
+            !state.renderedModal &&
+            state.open &&
+            (evt?.key === 'Escape' || evt?.key === 'Esc' || evt.charCode === 27)
```
Seavenly: Similar comment as above

src/library/zoid/modal/containerTemplate.jsx
```diff
@@ -30,13 +31,19 @@ export default ({ uid, frame, prerenderFrame, doc, event, state, props: { cspNon
...
             // Browser needs to repaint otherwise the transition happens immediately
             // Firefox requires 2 RAFs due to where they are called in the event loop
             requestAnimationFrame(() => {
                 requestAnimationFrame(() => {
                     overlay.classList.add(CLASS.MODAL_SHOW);
+                    if (state.renderedModal) {
+                        frame.focus();
+                    } else if (window.document.activeElement !== prerenderFrame) {
+                        prerenderFrame.focus();
+                    }
```
Seavenly: Should we guard `frame.focus()` with a similar `window.activeElement !== frame`? From your findings it appears that may be best practice to prevent accidentally altering the focus.

src/library/zoid/modal/prerenderTemplate.jsx
```diff
@@ -119,36 +119,127 @@ export default ({ doc, props, event }) => {
...
+            handleClose();
+        }
+    };
+
+    const handleRender = element => {
+        closeBtn = element.querySelector('#prerender-close-btn');
+        // we need to give chrome a moment before we can focus the close button
+        window.requestAnimationFrame(() => {
+            focusCloseButton();
+        });
```
Seavenly: Is there a specific reason we wrap `focusCloseButton()` in 1 RAF here, but below we use 2?

src/library/zoid/modal/prerenderTemplate.jsx
```diff
@@ -119,36 +119,127 @@ export default ({ doc, props, event }) => {
...
+                // TODO: should we report this failure to our log endpoint?
             }
         });
     };
 
+    const focusCloseButton = () => {
+        if (closeBtn) {
+            closeBtn.focus();
+        }
+    };
```
Seavenly: Since this helper function is so simple, should we consider just inlining `closeBtn?.focus()` where it's used instead to reduce mental overhead?

src/library/zoid/modal/prerenderTemplate.jsx
jadutter: That's the minimum RAF I found it needed in each case when manually testing.

src/library/zoid/modal/containerTemplate.jsx
```diff
@@ -17,6 +17,7 @@ export default ({ uid, frame, prerenderFrame, doc, event, state, props: { cspNon
     // cannot overlay across the entire screen
     if (context === 'popup') return undefined;
 
+    state.previousFocus = document.activeElement;
```
Seavenly: Since this runs as part of the container template initialization before a message has been interacted with does this capture anything meaningful, or is it safe to remove and let the `handleShow` set the value?

Seavenly (approved):

PR #1018, opened 2023-11-07
feat: update lander modal's onClose to sends postMessage in iFrame

perco12:
## Description

As a merchant. when I click esc on the lander , it sends a onClose postMessage to the parent window


src/components/modal/v2/lib/zoid-polyfill.js
```diff
@@ -83,6 +83,20 @@ const setupBrowser = props => {
             });
         },
         onClose: ({ linkName }) => {
+            function getParentOrigin(url) {
+                if (!url) return window.location.origin;
```
Seavenly: We would only need to send a `postMessage` if we're in an IFrame right? So if there's no `document.referrer` value then we know we're running as a full webpage and can just skip sending the `postMessage`

src/components/modal/v2/lib/zoid-polyfill.js
```diff
@@ -83,6 +83,20 @@ const setupBrowser = props => {
...
+
+                const domain = url.indexOf('/', url.indexOf('//') + 2);
+
+                if (domain > -1) {
+                    return url.slice(0, domain);
+                }
+                return url;
+            }
+
+            const targetOrigin = getParentOrigin(document.referrer);
```
Seavenly: Did we consider leveraging the `URL` class to reduce the need for our own logic?
``` js
const targetOrigin = new URL(document.referrer).origin;
```

src/components/modal/v2/lib/zoid-polyfill.js
```diff
@@ -83,6 +83,20 @@ const setupBrowser = props => {
...
+            window.parent.postMessage('paypal-messages-modal-close', targetOrigin);
```
Seavenly: If this call is optional based on whether or not we're inside an IFrame, would it make sense to wrap this in something like
``` js
if (window.parent !== window && document.referrer) {
    const targetOrigin = new URL(document.referrer).origin;
    window.parent.postMessage('paypal-messages-modal-close', targetOrigin);
}
```

tests/unit/spec/src/components/modal/v2/lib/zoid-polyfill.test.js
```diff
@@ -151,6 +151,19 @@ describe('zoidPollyfill', () => {
             })
         );
         logger.track.mockClear();
+
+        window.xprops.onClose({ linkName: 'Escape Key' });
+
+        expect(logger.track).toHaveBeenCalledTimes(1);
```
Seavenly: Can we also include an `expect` assertion for the `window.parent.postMessage` call that we're making?

tests/unit/spec/src/components/modal/v2/lib/zoid-polyfill.test.js
```diff
@@ -1,5 +1,6 @@
 import zoidPolyfill from 'src/components/modal/v2/lib/zoid-polyfill';
 import { logger } from 'src/utils';
+import { isIframe } from '../../../../../../../../src/components/modal/v2/lib/utils';
```
grablack: We can cut this down to 'src/components/modal/v2/lib/utils'. Same for the jest.mock path

tests/unit/spec/src/components/modal/v2/lib/zoid-polyfill.test.js
```diff
@@ -151,6 +155,49 @@ describe('zoidPollyfill', () => {
             })
         );
         logger.track.mockClear();
+
```
grablack: Can you add some comments to group the new test you added? This test file is pretty unwieldy because we have everything in 3 test functions. Hopefully comments will help

grablack (approved):

PR #1028, opened 2023-12-07
feat: update modal v4Design

surekhaw:
## Description
update v4 modal design

## Screenshots
TBD

## Testing instructions
TBD

### DIFF
--- a/src/components/modal/v2/styles/globals/_colors.scss
+++ b/src/components/modal/v2/styles/globals/_colors.scss
@@ -12,8 +12,6 @@ $offer-border-gray: #c6c6c6;
 $beige: #faf8f5;
 // Lander background
 $lander-beige: #e6e0d9;
-// v4.0 lander background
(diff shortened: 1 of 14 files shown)


src/components/modal/v2/styles/globals/_colors.scss
```diff
@@ -12,6 +12,8 @@ $offer-border-gray: #c6c6c6;
 $beige: #faf8f5;
 // Lander background
 $lander-beige: #e6e0d9;
+// v4.0 lander background
+$v4-lander-beige: #f6f6f6;
```
yeul: Since we use the `lander-` and non `lander-` styles separately, let's do the same here. So for anywhere we use `$lander-beige` we can create a new color called `$v4-lander-gray` or something like that since the color appears more gray than beige this time around. And then we can create a new color under `$beige` and call it `$v4-gray` for areas where we are using `$beige`.

PR #1044, opened 2024-01-16
feat:initial addition of new offer and updated validation

perco12:
## Description

Update OFFERS with new offers and update offer validation to support an array of strings along with singular string values


tests/unit/spec/src/zoid/message/validation.test.js
```diff
@@ -136,6 +136,19 @@ describe('validate', () => {
             expect(console.warn).not.toHaveBeenCalled();
         });
 
+        // multiple offers passed in test
```
surekhaw, line 142: should we test if 'REWARDS' or 'PAY_LATER' is passed in as a string as opposed to within in an array?

tests/unit/spec/src/zoid/message/validation.test.js
```diff
@@ -136,6 +136,19 @@ describe('validate', () => {
...
+
+        [
+            { offer: ['PAY_LATER', 'PAY_LATER_SHORT_TERM'], supportedOffer: 'PAY_LATER,PAY_LATER_SHORT_TERM' },
+            { offer: ['PAY_LATER', 'REWARDS'], supportedOffer: 'PAY_LATER,REWARDS' },
+            { offer: ['REWARDS', 'PAY_LATER_LONG_TERM'], supportedOffer: 'REWARDS,PAY_LATER_LONG_TERM' }
+        ].forEach(supportedOffer => {
```
surekhaw: any concern that `supportedOffer` could possibly lead to confusion as key name and also param?

surekhaw (approved):

JustinDoan (approved):

PR #1058, opened 2024-03-08
fix: setup global state from script in modal setup

Seavenly:
## Description

- Modal setup will now setup global state similar to the message to account for script attributes such as `data-pp-namespace`

## Screenshots

N/A

## Testing instructions

- Use `data-pp-attribute` to confirm that it respects the value by placing the standalone modal on `window.<namespace>`
- Confirm that a value of `paypal` does not cause adverse affects


danzhaaspaypal (approved):

src/library/controllers/message/setup.js
```diff
@@ -23,25 +19,12 @@ export default function setup() {
...
-            ...inlineScriptOptions
-        });
-    }
+    setupGlobalState();
 
     const { namespace } = getGlobalState().config;
 
     // Allow specified global namespace override
-    if (namespace) {
+    if (namespace && namespace !== 'paypal') {
```
danzhaaspaypal: should we consider using `getDefaultNamespace` instead of writing out this remotely set constant value?

src/library/controllers/modal/setup.js
```diff
@@ -13,10 +14,12 @@ export default function setup() {
     // Load the zoid components into memory so that the zoid interface can bootstrap between parent and child
     getModalComponent();
 
+    setupGlobalState();
+
     const { namespace } = getGlobalState().config;
 
     // Allow specified global namespace override
-    if (namespace) {
+    if (namespace && namespace !== 'paypal') {
```
danzhaaspaypal: same as above

tests/unit/spec/src/controllers/message/setup.test.js
```diff
@@ -62,7 +63,7 @@ describe('message setup', () => {
     test('Renames the global namespace', () => {
         const removeMockScript = insertMockScript({ account: 'DEV00000000NI', namespace: 'paypal2' });
 
-        expect(window.paypal).toEqual(expect.any(Object));
+        expect(window.paypal).toBeUndefined();
```
danzhaaspaypal, line 66: shouldn't we be checking `window.paypal2` since the namespace was set to `paypal2` on line 64?

danzhaaspaypal (approved):

PR #1065, opened 2024-03-20
feat: improve ql xo ax

surekhaw:
## Description

update screenreader text for ql xo modals

## Testing instructions
N/A


content/offers/US/v2_long-term_checkout.json
```diff
@@ -0,0 +1,43 @@
+{
```
yeul, line 1: If I'm looking at this right, it looks like you copied the content from the content repo into this file. This is likely what is causing the CI to fail on the PM XO modal because how we handle variables is different between our devServerProxy and actual live content and in our proxy config we're pointing to this file instead of `v2_long_term.json`. You'll want to replace the entire `meta` object with what's in `v2_long_term.json` here: https://github.com/paypal/paypal-messaging-components/blob/8240d03237dc159a74979805df11c4fde4fe9c15/content/offers/US/v2_long_term.json#L2-L29

tests/functional/v2/testFn/longTerm.js
```diff
@@ -46,7 +46,7 @@ export const showCorrectOfferInfo = async (contentWindow, modalContent, testName
     await contentWindow.waitForSelector(contentWrapper);
     await contentWindow.waitForSelector(`${offerRow}:first-child`);
     const offerFieldHeadline = await contentWindow.$eval(offerRow, element => element.innerText);
-    const newOfferFieldHeadline = offerFieldHeadline.replace('month.', '').replace(/\n/g, '');
+    const newOfferFieldHeadline = offerFieldHeadline.replace('per month', '').replace(/\n/g, '');
```
yeul:
```suggestion
    const newOfferFieldHeadline = offerFieldHeadline.replace('month.', '').replace('per month', '').replace(/\n/g, '');
```
Small update again for all instances of this regex in the file. I forgot that because different accounts use this same test file we have to account for both cases of the content (checkout and non-checkout).

yeul (approved):

PR #1078, opened 2024-04-23
feat: add button session id to modals

paypal-rosman:
## Description
add button session id to modals


src/utils/logger.js
```diff
@@ -27,8 +27,7 @@ function generateLogPayload(account, { meta, events: bizEvents, tracking }) {
     const components = Object.entries(meta)
         .filter(([, component]) => component.account === account)
         .map(([index, component]) => {
-            const { type, messageRequestId, stats = {}, trackingDetails } = component;
-
+            const { type, buttonSessionId = undefined, messageRequestId, stats = {}, trackingDetails } = component;
```
Seavenly: Do we need `= undefined`? That should already be the case when destructuring from an object.

paypal-rosman: fixed it, i thought attempting destructuring a variable that is undefined would cause a failure. I added a comment just in case

danzhaaspaypal (approved):

PR #1087, opened 2024-05-09
fix: update API domain

JordanClemons:
## Description

- Fix API domain to not include CORS

### DIFF
--- a/src/utils/sdk.js
+++ b/src/utils/sdk.js
@@ -237,6 +237,8 @@ export function getPayPalAPIDomain() {
         }
     }
 
+    // Not using `getPayPalAPIDomain` function call here because it outputs 'cors.api.paypal.com'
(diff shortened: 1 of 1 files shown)


src/utils/sdk.js
```diff
@@ -239,7 +238,10 @@ export function getPayPalAPIDomain() {
     }
 
     if (__MESSAGES__.__TARGET__ === 'SDK') {
-        return getSDKPayPalAPIDomain();
+        if (getEnv() === 'sandbox') {
```
gillycheesesteak, line 243: Can we add a comment here about why we don't use the getPayPalAPIDomain function?

JordanClemons: Added

gillycheesesteak (approved):

JustinDoan:
LGTM

PR #1097, opened 2024-05-21
refactor: Update Calculator to show error when below/above threshold is met

perco12:
## Description

Update the calculator to not show the amounts if below/above error threshold is met.


yeul (approved):

src/components/modal/v2/parts/Calculator.jsx
```diff
@@ -218,7 +218,7 @@ const Calculator = ({ setExpandedState, calculator, aprDisclaimer, useV4Design,
                     />
                 </div>
             ) : null}
-            {country === 'US' && (
+            {country === 'US' && !error && (
```
yeul:
```suggestion
            {country === 'US' && (
```
I just realized that the check you added here removed the disclaimer under the error message and we always want that to show up regardless so we'll want to revert the change.

PR #1115, opened 2024-08-14
feat: playwright accessibility test for messages and modals

perco12:


perco12:
closing as didnt pull in latest so showing more commits then needed

PR #1131, opened 2024-09-05
refactor: update messageType values

yeul:


src/server/locale/US/CONTEXTUAL_COMPONENTS/BUTTONS/mutations/index.js
```diff
@@ -18,35 +18,43 @@ export default function getMutations(id, type) {
...
         // Short Term Non-qualifying
         case 'ST_NQ':
+        case 'PLST_NQ':
             return shortTermSingleNQ[type];
         // Generic Messages
         // Generic Pay Later message
         case 'GENERIC_PL':
             return genericPayLater[type];
         // // Generic Pay Later - Buy now, Pay Later message
-        case 'GENERIC_PL_BNPL':
```
Seavenly, line 49: Is this removal intentional? It seems like in every other case we're just adding, so wanted to double check.

src/server/locale/US/CONTEXTUAL_COMPONENTS/MARKS/mutations/index.js
```diff
@@ -18,35 +18,43 @@ export default function getMutations(id, type) {
...
         // Short Term Non-qualifying
         case 'ST_NQ':
+        case 'PLST_NQ':
             return shortTermSingleNQ[type];
         // Generic Messages
         // Generic Pay Later message
         case 'GENERIC_PL':
             return genericPayLater[type];
         // // Generic Pay Later - Buy now, Pay Later message
-        case 'GENERIC_PL_BNPL':
```
Seavenly, line 49: Same comment as above.

Seavenly (approved):

src/server/locale/US/CONTEXTUAL_COMPONENTS/BUTTONS/mutations/index.js
yeul: Good catch, fixed that up for buttons and marks. 👍

Seavenly (approved):

PR #1149, opened 2024-10-16
feat: Flex-Test Accessibility Automation & Run Tests Only In chrome (DTCRCMERC-3553)

erikacolette29:


tests/playwright/tests/sdk/messages/multi.spec.js
```diff
@@ -46,3 +46,11 @@ messageTest.describe('Multi Messages', () => {
         await messageAxeCoreScan(messageIframe);
     });
 });
+messageTest.describe('Flex Test', () => {
+    messageTest('Flex', async ({ navigatePage, page, loadMessage, messageAxeCoreScan }) => {
+        await navigatePage({ account: 'DEV0USGENERIC', url: '/accessibility/flexsdk.html' });
+        const messageIframe = await loadMessage();
+        await page.waitForTimeout(5000);
```
perco12, line 53:
```suggestion
        await page.waitForTimeout(5000);
        const messageIframe = await loadMessage();
```
can we switch these two lines as the `flexsdk.html` doesn't load all the banners yet before it grabs the `loadMessage`

perco12:
```suggestion
        await page.waitForTimeout(8000);
        const messageIframe = await loadMessage();
```
can we switch these lines and increase the timeout as not all banners are loaded in the 5 seconds @erikacolette29 ?

perco12 (approved):

PR #1165, opened 2024-12-13
chore: update snapshots

jadutter:


jadutter (commented):
~LGTM~

tajha17 (approved):

jadutter:
These snapshots appear to have been updated with ubuntu 22. 
We will want to make sure they are using ubuntu 24 when they update. 
See [this comment](https://github.com/paypal/paypal-messaging-components/pull/1166#issuecomment-2557742264) for further details.

We may need to either 
* wait for GitHub to finish their rollout by Jan 17th
* temporarily set our workflows to explicitly use ubuntu 24 instead of latest

PR #1161, opened 2024-12-04
feat: adds prop update listening to modal browser zoid polyfill

danzhaaspaypal:
## Description
Updates the modal lander to support communicating bi-directionally with the v6 modal wrapper

## Testing instructions

1. Prepare local env to test corresponding PR `feat: add learn more modal wrapper for messages` in an internal repo (DM Dan for more details).
2. in that internal repo, go to `LearnMore.ts`, and on line 52, append `&stage_tag=merc3611_test` to end of url query.  
3. Try combinations of the below lines inserted onto the demo page (again DM Dan for more details):

```
try {
    const clientToken = await getClientToken();
    const sdkInstance = await window.paypal.createInstance({
      clientToken,
      components: ["paypal-payments", "paypal-messages"],
    });
    const msgInstance = sdkInstance.createPayPalMessages({buyerCountry: "us"});
    const learnMore = await msgInstance.createLearnMore({amount:"50", presentationType:"modal"});
    // const learnMore = await msgInstance.createLearnMore({amount:"50", presentationType:"popup"});
    // const learnMore = await msgInstance.createLearnMore({amount:"50", presentationType:"redirect"});
    // const learnMore = await msgInstance.createLearnMore({amount:"50", presentationType:"auto"});
    // await learnMore.update({amount:"70", offerTypes:["PAY_LATER_LONG_TERM"]});
    // await learnMore.update({amount:"70", offerTypes:["PAY_LATER_SHORT_TERM"]});
    // await learnMore.open({amount:"70"});
    // await learnMore.update({amount:"90"});
    // await learnMore.open({amount:"90"});
```

PR #1170, opened 2025-01-28
feat: sends modal event data to modal wrapper for use in hooks

danzhaaspaypal:
## Description
Sends messages to browser modal wrapper on modal events onShow, onClose, and onCalculate.

## Testing instructions
See corresponding PR 637 in core-web-sdk for testing instructions.


src/components/modal/v2/lib/postMessage.js
```diff
@@ -0,0 +1,54 @@
...
+
+        if (type === 'ack') {
+            this.type = POSTMESSENGER_EVENT_TYPES.ACK;
+            this.eventPayload = POSTMESSENGER_ACK_PAYLOAD;
+        } else if (type === 'message') {
+            this.type = POSTMESSENGER_EVENT_TYPES.MESSAGE;
+            this.eventPayload = eventPayload;
+        }
+    }
+}
```
Seavenly: Does this need to be a class or could it just be a `createMessage(type, eventName, eventPayload)` that returns a simple object? Using a class here without any methods that just holds data seems like may be reaching for a more complex data structure.

If we include `sendEvent` as a method on this class then it makes more sense. Maybe this could look like
``` js
const event = new PostMessengerEvent(/* ... */);
event.sendTo(clientOrigin);
```

src/components/modal/v2/lib/postMessage.js
```diff
@@ -0,0 +1,54 @@
...
+
+    const validChars = '0123456789abcdefghijklmnopqrstuvwxyz';
+    const stringLength = 32;
+    let randomId = '';
+    for (let index = 0; index < stringLength; index++) {
+        const randomIndex = Math.floor(Math.random() * validChars.length);
+        randomId += validChars.charAt(randomIndex);
+    }
+    return randomId;
+}
```
Seavenly: Left a similar comment on a previous PR. Is there an existing util in `belter` that serves the same purpose that we could leverage?

src/components/modal/v2/lib/zoid-polyfill.js
```diff
@@ -1,10 +1,17 @@
...
+import { sendEvent, PostMessengerMessage } from './postMessage';
 
 const IOS_INTERFACE_NAME = 'paypalMessageModalCallbackHandler';
 const ANDROID_INTERFACE_NAME = 'paypalMessageModalCallbackHandler';
+// these constants should maintain parity with MESSAGE_MODAL_EVENT_NAMES in core-web-sdk
+export const POSTMESSENGER_EVENT_NAMES = {
+    CALCULATE: 'paypal-messages-modal-calculate',
+    CLOSE: 'paypal-messages-modal-close',
+    SHOW: 'paypal-messages-modal-show'
+};
```
Seavenly: Should this be moved into the `postMessage` file to keep it all together?

src/components/modal/v2/lib/postMessage.js
```diff
@@ -0,0 +1,54 @@
...
+
+export function sendEvent(payload, trustedOrigin) {
```
Seavenly: The `trustedOrigin` originates from `props.origin` which won't exist for all integration (e.g. lander pages). Should we have some quick short-circuit logic that early returns if the origin is undefined or empty?

src/components/modal/v2/lib/zoid-polyfill.js
```diff
@@ -13,28 +20,28 @@ function listenAndAssignProps(newProps, propListeners) {
...
         origin: eventOrigin,
         data: { eventName, id, eventPayload: newProps }
     } = updatedPropsEvent;
-    const clientOrigin = decodeURIComponent(initialProps.origin);
 
     if (eventOrigin === clientOrigin && eventName === 'PROPS_UPDATE' && newProps && typeof newProps === 'object') {
-        // send event ack so PostMessenger will stop reposting event
-        sendEventAck(id, clientOrigin);
+        // send event ack with original event id so PostMessenger will stop reposting event
+        sendEvent(new PostMessengerMessage('ack', id));
```
Seavenly: Is this missing the `clientOrigin` as the second argument?

src/components/modal/v2/lib/postMessage.js
```diff
@@ -0,0 +1,54 @@
...
+    constructor(type, eventName, eventPayload) {
+        this.eventName = eventName;
+        this.id = createUUID();
+
+        if (type === 'ack') {
+            this.type = POSTMESSENGER_EVENT_TYPES.ACK;
+            this.eventPayload = POSTMESSENGER_ACK_PAYLOAD;
+        } else if (type === 'message') {
+            this.type = POSTMESSENGER_EVENT_TYPES.MESSAGE;
+            this.eventPayload = eventPayload;
```
Seavenly: I'd like to be extra cautious here. This is the data that we are exposing publicly for anyone to access. Right now nothing sensitive is being exposed, but in the future that could accidentally change. Can we add an extra layer of validation here by only allowing explicitly approved fields to be included? Something like `createSafePayload(eventPayload)` that returns an object that only includes allow listed fields managed by the `createSafePayload` function. We should also add an explicit comment that explains the context and reason for this as it may seem redundant.

src/components/modal/v2/lib/zoid-polyfill.js
```diff
@@ -115,10 +129,10 @@ const setupBrowser = props => {
             });
         },
         onClose: ({ linkName }) => {
-            if (isIframe && document.referrer) {
-                const targetOrigin = new window.URL(document.referrer).origin;
```
Seavenly, line 132: Removing this will break some existing integrations as this value is used by integrations not passing in the `props.origin` manually today. Does this by chance work in place of the manual `origin` query param we introduced for our needs to simplify the new integration?`

src/components/modal/v2/lib/postMessage.js
danzhaaspaypal: I appreciate the suggestion building off my idea to make it an object. Considered adding a method to that class, then realized we would get a DataCloneError due to postMessage trying to clone the object and it having a function. So I think your suggestion we toss the class in favor of a function returning a simple object is the way to go.

src/components/modal/v2/lib/postMessage.js
danzhaaspaypal: See #1161  for response

src/components/modal/v2/lib/zoid-polyfill.js
danzhaaspaypal: Oop, will fix

src/components/modal/v2/lib/zoid-polyfill.js
danzhaaspaypal: Oh good catch! I can make the document.referrer's origin a fallback in the case the origin isn't passed manually so both cases are covered

src/components/modal/v2/lib/zoid-polyfill.js
```diff
@@ -110,6 +112,13 @@ const setupBrowser = props => {
             });
         },
         onCalculate: ({ value }) => {
+            const eventPayload = {
+                // for data security, also add new params to createSafePayload in ./postMessage.js
+            };
```
Seavenly: Should we just make the `eventPayload` optional so that it can be omitted instead of creating an empty object to pass in?

src/components/modal/v2/lib/zoid-polyfill.js
```diff
@@ -43,11 +44,12 @@ const getAccount = (merchantId, clientId, payerId) => {
 
 const setupBrowser = props => {
     const propListeners = new Set();
+    const clientOrigin = decodeURIComponent(props.origin);
```
Seavenly: Are we able to test if `new window.URL(document.referrer).origin` would be sufficient for our use-cases? Or is there a scenario we know where the query param is required? Maybe when inside a webview? If not though it would be nice to simplify around that since that's what the one post message hook we had prior was using.

src/components/modal/v2/lib/postMessage.js
```diff
@@ -0,0 +1,71 @@
...
+        'linkName' // close event
+    ];
+
+    const safePayload = {};
+    const entries = Object.entries(unscreenedPayload);
+    entries.forEach(entry => {
+        const [key, value] = entry;
+        if (allowedFields.includes(key)) {
+            safePayload[key] = value;
+        }
```
Seavenly: In order to help us debug this in the future when we inevitably forget this filter exists, should we add an `else` that includes a simple console warn or something to help point this out?

src/components/modal/v2/lib/zoid-polyfill.js
danzhaaspaypal: Documenting: as discussed in Slack, this will become tech debt

src/components/modal/v2/lib/zoid-polyfill.js
danzhaaspaypal: I had thought this would be more convenient for future extensibility and harmless in the meantime; is there a concern / is it a code smell?

src/components/modal/v2/lib/postMessage.js
danzhaaspaypal: Is this about what you have in mind?
`attn PayPal dev: modal hook payload params are screened by function createSafePayload in postMessage.js for data security. Please consider if param ${key} is secure for posting, and add the secure param(s) to that function's allowed fields.`

Seavenly: Lets shorten this up a bit since it's not a scenario that should happen in production and technically adds weight to the file size. We mainly just need something that we can quickly see in the console and then find the reference where it is occurring and then the code comments can expand on the why

Seavenly (approved):

PR #1175, opened 2025-02-19
feat: add sdk log auth header

surekhaw:
## Description
add auth header to sdk log events

## Screenshots
N/A

## Testing instructions
TBD

### DIFF
--- a/src/utils/logger.js
+++ b/src/utils/logger.js
@@ -8,7 +8,7 @@ import { ZalgoPromise } from '@krakenjs/zalgo-promise/src';
 import { getGlobalUrl } from './global';
 import { request } from './miscellaneous';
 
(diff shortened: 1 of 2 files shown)


src/utils/logger.js
```diff
@@ -212,3 +212,14 @@ logger.addTrackingBuilder(() => {
         timestamp: new Date().getTime()
     };
 });
+
+logger.addHeaderBuilder(() => {
+    if (__MESSAGES__.__TARGET__ === 'SDK') {
+        const encodedClientId = btoa(getClientId());
+        return {
+            // Send authorization header with tracking event
+            authorization: `Basic ${encodedClientId}`
```
paypal-rosman: capital A? https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Authorization

PR #1190, opened 2025-04-01
feat: supports v6 custom close button

danzhaaspaypal:
## Description

Logs modal close when triggered by the custom close button.
Also removes outdated focus code that interfered with giving custom close button focus.


src/components/modal/v2/lib/zoid-polyfill.js
```diff
@@ -15,20 +16,43 @@ function updateProps(newProps, propListeners) {
...
+        origin: eventOrigin,
+        data: { eventName, id }
+    } = event;
+    if (eventOrigin !== clientOrigin) {
+        return;
+    }
+    if (eventName === 'PROPS_UPDATE') {
+        handlePropsUpdateEvent(propListeners, event);
+    }
+    if (eventName === WRAPPER_CLOSE_MESSAGE_NAME) {
```
Seavenly: We should probably handle both of these event names the same to keep consistency.

Seavenly (approved):

src/components/modal/v2/lib/zoid-polyfill.js
danzhaaspaypal: When you say handle them, do you mean how they are referenced? Either both as constants declared above or both as strings? If so do you have a preference?

Seavenly: Yea exactly. I should have been a little more clear there. Since they are only referenced here I think just having the string values prevents an extra mental hop to the actual value.

danzhaaspaypal: Counterpoint, I kind of reference them across repos, by adding this at the top of the file here in the messaging repo...
```
// these constants should maintain parity with MESSAGE_MODAL_EVENT_NAMES in core-web-sdk
export const WRAPPER_CLOSE_MESSAGE_NAME = 'MODAL_CLOSED';
```

... and this at the top of LearnMore.ts in the core-web-sdk repo:
```
// these constants should maintain parity with POSTMESSENGER_EVENT_NAMES in paypal-messaging-components
// https://github.com/paypal/paypal-messaging-components/blob/3283443d3a970471aaf914761a0371a894c8d90a/src/components/modal/v2/lib/zoid-polyfill.js#L10
const WRAPPER_CLOSE_MESSAGE_NAME = 'MODAL_CLOSED'
```

danzhaaspaypal: But I guess overall changing the constants to strings simplifies so let's go with that

Seavenly (approved):

PR #1194, opened 2025-04-25
refactor: update XO modal designs to 5.2

yeul:


src/components/modal/v2/parts/Calculator.jsx
```diff
@@ -245,6 +252,7 @@ const Calculator = ({
                 </div>
             )}
         </div>
+        // </div>
```
surekhaw: don't think we need this

surekhaw (approved):

PR #1196, opened 2025-04-29
refactor: adds v6 modal tracking key integration_identifier

danzhaaspaypal:
Adds `integration_identifier` for tracking when a modal is integrated via the standalone vs message-linked integration.

`integration_identifier`: describes if the modal was integrated as standalone or rather with a message.  Sample value: `messagesModal`

## Description
Allows tracking of `integration_identifier`.
Takes `integration_identifier` query params that originated in v6 wrapper and passes it to glog as `INTEGRATION_ID`.

## Screenshots
Standalone modal sample output (has `INTEGRATION_ID`)
Screenshot 2025-05-01 at 10 14 45 AM

## Testing instructions
When looking at v6 demo page, check the glog output of modal events `onShow`, `onCalculate`, and `onClose`.  For the standalone modal only, component key `INTEGRATION_ID` should have value `messagesModal`; the key will be absent for a normal message's modal.

### DIFF
--- a/src/components/modal/v2/lib/hooks/calculator.js
+++ b/src/components/modal/v2/lib/hooks/calculator.js
@@ -70,7 +70,8 @@ export default function useCalculator({ autoSubmit = false } = {}) {
         devTouchpoint,
         disableSetCookie,
         features,
-        language
+        language,
+        integrationIdentifier
     } = useXProps();
 
     const [state, dispatch] = useReducer(reducer, {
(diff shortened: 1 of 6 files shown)


merlinpaypal (approved):
Not blockers but src/components/modal/v2/lib/postMessage.js and src/components/modal/v2/parts/InlineLinks.jsx have lint warnings. Since those seem intentional, perhaps we could add `// eslint-disable-next-line no-console` and `// eslint-disable-next-line react/no-danger` to those lines so we don't get the warnings.

The actual PR content looks good to me

PR #1205, opened 2025-06-04
fix: add validation

tajha17:

### DIFF
--- a/src/utils/sdk.js
+++ b/src/utils/sdk.js
@@ -275,9 +275,9 @@ export function getFeatures(featureProps) {
 export function getURIPopup(lander, label) {
     try {
         // eslint-disable-next-line compat/compat
-        const parsed = new URL(lander, window.location.origin);
-        const isHttp = parsed.protocol === 'https:' || parsed.protocol === 'http:';
-        const isPayPalDomain = /\.paypal\.com$/i.test(parsed.hostname) || /\.paypal\.com:\d+$/i.test(parsed.host);
+        const parsed = new URL(lander);
+        const isHttp = parsed.protocol === 'https:';
+        const isPayPalDomain = /\.paypal\.com$/i.test(parsed.hostname);
 
         if (isHttp && isPayPalDomain) {
             return window.open(lander, label, 'width=460,height=900');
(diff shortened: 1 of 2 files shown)


src/utils/sdk.js
```diff
@@ -275,5 +273,18 @@ export function getFeatures(featureProps) {
 
 // open mini-browser with message lander url
 export function getURIPopup(lander, label) {
-    return window.open(lander, label, 'width=460,height=900');
+    try {
+        // eslint-disable-next-line compat/compat
+        const parsed = new URL(lander, window.location.origin);
+        const isHttp = parsed.protocol === 'https:' || parsed.protocol === 'http:';
```
Seavenly: Let's limit the protocol to `https`

src/utils/sdk.js
```diff
@@ -275,5 +273,18 @@ export function getFeatures(featureProps) {
...
+        const isPayPalDomain = /\.paypal\.com$/i.test(parsed.hostname) || /\.paypal\.com:\d+$/i.test(parsed.host);
```
Seavenly: Does `/\.paypal\.com:\d+$/i.test(parsed.host)` provide anything extra that the `hostname` check isn't doing since the pattern is for any port?

src/utils/sdk.js
```diff
@@ -275,5 +273,18 @@ export function getFeatures(featureProps) {
 
 // open mini-browser with message lander url
 export function getURIPopup(lander, label) {
-    return window.open(lander, label, 'width=460,height=900');
+    try {
+        // eslint-disable-next-line compat/compat
+        const parsed = new URL(lander, window.location.origin);
```
Seavenly: We may want to consider removing the second argument here. The `lander` should always be a full URL and the `window.location.origin` will normally be a merchant site.

src/utils/sdk.js
tajha17: yeah the port is redundant 0d00654def5ade426a8d6c4639f60d2d6dfdfcd0

src/utils/sdk.js
tajha17: 0d00654def5ade426a8d6c4639f60d2d6dfdfcd0

Seavenly (approved):

PR #1217, opened 2025-08-18
feat: Implements Global Session ID

jonacox-paypal:
## Description
- Adds a global_session_id for correlating events across different use cases
- getOrCreateSessionID uses belter for generating uid
- Not applicable in standalone use case

## Screenshots
Passes global_session_id to logger for tracking
image

## Testing instructions
Validated sdk via cdn stage bundle as well as downstream internal applications to accept the event for tracking purposes

### DIFF
--- a/src/utils/sdk.js
+++ b/src/utils/sdk.js
@@ -168,16 +168,13 @@ export function getStorage() {
 // and used to correlate with button events
 export function getOrCreateGlobalSessionID() {
     if (__MESSAGES__.__TARGET__ === 'SDK') {
-        const globalSessionID = getSDKGlobalSessionID();
-
-        if (globalSessionID) {
-            return globalSessionID;
-        } else {
-            const globalSessionIDValue = createBelterID();
-            setSDKGlobalSessionID(globalSessionIDValue);
-
-            return globalSessionIDValue;
+        let globalSessionID = getSDKGlobalSessionID();
+        if (!globalSessionID) {
+            globalSessionID = createBelterID();
+            setSDKGlobalSessionID(globalSessionID);
         }
+
+        return globalSessionID;
     } else {
         return undefined;
     }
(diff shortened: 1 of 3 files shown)


src/utils/sdk.js
```diff
@@ -161,6 +163,26 @@ export function getStorage() {
...
+            return globalSessionID;
+        } else {
+            const globalSessionIDValue = createBelterID();
+            setSDKGlobalSessionID(globalSessionIDValue);
+
+            return globalSessionIDValue;
+        }
+    } else {
+        return undefined;
+    }
```
paypal-rosman: we can update 171 to use a `let` so we can re-assign the variable to prevent creating another variable. something like this:

```
let globalSessionID = getSDKGlobalSessionID();
    if (!globalSessionID) {
        globalSessionID = createBelterID();
        setSDKGlobalSessionID(globalSessionID);
    }

    return globalSessionID;
```

paypal-rosman (approved):

src/utils/sdk.js
jonacox-paypal: Thanks Rene, Ive made the update in latest commit.

src/utils/sdk.js
```diff
@@ -161,6 +163,23 @@ export function getStorage() {
...
+    if (__MESSAGES__.__TARGET__ === 'SDK') {
+        let globalSessionID = getSDKGlobalSessionID();
+        if (!globalSessionID) {
+            globalSessionID = createBelterID();
+            setSDKGlobalSessionID(globalSessionID);
+        }
+
+        return globalSessionID;
+    } else {
+        return undefined;
```
Seavenly: I mentioned we could just leave this as `undefined`. Let's actually just use the same fallback as `getSessionID`, so `getStorage().getSessionID()`. They can be the same value that's fine.

Seavenly (approved):

src/utils/sdk.js
jonacox-paypal: Thanks Nate, Ive made the update.

PR #1236, opened 2025-10-02
refactor: use new SDK method for global identifier

jonacox-paypal:
## Description
- Uses new SDK getPayPalSession method
- Uses this method in place of getOrCreateGlobalSessionID method

## Screenshots
image

### DIFF
--- a/src/components/message/Message.js
+++ b/src/components/message/Message.js
@@ -227,7 +227,9 @@ const Message = function ({ markup, meta, parentStyles, warnings }) {
                                 // getRequestDuration runs in the child component (iframe/banner message),
                                 // passing a value to onReady and up to the parent component to go out with
                                 // the other stats
-                                requestDuration: getRequestDuration()
+                                requestDuration: getRequestDuration(),
+                                // Global Session ID allows messages to be correlated to button events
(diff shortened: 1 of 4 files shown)


jeremy-herman (approved):

danzhaaspaypal (approved):

src/components/message/Message.js
```diff
@@ -81,7 +82,9 @@ const Message = function ({ markup, meta, parentStyles, warnings }) {
         // getRequestDuration runs in the child component (iframe/banner message),
         // passing a value to onReady and up to the parent component to go out with
         // the other stats
-        requestDuration: getRequestDuration()
+        requestDuration: getRequestDuration(),
+        // Global Session ID allows messages to be correlated to button events
+        globalSessionID: getGlobalSessionID()
```
Seavenly, line 86: There's another spot below where `onReady` can be called for the rerender use case where we should include this to stay consistent and prevent accidentally removing the value if it were to be omitted.

jonacox-paypal: Thanks Nate, I've addressed the re-render logic

Seavenly (approved):

PR #1249, opened 2025-12-10
feat: Add language_requested field to fpti

anushreeshah2405:
## Description

**Ticket** - Extend instrumentation to include `language_requested`

- Add `language_requested` field to FPTI logging to track the requested language parameter in analytics

**Changes**

- Modified `component.js` (message and modal) to extract the `language` and `locale` props in the onReady handler
- Added `language_requested` field to the component tracking data passed to logger
- Updated `logger.js` to include `language_requested` in the log payload with a fallback value of **'undefined'** when not present


danzhaaspaypal (approved):

danzhaaspaypal (changes requested):
requesting `language_requested` be added to the modal event as well please

danzhaaspaypal (approved):

surekhaw (approved):

PR #1260, opened 2026-01-21
refactor: Feature/payloadfields

saikiran-m7:
## Description
Update all references to payload fields to match the name that gets sent in the request.
Locate and identify usage of old fpti key names in the logger for paypal-messaging-components.

### DIFF
--- a/src/utils/logger.js
+++ b/src/utils/logger.js
@@ -11,7 +11,7 @@ import { request } from './miscellaneous';
 import { getLibraryVersion, getDisableSetCookie, getClientId } from './sdk';
 
 function generateLogPayload(account, { meta, events: bizEvents, tracking }) {
-    const { deviceID, sessionID, integration_type, messaging_version, globalSessionID } = meta.global ?? {};
+    const { deviceID, sessionID, integration_type, integration_version, globalSessionID } = meta.global ?? {};
 
     let clientID;
     if (account.startsWith('client-id:')) {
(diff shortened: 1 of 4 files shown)


src/utils/logger.js
```diff
@@ -39,13 +39,14 @@ function generateLogPayload(account, { meta, events: bizEvents, tracking }) {
             merchant_profile_valid = merchant_profile_valid ?? trackingDetails.MERCHANT_PROFILE_VALID;
             buyer_profile_hash = buyer_profile_hash ?? trackingDetails.BUYER_PROFILE_HASH;
             buyer_profile_valid = buyer_profile_valid ?? trackingDetails.BUYER_PROFILE_VALID;
-            partner_attribution_id = partner_attribution_id ?? stats.bn_code;
+            partner_attribution_id = partner_attribution_id ?? stats.partner_attribution_id;
```
JordanClemons, line 42: Does `partner_attribution_id` already exist on the `stats` object?

saikiran-m7: Yes, it's there in Stats. However, I see the changes need to be done only in 2 files. If anything I missed please do let me know.
image

perco12: correct me if I am wrong, but you will also need to update `bn_code` in stats.js to be `partner_attribuition_id` or its going to be undefined when it looks for stats.partner_attribuition_id. 
[Location in stats.js ](https://github.com/paypal/paypal-messaging-components/blob/9c275587e0fc9713c00dea7b325cccc84e22f3fd/src/utils/stats.js#L64)

perco12: to add on to that, do we also need to update all other locations that are using `bn_code` like [here](https://github.com/paypal/paypal-messaging-components/blob/9c275587e0fc9713c00dea7b325cccc84e22f3fd/src/library/zoid/modal/component.js#L349) or just in logger @JordanClemons ?

JordanClemons: We should yeah

saikiran-m7: @JordanClemons @perco12 Resolved the comments. Please review and let me know the changes.

perco12 (approved):

perco12 (approved):

PR #1268, opened 2026-02-02
feat: add FAQ links to warning messages for merchant troubleshooting

papallavi:
## Description

Add help_url field to all warning messages directing merchants to troubleshooting documentation. Enhances console output to display FAQ links for common issues like message hidden, invalid options, and integration errors.
https://paypal.atlassian.net/browse/DTCRCMERC-3133

### DIFF
--- a/globals.js
+++ b/globals.js
@@ -43,7 +43,7 @@ module.exports = (env = { TARGET: 'sdk' }) => ({
             __CREDIT_APPLY__: '/credit-application/paypal-credit-card/da/us/billing'
         },
         __FAQ__: {
-            __BASE_URL__: 'https://developer.paypal.com/docs/business/pay-later'
+            __BASE_URL__: 'https://developer.paypal.com/docs/checkout/pay-later/us'
         }
     }
 });
--- a/globals.js
+++ b/globals.js
@@ -41,6 +41,9 @@ module.exports = (env = { TARGET: 'sdk' }) => ({
             __MODAL__: '/credit-presentment/smart/modal',
             __LOGGER__: '/credit-presentment/glog',
             __CREDIT_APPLY__: '/credit-application/paypal-credit-card/da/us/billing'
+        },
+        __FAQ__: {
+            __BASE_URL__: env.FAQ_BASE_URL || 'https://developer.paypal.com/docs/business/pay-later'
         }
     }
 });
--- a/globals.js
+++ b/globals.js
@@ -43,7 +43,7 @@ module.exports = (env = { TARGET: 'sdk' }) => ({
             __CREDIT_APPLY__: '/credit-application/paypal-credit-card/da/us/billing'
         },
         __FAQ__: {
-            __BASE_URL__: env.FAQ_BASE_URL || 'https://developer.paypal.com/docs/business/pay-later'
+            __BASE_URL__: 'https://developer.paypal.com/docs/business/pay-later'
         }
     }
 });
--- a/src/library/zoid/message/component.js
+++ b/src/library/zoid/message/component.js
@@ -154,7 +154,7 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
                             if (getURIPopup(lander, offerType) == null) {
                                 logger.warn('Blocked unsafe lander URL', {
                                     lander,
-                                    help_url: getFaqUrl('INVALID_OPTIONS')
+                                    help_url: getFaqUrl('GENERAL')
                                 });
                             }
                         } else {
@@ -311,7 +311,7 @@ export default createGlobalVariableGetter('__paypal_credit_message__', () =>
                                 logger.warn('render_warning', {
                                     description: warning,
                                     container: getContainer(),
-                                    help_url: getFaqUrl('RENDER_WARNING')
+                                    help_url: getFaqUrl('RENDERING')
                                 });
                             });
                         }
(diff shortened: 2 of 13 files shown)


src/utils/logger.js
```diff
@@ -209,6 +209,11 @@ logger.addMetaBuilder(() => {
 });
 
 logger.addPayloadBuilder(payload => {
+    // Append help URL to description for better console visibility
+    if (payload.help_url && payload.description) {
+        payload.description = `${payload.description}\n\nFor help, see: ${payload.help_url}`; // eslint-disable-line no-param-reassign
```
Braluna-pp: Avoid mutating `description` for telemetry payloads,

**Reason:** payload builders run before transport; this change alters data sent to glog and could break downstream parsers/dashboards that expect a stable description. 

Consider leaving description unchanged and rely on help_url, or gate the append to console‑only logging.

src/utils/faq.js
```diff
@@ -0,0 +1,25 @@
...
+    GENERAL: '/troubleshooting/'
+};
+
+/**
+ * Generate FAQ URL for a given topic
+ * @param {string} topic - The FAQ topic identifier
+ * @returns {string} Full URL to the FAQ section
+ */
+export function getFaqUrl(topic) {
+    const basePath = __MESSAGES__?.__FAQ__?.__BASE_URL__ ?? 'https://developer.paypal.com/docs/business/pay-later';
```
Braluna-pp: Normalize FAQ base URL to avoid `//` joins,

**Reason:** `FAQ_BASE_URL` is env‑provided and may include a trailing slash; concatenating with /troubleshooting/... can yield //, which can 404 in some doc setups. Normalize with `basePath.replace(/\/$/, '')` before concatenation.

src/library/zoid/message/component.js
Braluna-pp: Missing FAQ link for unsafe lander warning (common error?)

**Reason:** this warning is surfaced to merchants but doesn’t include help_url, so the new FAQ linking is inconsistent with the “common errors” AC. Either add a FAQ topic for this warning or explicitly scope it out in the PR description.

Braluna-pp:
In `/messages/interface.test.js` consider doing the following,

Creating tests that seert `help_url` for warnings.

**Reason:** current tests only check selector/location and won’t catch regressions to the new help_url behavior. Add assertions for help_url (and optionally appended description) in warning tests.

Braluna-pp (approved):
Looks good to me

src/utils/faq.js
```diff
@@ -0,0 +1,29 @@
...
+
+// Topic-to-path mapping for FAQ sections
+const FAQ_PATHS = {
+    MESSAGE_HIDDEN: '/troubleshooting/#message-hidden',
+    INVALID_OPTIONS: '/troubleshooting/#invalid-options',
+    INVALID_SELECTOR: '/troubleshooting/#integration',
+    NOT_IN_DOCUMENT: '/troubleshooting/#integration',
+    RENDER_WARNING: '/troubleshooting/#rendering',
+    UNSAFE_LANDER: '/troubleshooting/#invalid-options',
+    GENERAL: '/troubleshooting/'
```
yeul: Some of these are duplicates so to me it seems like it is unnecessary to create unique enum values for the same link.

globals.js
```diff
@@ -41,6 +41,9 @@ module.exports = (env = { TARGET: 'sdk' }) => ({
             __MODAL__: '/credit-presentment/smart/modal',
             __LOGGER__: '/credit-presentment/glog',
             __CREDIT_APPLY__: '/credit-application/paypal-credit-card/da/us/billing'
+        },
+        __FAQ__: {
+            __BASE_URL__: env.FAQ_BASE_URL || 'https://developer.paypal.com/docs/business/pay-later'
```
yeul:
```suggestion
            __BASE_URL__: 'https://developer.paypal.com/docs/business/pay-later'
```
I don't think the `||` check is necessary in this case

Braluna-pp (approved):

globals.js
```diff
@@ -41,6 +41,9 @@ module.exports = (env = { TARGET: 'sdk' }) => ({
             __MODAL__: '/credit-presentment/smart/modal',
             __LOGGER__: '/credit-presentment/glog',
             __CREDIT_APPLY__: '/credit-application/paypal-credit-card/da/us/billing'
+        },
+        __FAQ__: {
+            __BASE_URL__: 'https://developer.paypal.com/docs/business/pay-later'
```
yeul:
```suggestion
            __BASE_URL__: 'https://developer.paypal.com/docs/checkout/pay-later/us'
```
Let's update the URL to the above since it is the more up to date url to base off of. It is redirecting here anyway.

src/utils/faq.js
```diff
@@ -0,0 +1,27 @@
...
+ * Used in warning messages to direct merchants to troubleshooting documentation
+ */
+
+// Topic-to-path mapping for FAQ sections
+const FAQ_PATHS = {
+    MESSAGE_HIDDEN: '/troubleshooting/#message-hidden',
+    INVALID_OPTIONS: '/troubleshooting/#invalid-options',
+    INTEGRATION: '/troubleshooting/#integration',
+    RENDER_WARNING: '/troubleshooting/#rendering',
+    GENERAL: '/troubleshooting/'
```
yeul: So some of these FAQ paths do not actually exist yet, so I think we can condense this even more:

We should be able to simplify to only these two:
RENDERING - '/integrate/#enable-pay-later-messaging-on-your-website'
GENERAL - '/integrate/reference/'

So for any situations where you included `MESSAGE_HIDDEN`, `INTEGRATION`, `RENDER_WARNING`, let's just condense that into just `RENDERING`

And then for `INVALID_OPTIONS`, we can update it to the new `GENERAL` reference link instead.

yeul (approved):

PR #1272, opened 2026-02-09
feat: improve CLS with single-line text placeholder (v5)

Braluna-pp:
## Why
We require the v5 JS SDK text message placeholder to reserve exactly one line during initial blank render, then expand naturally after the message loads, and collapse when no message is returned.

## Root cause
The placeholder height logic in `containerTemplate` reserved two lines for text layout before render, which is larger than the acceptance criteria and increases layout shift risk on cache miss paths.

## What changed
- Updated text-layout placeholder reservation from 2 lines to 1 line in `containerTemplate`.
- Kept existing height strategy (validated text size range with fallback) and existing line-height multiplier.
- Preserved resize-driven expansion behavior for rendered messages.
- Preserved immediate placeholder clear on first resize signal, including `0x0` no-message path.

## Files changed
- `src/library/zoid/message/containerTemplate.jsx`
- `tests/unit/spec/src/zoid/message/containerTemplate.test.js`

## Lighthouse metrics
image

## Video

https://github.com/user-attachments/assets/84cbceab-54cc-428f-9a04-f0b5fa51e00a

## Acceptance criteria mapping
- [x] Reserve a single line of blank height on initial blank render.
- [x] Allow message to expand beyond one line once rendered.
- [x] Collapse reserved space when no message is returned (e.g., 422 path represented by `0x0` resize).

## Validation
- Pre-commit checks passed (lint/prettier + unit tests).
- Added/updated unit coverage for:
  - one-line placeholder reservation
  - fallback behavior for invalid/out-of-range text size
  - placeholder clear on first resize with non-zero dimensions
  - placeholder clear on first resize with `0x0` no-message signal

### DIFF
--- a/src/library/zoid/message/containerTemplate.jsx
+++ b/src/library/zoid/message/containerTemplate.jsx
@@ -9,7 +9,7 @@ const DEFAULT_TEXT_SIZE_PX = 14;
 const MIN_TEXT_SIZE_PX = 10;
 const MAX_TEXT_SIZE_PX = 16;
 const TEXT_LINE_HEIGHT = 1.3;
-const TEXT_LINE_COUNT = 2;
+const TEXT_LINE_COUNT = 1;
 
 const getPlaceholderHeightPx = style => {
     if (style?.layout !== 'text') {
--- a/src/library/zoid/message/containerTemplate.jsx
+++ b/src/library/zoid/message/containerTemplate.jsx
@@ -5,14 +5,49 @@ import { EVENT } from '@krakenjs/zoid/src';
 import { getOverflowObserver, createTitleGenerator } from '../../../utils';
 
 const getTitle = createTitleGenerator();
+const DEFAULT_TEXT_SIZE_PX = 14;
+const MIN_TEXT_SIZE_PX = 10;
+const MAX_TEXT_SIZE_PX = 16;
+const TEXT_LINE_HEIGHT = 1.3;
+const TEXT_LINE_COUNT = 2;
+
+const getPlaceholderHeightPx = style => {
+    if (style?.layout !== 'text') {
+        return null;
+    }
+
+    const rawSize = style?.text?.size;
+    const parsedSize = typeof rawSize === 'number' ? rawSize : Number(rawSize);
+    const size =
+        Number.isFinite(parsedSize) && parsedSize >= MIN_TEXT_SIZE_PX && parsedSize <= MAX_TEXT_SIZE_PX
+            ? parsedSize
+            : DEFAULT_TEXT_SIZE_PX;
+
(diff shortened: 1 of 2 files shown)


src/library/zoid/message/containerTemplate.jsx
```diff
@@ -5,14 +5,49 @@ import { EVENT } from '@krakenjs/zoid/src';
 import { getOverflowObserver, createTitleGenerator } from '../../../utils';
 
 const getTitle = createTitleGenerator();
+const DEFAULT_TEXT_SIZE_PX = 14;
+const MIN_TEXT_SIZE_PX = 10;
+const MAX_TEXT_SIZE_PX = 16;
+const TEXT_LINE_HEIGHT = 1.3;
+const TEXT_LINE_COUNT = 1;
```
perco12, line 12: just curious about these defaults. Where did we get these values, are these like the culmuative average of all our messages text height etc no matter layout?

Braluna-pp: Hello @perco12 ,

These values were not picked from an average.

They come from the current text-layout rules we already use:
- `10..16` matches the allowed `style.text.size` range (`parseTextSize` + locale validOptions)
- `14` matches the default base font size when `style.text.size` is not valid/provided
- `1.3` matches the text layout line-height

And this placeholder only runs for `layout === "text"` (not flex). We also clear the placeholder on the first resize event, so it is just a temporary reserve height before real dimensions arrive.

tests/unit/spec/src/zoid/message/containerTemplate.test.js
```diff
@@ -0,0 +1,87 @@
...
+
+    test('Clears placeholder min-height on first resize with rendered content dimensions', () => {
+        const { event, container } = renderTemplate({ layout: 'text', text: { size: 12 } });
+        expect(container.style.minHeight).toBe('15.6px');
+
+        event.trigger(EVENT.RESIZE, { width: 100, height: 0 });
+        expect(container.style.minHeight).toBe('');
+    });
+
+    test('Clears placeholder min-height on first resize with no-message dimensions (0x0)', () => {
```
perco12, line 75: This test, is it saying when we have no content return to us that it just clear or?

Braluna-pp: Hello @perco12 ,

That is correct,

This test is covering the case where the first resize comes back as `0x0`, so we did not get real message dimensions yet. In that case we still want to clear the temporary placeholder `min-height` we added during the initial text-layout render.

The placeholder is only there to reserve space before the first resize. Once the resize event fires, whether it gives us real dimensions or `0x0`, we should restore the original `min-height` so we do not leave fake empty space behind.

perco12 (approved):
Image

surekhaw (approved):

PR #1287, opened 2026-03-19
Pay in 4 qualifying messages to support Spending Power CTA

saikiran-m7:
## Description
We only want to show the “Check Spending Power” disclaimer on the Pay in 4 qualifying message if the customer falls into the ELMO experiment treatment, otherwise they should show the standard “Learn more” disclaimer. This will be our way to gate the Spending Power/Prequalification experience.

## Screenshots / Videos
https://localhost.paypal.com:8080/standalone.html

image

## Testing instructions
**npm run dev:standalone**
stage link: https://www.paypalobjects.com/upstream/bizcomponents/stage/payin4spendingpowercta00/js/messaging.js


src/server/locale/US/PAY_LATER_SHORT_TERM/mutations/short_term_q.js
```diff
@@ -69,7 +69,7 @@ export default {
             'default',
             ({ textSize }) => ({
                 styles: [
-                    textWrap(textSize * 32, textSize, 'US'),
+                    textWrap(textSize * 36, textSize, 'US'),
```
perco12, line 72: Why did we need to increase the text size? Is the Wrapping off now with the extra words?

saikiran-m7: Yeah. It was not wrapping up properly, so decided to change the size 32 to 36 and after the change it's looking good.

content/messages/US/short_term_q.json
yeul, line 8:
```suggestion
        "lander": "",
        "prequalExperience": "false",
```
Can we add `"prequalExperience": "false"` here so that others can use this to develop the modal experience locally?

saikiran-m7: Added prequalExperience.

yeul (commented):
I think actually what might be good for local development is to create a `prequalification` subfolder like `content/messages/US/prequalification/short_term_q.json` and just paste the updated content there. Then we can revert the original `short_term_q.json` back to normal (with the Learn more text) so people can still reference what the non-experiment version looks like. 

Then, under [this line](https://github.com/paypal/paypal-messaging-components/blob/7b5f58860edff115726779267ae5145c9b64b58a/utils/devServerProxy/config/US/index.js#L135) we can create a new object called `DEV_US_SHORT_TERM_CYSP`. You can copy and paste the existing `DEV_US_SHORT_TERM` object to do this. All you'll want to do in this new object is update the `short_term_q.json` in your object to be `/prequalification/short_term_q.json` then you should be good to go.

saikiran-m7:
@yeul Added the changes as per the review comment.

content/messages/US/prequalification/short_term_q.json
```diff
@@ -0,0 +1,32 @@
...
+        [["Buy now and pay later."], ["xsmall.2"]],
+        [
+            ["Pay in 4 interest-free payments of {formattedPeriodicPayment}", [".", "weak"]],
+            ["medium", "large", "xlarge"]
+        ]
+    ],
+    "subHeadline": [],
+    "disclaimer": [
+        [
+            "<span aria-hidden='true'>Check Spending Power</span><span class='sr-only'> Check Spending Power</span>",
```
yeul: Small nitpick! 
```suggestion
            "<span aria-hidden='true'>Check spending power</span><span class='sr-only'> Check spending power</span>",
```

yeul (commented):
One more small comment and we should be good to go!

yeul (approved):

perco12 (approved):

PR #1304, opened 2026-04-23
feat: add prequalification route to CYSP CTA button

perco12:
## Description

Add functionality to prequalify CTA by redirecting to new acquisition route. 

## Review

### DIFF
--- a/src/components/modal/v2/lib/utils.js
+++ b/src/components/modal/v2/lib/utils.js
@@ -1,5 +1,5 @@
 import { isIosWebview, isAndroidWebview } from '@krakenjs/belter/src';
-import { request, memoize, ppDebug } from '../../../../utils';
+import { request, memoize, ppDebug, getGlobalUrl } from '../../../../utils';
 import validate from '../../../../library/zoid/message/validation';
 
 export const getContent = memoize(
@@ -133,4 +133,29 @@ export function validateProps(updatedProps) {
         }
     });
     return validatedProps;
+}
+
+/**
+ * Redirect the modal window to the CAP_s prequalification route.
+ * Accepts query params (e.g., token, offer).
+ *
+ * @param {Object} params - query parameters to append to the URL
+ */
+export function openPrequalification(params = {}) {
+    const { token, offer } = params;
+    const query = Object.entries({ offer })
+        .filter(([, val]) => typeof val !== 'undefined' && val !== null && val !== '')
+        // Encode values to keep URL valid; keys are expected to be safe identifiers.
+        .map(([key, val]) => `${key}=${encodeURIComponent(val)}`)
+        .join('&');
+    // TODO: sessionIdentifier is being discussed; currently using ecToken.
+    // Base URL comes from globals.js: __PREQUALIFICATION__.
+    const url = `${getGlobalUrl('PREQUALIFICATION')}?token=${encodeURIComponent(token ?? '')}${
+        query ? `&${query}` : ''
+    }`;
+
+    // Preflight the route to avoid navigating away from the modal if prequalify is unavailable.
+    return request('GET', url).then(() => {
+        window.location.assign(url);
+    });
 }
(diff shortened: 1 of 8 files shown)


src/components/modal/v2/lib/utils.js
```diff
@@ -134,3 +134,28 @@ export function validateProps(updatedProps) {
...
+        // Encode values to keep URL valid; keys are expected to be safe identifiers.
+        .map(([key, val]) => `${key}=${encodeURIComponent(val)}`)
+        .join('&');
+    // TODO: sessionIdentifier is being discussed; currently using ecToken.
+    // Base URL comes from globals.js: __PREQUALIFICATION__.
+    const url = `${getGlobalUrl('PREQUALIFICATION')}?token=${encodeURIComponent(token ?? '')}${
+        query ? `&${query}` : ''
+    }`;
+
+    // Preflight the route to avoid navigating away from the modal if prequalify is unavailable.
```
perco12: I put a preflight on it, but thoughts on should we instead let CAP_S handle the error if the prequalify route fail or?

src/components/modal/v2/parts/views/LongTerm/Content.jsx
```diff
@@ -82,14 +82,20 @@ export const LongTerm = ({
...
+    productMeta: { useV4Design, useV5Design, showPromoContent, prequalExperience, product },
     openProductList,
     useNewCheckoutDesign,
     use5Dot1Design
 }) => {
     const [expandedState, setExpandedState] = useState(false);
-    const { amount, onClick, onClose } = useXProps();
+    const { amount, onClick, onClose, ecToken } = useXProps();
     const { views, country } = useServerData();
+    const offer = product ?? 'PAY_LATER_LONG_TERM';
```
perco12: now I did from product, but thoughts on if we should be getting offer from Xprops? probably need to turn offerTypes into string etc, so product might be the best choice, but would love to hear thoughts

yeul: I honestly think we're actually okay to simplify and remove this line and instead just map it to offer in `usePrequalification`. We should always have a `product` in the modal content and if we don't we have other issues haha.

yeul: I would avoid putting it in xprops because that would mean we are relying on merchant inputs. We can't really rely on that since no merchant is going to update their integration unless absolutely necessary and we want to make sure we can control the experience fully. If a merchant passes in an offer type for a specific modal, we'll just return the corresponding content (assuming everything is eligible) and then from there we just pull `product` from the `meta`

src/components/modal/v2/lib/utils.js
yeul: Yeah I think if the CAP_s route is unavailable for some reason we should just let CAP_s handle it. We do similar for the PPC modal as well where we just navigate and let the other app handle things.

src/components/modal/v2/parts/views/LongTerm/Content.jsx
```diff
@@ -82,14 +82,20 @@ export const LongTerm = ({
...
     },
-    productMeta: { useV4Design, useV5Design, showPromoContent, prequalExperience },
+    productMeta: { useV4Design, useV5Design, showPromoContent, prequalExperience, product },
     openProductList,
     useNewCheckoutDesign,
     use5Dot1Design
 }) => {
     const [expandedState, setExpandedState] = useState(false);
-    const { amount, onClick, onClose } = useXProps();
+    const { amount, onClick, onClose, ecToken } = useXProps();
```
yeul, line 92: Just commenting to say I think pulling `ecToken` from xprops is okay for now because we don't yet know what the session token will be yet/where it will come from. The end goal though is to not having it come from an input like this and rather some other service. We'll make a follow up ticket to tweak things when that's decided though!

perco12: agreed. I put todos in the function to mention that

yeul (approved):

surekhaw (approved):

PR #1321, opened 2026-05-19
feat: UK Pay in 30 days Offer Type expansion for GB

lyerabati16:
## Review


src/components/modal/v2/parts/views/PayIn1/styles.scss
```diff
@@ -3,6 +3,19 @@
 @use '../../../styles/globals/variables';
 
 .content__body {
+    .learnMoreLink__container {
+        font-size: 14px;
+
```
smoyalpaypal, line 8: This change seems to be irrelevant.
it is applicable to all countries.

Have a relook

lyerabati16: The .learnMoreLink__container was originally added by Surekha in a previous merged ModXO PR to size/space the "Learn more" link in the Pay in 1 modal (14px text, tighter top margin on tablet, no left indent on mobile). It's expected to be country-agnostic since PayIn1/styles.scss is the shared view file used across all countries' Pay in 1 modal, not GB-specific.

It ended up duplicated because a later commit re-added the same block. Removed the duplicate; fix is pushed.

smoyalpaypal (approved):

dist/bizcomponents/js/versioned/messaging@1.85.0.js
surekhaw, line 1: /dist files should not be included in the PR

tests/functional/snapshots/GB/DEV000GBPI30Q/flex/color-blue_ratio-1x1-100-snap.png
surekhaw, line 1: Why are some of these flex snapshots blank?

tests/functional/v2/snapshots/modal/sdk/GB/DEV_GB_MULTI/desktop/200-Navigates_to_product_list_via_See_other_ways_link-desktop-snap.png
surekhaw, line 1: blank?

src/library/controllers/modal/interface.js
```diff
@@ -139,8 +139,15 @@ const memoizedModal = memoizeOnProps(
 
             // TODO: Remove after DE universal modal ramp is complete. Allows both old and new modals to work with DE messages while ramping.
             const validDEProductValues = ['PAY_LATER_LONG_TERM', 'PAY_LATER_PAY_IN_1', 'PAY_LATER_SHORT_TERM'];
+            // TODO: Remove after GB universal modal ramp is complete. Allows both Pi30 and Pay in 3 modals to work with GB messages while ramping.
```
Braluna-pp, line 142: This adds GB-specific product policy into the shared modal controller:

```js
const validGBProductValues = ['PAY_LATER_PAY_IN_1', 'PAY_LATER_SHORT_TERM'];
...
} else if (options?.offerCountry === 'GB') {
    productState = validGBProductValues;
}
```

This weakens the guard from “is this requested product actually available in the current modal state?” to “is this country GB?”. It also turns this shared path into a country switchboard, next to the existing DE exception.

Can we move this rule into locale/config data and keep this controller generic? Ideally this validation still checks the real `zoidComponent.state.products`, with any ramp-specific allowance coming from a product capability/config layer instead of hardcoded country branches.

src/server/locale/GB/index.js
```diff
@@ -1,13 +1,13 @@
-import getMutations from './mutations';
-import logos from '../../message/logos';
-import styles from './styles';
-import validOptions from './validOptions';
+import gpl from './GPL';
```
Braluna-pp: This split creates two near-identical GB locale trees:

```js
import gpl from './GPL';
import pi30 from './Pi30';
...
return pi30;
...
return gpl;
```

`GB/GPL` and `GB/Pi30` repeat the same locale wrapper shape, `localeClass`, `productName`, styles, and valid options. The real product-specific difference is mutation selection.

Can we keep one GB locale module and make the product-specific part a small mutation selector? That would remove duplicate directory surface and reduce the chance that future GB style/options changes drift between Pi3 and Pi30.

tests/functional/v2/config/GB/DEV_GB_MULTI.js
```diff
@@ -0,0 +1,27 @@
+export const DEV_GB_MULTI = {
+    testFileName: 'multiProductGB',
+    country: 'GB',
+    description: 'GB merchant eligible for Pay in 3 and Pay in 30 Days',
+    amounts: [
```
Braluna-pp, line 5: The new GB multi-product fixture tests happy-path amounts, but misses the exact requirement boundaries:

```js
amounts: [
    { value: '50.00', ... },
    { value: '200.00', ... }
]
```

The product requirement says Pi30 should show through `74.99`, and Pi3 should start at `75.00`. Without those exact cases, a cutover bug can pass these snapshots.

Can we add boundary cases for `74.99` and `75.00` here, and add `900.00` to the Pay-in-1-only fixture? Those are runtime business rules, not type-system coverage, so they are worth testing directly.

tests/functional/v2/testFn/multi.js
```diff
@@ -64,6 +86,12 @@ export const clickProductListTiles = async (contentWindow, modalContent, account
 
         // Switch to short term view
         await switchViews(3, 'longTerm');
+    } else if (account === 'DEV_GB_MULTI') {
```
Braluna-pp, line 89: This adds another account-specific branch into a shared helper:

```js
} else if (account === 'DEV_GB_MULTI') {
    await switchViews(2, 'shortTerm');
    await switchViews(3, 'payIn1');
}
```

This works, but it makes the shared product-list helper harder to maintain as more country/product combinations appear. It also only proves fixed child positions work; it does not directly assert the product-list requirement that ineligible products are hidden.

Can we move expected tile order/products into the account config and have this helper assert visible tile names/count before clicking? That keeps the helper generic and makes the requirement explicit.

dist/bizcomponents/js/versioned/messaging@1.85.0.js
```diff
@@ -0,0 +1,3 @@
+/* version: 1.85.0 */
```
Braluna-pp: This PR adds generated `dist` artifacts for `@1.85.0`:

```txt
dist/bizcomponents/js/versioned/messaging@1.85.0.js
```

But this branch’s `package.json` version is `1.89.0`, and `origin/develop` already has `@1.89.0` versioned bundles. These generated files look stale relative to the current branch.

Can we remove these `@1.85.0` dist files, or regenerate dist from the current release workflow if this PR is intentionally supposed to include assets? Keeping stale generated bundles makes review noisier and can confuse which artifact version is actually valid.

src/library/controllers/modal/interface.js
lyerabati16: Thanks for flagging this, agree the country branching shape in the controller isn't ideal,
On moving it into locale/config: not straightforward. locale/* and this modal controller are separate, independently-bundled targets with no existing cross-dependency, and there's no existing product capability layer in the repo to hook into building one would be new shared architecture, not a config swap, and affects every ramped country, not just GB. This pattern was originally introduced by Julia in the universal modal launch (#726); @surekhaw can you please help here , and two quick questions.

Is the DE override still needed, or has DE's ramp completed (per its own TODO)?
Does AT need this too? content/modals/AT/ already has the same long_term/pay_in_1/product_list shape as DE/GB looks like it may be a two-product market with the same gap, but we can't tell rollout status from this repo.

tests/functional/v2/testFn/multi.js
lyerabati16: @Braluna-pp I looked at moving expected tile order into config and asserting visible tiles before clicking, and can do that cleanly for GB alone without touching this shared helper, but that would leave clickProductListTiles itself exactly as branchy as it is today (GB's else if would stay, unchanged), since genuinely making this helper generic means removing the account-name branching for every account it serves, not just adding a parallel check for one.

Doing that properly means adding the same config-driven tile list to US/ES/IT/FR as well, and re-verifying the Puppeteer FT suite across all 5 countries not just GB since it changes shared code every one of them depends on. That's more scope than we want to take on inside a GB-specific PR, and a GB only version of the fix would be a halfmeasure that doesn't actually satisfy "keep the helper generic."

We'd  like to defer this as a follow-up covering all multi-product countries at once, rather than partially generalizing here.

Braluna-pp: @lyerabati16 , I agree lets make this a follow up PR

lyerabati16: Will create a ticket for this and follow up later, so resolving this comment

tests/functional/snapshots/GB/DEV000GBPI30Q/flex/color-blue_ratio-1x1-100-snap.png
lyerabati16: @surekhaw Flex snapshot: Not a bug, checked it byte-for-byte (SHA-256) against FR's existing baseline and they're identical. Every locale renders blank at this 100px height; GB's just doing the same thing FR/ES/IT/US already do.

SDK desktop snapshot: Also not a bug. The SDK modal opens as an actual browser popup on desktop, and Puppeteer's screenshot can only capture the original tab, so it grabs the grey backdrop behind the popup instead of the popup itself. This exact image has been the checked-in baseline for sdk+desktop tests across every country here since 2022, so it's a long-standing quirk, not something GB introduced.

tests/functional/v2/snapshots/modal/sdk/GB/DEV_GB_MULTI/desktop/200-Navigates_to_product_list_via_See_other_ways_link-desktop-snap.png
lyerabati16: As mentioned in the above, its an existing behaviour

content/modals/GB/short_term.json
```diff
@@ -42,10 +42,11 @@
             "Remaining payments are taken automatically."
         ],
         "disclosure": [
-            "Pay in 3 is a credit agreement. Carefully consider whether the purchase is affordable and how you will make the repayments. Be aware of the possible impact of using Pay in 3 and of missing payments, including making other borrowing more difficult or more expensive. Pay in 3 eligibility is subject to status and approval. 18+ UK residents only. Available for purchases of {formattedMinAmount} to {formattedMaxAmount}. ",
```
surekhaw, line 45: legal disclosure is now `Pay in 3 is a credit agreement. Carefully consider...`. This change should be reverted.

content/modals/GB/short_term_xo.json
```diff
@@ -74,7 +74,7 @@
             "link": "https://www.paypal.com/link/to/short-term-application"
         },
         "disclosure": [
-            "Pay in 3 is a credit agreement. Carefully consider whether the purchase is affordable and how you will make the repayments. Be aware of the possible impact of using Pay in 3 and of missing payments, including making other borrowing more difficult or more expensive. Pay in 3 eligibility is subject to status and approval. 18+ UK residents only. Available for purchases of {formattedMinAmount} to {formattedMaxAmount}. ",
```
surekhaw, line 77: revert legal disclosure update

content/modals/GB/pay_in_1.json
```diff
@@ -0,0 +1,47 @@
...
+    "content": {
+        "headline": "Buy now. Pay in 30 Days.",
+        "subheadline": "Available for purchases of {formattedMinAmount} - {formattedMaxAmount}. Eligibility subject to status.",
+        "instructions": [
+            "Apply for Pay in 30 Days.",
+            "If approved, complete your purchase.",
+            "One repayment is taken automatically 30 days after your order is processed."
+        ],
+        "disclosure": [
+            "Pay in 30 Days is an unregulated credit agreement, so you will have fewer protections than under a regulated credit agreement. Carefully consider whether the purchase is affordable and how you will repay. Missing payment may affect your ability to use us in the future. Eligibility is subject to status and approval. 18+ UK residents only. Available for purchases £1 - £900. ",
```
surekhaw: current legal disclosure is now `Pay in 30 Days is a credit agreement. Carefully consider whether the purchase is affordable and how you will repay. Missing payment may affect your ability……`

content/modals/GB/pay_in_1.json
```diff
@@ -0,0 +1,47 @@
...
+            [
+                "Terms and conditions apply",
```
surekhaw, line 36: current terms and conditions link is `https://www.paypal.com/uk/digital-wallet/paypal-payin30/terms`

utils/devServerProxy/config/GB/index.js
```diff
@@ -1,4 +1,100 @@
...
+                    minAmount: 1,
+                    maxAmount: 900
+                }
+            ],
+            PAY_LATER_SHORT_TERM: [
+                {
+                    totalPayments: 3,
+                    apr: 0,
+                    nominalRate: 0,
+                    minAmount: 30,
```
surekhaw: this should be 20

content/modals/GB/product_list.json
```diff
@@ -0,0 +1,29 @@
...
+                    "viewName": "PAY_LATER_SHORT_TERM"
+                },
+                {
+                    "header": "Pay in 30 Days",
+                    "body": "Full amount due in 30 days, nothing due today. Interest-free.",
+                    "viewName": "PAY_LATER_PAY_IN_1"
+                }
+            ]
+        },
+        "disclosure": "Pay in 3 and Pay in 30 Days are unregulated credit agreements. Missing payments can impact future borrowing. Eligibility subject to status. Terms and conditions apply."
```
surekhaw: legal disclosure has been updated to `Pay in 3 and Pay in 30 Days are credit agreements...`

src/server/locale/GB/mutations/pi30nq.js
```diff
@@ -0,0 +1,116 @@
...
+                    {
+                        tag: 'medium',
+                        br: ['purchases']
+                    },
+                    { tag: 'xsmall' }
+                ]
+            })
+        ],
+        [
+            'logo.type:inline',
```
surekhaw, line 77: there is an extra period: 
Image

src/server/locale/GB/mutations/pi30nq.js
```diff
@@ -0,0 +1,116 @@
...
+                    textWrap(textSize * 55, textSize, 'GB'),
+                    xSmallFallback(textSize * 16),
+                    altNoWrap(textSize * 10.6),
+                    messageLogoWidth(textSize * 1.75, textSize * 4, textSize * 1.25)
+                ],
+                logo: Logo.PP_PAYPAL.COLOR[0]
+            })
+        ],
+        [
+            'logo.type:none',
```
surekhaw, line 62: there is an extra period:

Image

src/server/locale/GB/mutations/pi30q.js
```diff
@@ -0,0 +1,94 @@
...
+                    textWrap(textSize * 32, textSize, 'GB'),
+                    xSmallFallback(textSize * 11.5),
+                    altNoWrap(textSize * 10.6),
+                    messageLogoWidth(textSize * 1.75, textSize * 4, textSize * 1.25)
+                ],
+                logo: Logo.PP_PAYPAL.COLOR[0]
+            })
+        ],
+        [
+            'logo.type:none',
```
surekhaw, line 59: extra period
Image

src/server/locale/GB/mutations/pi30q.js
```diff
@@ -0,0 +1,94 @@
...
+            ({ textSize }) => ({
+                styles: [xSmallFallback(textSize * 18)],
+                logo: false,
+                headline: [{ tag: 'medium' }, { tag: 'xsmall' }]
+            })
+        ],
+        [
+            'logo.type:inline',
```
surekhaw, line 70: extra period

Image

content/modals/GB/pay_in_1.json
lyerabati16: @surekhaw Is this updated not from July15? this should be updated now? https://www.figma.com/design/WZ10fLxKXKFhDkzoiMIslM/Upstream-5.1-Rebrand?node-id=6197-3078&m=dev

surekhaw: yes,  please update now so that it will be live for July 15 release

src/server/locale/GB/mutations/pi30nq.js
surekhaw: please remove extra period after Credit option so that it reads Credit option with PayPal not Credit option. with PayPal.

src/server/locale/GB/mutations/pi30nq.js
```diff
@@ -0,0 +1,116 @@
+import Logo from '../../../message/logos';
+import {
+    textWrap,
+    messageLogoWidth,
+    xSmallFallback,
+    logo20x1,
+    altNoWrap,
+    setLogoTop
+} from '../../../message/mediaQueries';
+import { textLogoMutations, flexLogoMutations } from '../../../message/logoMutations';
```
surekhaw, line 10: Credit option is different text size for Pi30 Q and NQ flex than for Pi3 flex. Was  there a design requirement / guidance to indicate a preference or a different experience for Pi30?

Image

src/library/controllers/modal/interface.js
surekhaw: The DE TODO comment is 4 years old, so I'm quite confident that the ramp completed, but I am unable to confirm definitively

content/modals/GB/product_list.json
```diff
@@ -0,0 +1,29 @@
+{
+    "meta": {
+        "product": "PRODUCT_LIST",
+        "useV5Design": "true",
+        "v5.1": "true"
+    },
+    "content": {
+        "headline": "Buy now, pay later",
+        "subheadline": "Get more info on Pay Later options.",
+        "instructions": {
```
surekhaw, line 10: I don't see this line in the figma

surekhaw (approved):

PR #1322, opened 2026-05-22
Implement UM token for CYSP prequalification flow

perco12:
## Description

Implements auto-generated session token for the prequalification flow, removing the external token dependency.

## Screenshots / Videos

N/A

## Testing instructions

Unit tests cover token format, uniqueness, and URL construction.


saikiran-m7 (approved):
Tested it locally, UM token is generating properly.

Image

yeul (approved):

PR #1331, opened 2026-06-10
feat(DTCRCMERC-4687): implement v5 text-layout styles for renderV2Message

sailaya99:
Description

Implements v5 text-layout styles for the renderV2Message SSR module (DTCRCMERC-4687).

Previously, renderV2Message rendered with a bare .pp-message container and minimal CSS — no color treatment, no logo positioning styles, no font configuration, and no alignment support. This PR introduces a static, class-based stylesheet (no legacy mutation system) that covers all v5 text style options:

- text.color → CSS class on .main/.action spans (black, white, monochrome, grayscale)
- text.size → font-size in px (10–16)
- text.align → text-align on .pp-message (left, center, right)
- logo.position → margin/display treatment on .logo (left, right, top)
- logo.type → passed through as a root data-pp-style-logo-type attribute
- text.fontFamily / text.fontSource → @font-face generation with https-only URL validation and XSS-safe font name filtering
- greyscale alias normalized to grayscale via validateStyle before reaching render()

Root data-pp-style-* attributes are added to the .pp-message div for downstream observability and potential CSS hook use.


tests/unit/spec/server/v2/render.test.js
```diff
@@ -219,4 +232,203 @@ describe('v2 render', () => {
...
+
+    test.each([['primary'], ['alternative'], ['inline'], ['none']])('logo type: %s', logoType => {
+        const options = {
+            style: {
+                layout: 'text',
+                logo: { type: logoType, position: 'left' },
+                text: { color: 'black', size: 12, align: 'left' }
+            }
+        };
+        expect(render(options, contentWithLogo)).toMatchSnapshot();
```
Braluna-pp: This snapshots the full rendered HTML for every matrix case, which creates a 1,240-line snapshot file because each case repeats the entire inline stylesheet. That makes future CSS changes noisy: one real stylesheet change will rewrite many snapshot blocks.

Problematic pattern:

```js
expect(render(options, contentWithLogo)).toMatchSnapshot();
```

Can we keep one full representative render snapshot, snapshot the stylesheet once, and use focused matrix assertions for the cases? For example:

```js
test('renders the v2 stylesheet once', () => {
    const result = render(baseOptions, baseV2Content);
    expect(result.match(/<style>[\s\S]*<\/style>/)[0]).toMatchSnapshot();
});

test.each([['primary'], ['alternative'], ['inline'], ['none']])('maps logo type: %s', logoType => {
    const result = render(
        {
            style: {
                layout: 'text',
                logo: { type: logoType, position: 'left' },
                text: { color: 'black', size: 12, align: 'left' }
            }
        },
        contentWithLogo
    );

    expect(result).toContain(`data-pp-style-logo-type="${logoType}"`);
});
```

That keeps the acceptance coverage while avoiding a huge low-signal snapshot file.

Braluna-pp (commented):
Thermo-nuclear code quality review: a few maintainability and coverage issues worth tightening before this lands.

src/server/v2/message.jsx
```diff
@@ -53,12 +53,21 @@ export default function V2Message({ options, v2Content }) {
     const actionLabel = buildContentLabel(actionItems);
 
     return (
-        <div className="pp-message">
+        <div
+            className="pp-message"
+            data-pp-style-layout={style.layout}
+            data-pp-style-logo-position={logoPosition}
```
Braluna-pp: This emits `data-pp-style-logo-position="inline"` when `logo.type` is `inline`, because `inline` is being modeled as a pseudo-position. That muddies the contract: `inline` is a logo type, while the accepted v5 positions are `left`, `right`, and `top`.

Problematic source:

```js
const logoPosition = style.logo?.type === 'inline' ? 'inline' : style.logo?.position ?? 'left';
const logoType = style.logo?.type ?? 'primary';
```

Can we keep type and position separate and pass both into the logo helper? For example:

```js
const logoType = style.logo?.type ?? 'primary';
const logoPosition = style.logo?.position ?? 'left';

const { logoBlock, hasInitialLogo, hasRightLogo, mainBlocks } = buildLogoConfiguration({
    logoType,
    logoPosition,
    mainItems
});
```

Then the helper can branch on `logoType === 'inline'` without inventing a fourth position:

```js
export function buildLogoConfiguration({ logoType, logoPosition, mainItems }) {
    if (logoType === 'inline') {
        return { hasInitialLogo: false, hasRightLogo: false, logoBlock: undefined, mainBlocks: mainItems };
    }

    const logoBlock = mainItems.find(item => item.type === 'IMAGE');
    const mainBlocks = mainItems.filter(item => item.type !== 'IMAGE');

    return {
        hasInitialLogo: !!logoBlock && (logoPosition === 'left' || logoPosition === 'top'),
        hasRightLogo: !!logoBlock && logoPosition === 'right',
        logoBlock,
        mainBlocks
    };
}
```

tests/unit/spec/server/v2/render.test.js
```diff
@@ -219,4 +232,203 @@ describe('v2 render', () => {
...
+    test('maps validated v5 text style options to root data attributes', () => {
+        const options = {
+            style: {
+                layout: 'text',
+                logo: { type: 'primary', position: 'top' },
+                text: { align: 'center', color: 'white', size: 16 }
+            }
+        };
+        const result = render(options, baseV2Content, mockLog);
+        expect(result).toContain('data-pp-style-layout="text"');
```
Braluna-pp, line 391: This proves the SSR output contains the expected data attributes, but it still does not prove AC 12: merchant page CSS cannot override the v2 text message stylesheet under the normal SDK iframe/message rendering path.

Current coverage is essentially:

```js
expect(result).toContain('data-pp-style-layout="text"');
expect(result).toContain('data-pp-style-logo-position="top"');
expect(result).toContain('data-pp-style-logo-type="primary"');
```

Can we add a small integration-style test around the normal message iframe path, or link this story to an existing test that already proves iframe isolation? Shape of the missing proof:

```js
test('renders v2 message styles inside the iframe document, isolated from merchant page CSS', async () => {
    // Arrange merchant page CSS that would override .pp-message if rendering happened in-page.
    // Render through the normal message iframe path.
    // Assert the iframe document contains the v2 stylesheet and computed styles match v2 CSS.
});
```

Without that, the PR proves the stylesheet is emitted, but not that merchant CSS cannot override it in the production rendering path.

tests/unit/spec/server/v2/render.test.js
sailaya99: Done — snapshot file is down to 119 lines. Kept one full render snapshot, one stylesheet snapshot, and replaced the matrix cases with test.each + toContain() assertions so a CSS change only touches the two snapshot blocks, not every case.

src/server/v2/message.jsx
sailaya99: Fixed. message.jsx now derives logoType and logoPosition independently and passes both to buildLogoConfiguration. The helper branches on logoType === 'inline' directly without inventing a fourth position value, so data-pp-style-logo-position always reflects the real position.

tests/unit/spec/server/v2/render.test.js
sailaya99: Added a v2 render stylesheet isolation describe block with two tests: one asserting the stylesheet is embedded inline (no <link>), and one parsing the CSS output to assert every selector is scoped to .pp-message or body. This is the extent of what's provable at the unit test level. A true computed-style assertion against merchant CSS overrides would require Playwright/Puppeteer infrastructure not currently in this repo — happy to add it or link to an existing E2E test if one covers this path.

src/server/v2/utils/buildLogoConfiguration.js
```diff
@@ -1,5 +1,5 @@
-export function buildLogoConfiguration({ logoPosition, mainItems }) {
-    if (logoPosition === 'inline') {
+export function buildLogoConfiguration({ logoType, logoPosition, mainItems }) {
+    if (logoType === 'inline') {
```
Braluna-pp: **Description**

Inline logos bypass the new logo color styling. `logo.type="inline"` keeps the image inside `mainBlocks`, so it is rendered by the generic `renderBlock()` path under `.main`. The new white/monochrome/grayscale filters only apply to `.logo.* img`, so inline logos miss the styling path. This splits logo behavior across two render paths and makes the inline case easy to break.

**Problematic code snippet**

```js
if (logoType === 'inline') {
    return { hasInitialLogo: false, hasRightLogo: false, logoBlock: undefined, mainBlocks: mainItems };
}
```

```css
.pp-message .logo.white img { filter: brightness(0) invert(1); }
.pp-message .logo.monochrome img { filter: grayscale(100%) brightness(0); }
.pp-message .logo.grayscale img { filter: grayscale(100%); }
```

**Suggested code snippet change**

```js
// Keep logo rendering behind one logo-aware path, including inline.
const logoClasses = mapClasses({
    logo: true,
    inline: logoType === 'inline',
    [textColor]: true,
    [logoType]: true
});

function renderLogo(block, className = logoClasses) {
    return block ? (
        <span role="img" aria-label={block.alternative_text || 'PayPal'} className={className}>
            {renderBlock(block)}
        </span>
    ) : null;
}
```

Then render inline logos through `renderLogo()` instead of leaving the image inside the generic `.main` path.

src/server/v2/styles.js
```diff
@@ -1,15 +1,62 @@
...
+    'monospace',
+    'cursive',
+    'fantasy',
+    'system-ui',
+    'ui-serif',
+    'ui-sans-serif',
+    'ui-monospace'
+]);
+
+const isSafeFontName = value => typeof value === 'string' && /^[\w\s-]+$/.test(value.trim());
```
Braluna-pp: **Description**

This reimplements v5 font handling inside the v2 stylesheet generator even though the repo already owns this behavior in `src/server/message/font.js`. Safe font names, safe font sources, generated `@font-face` names, and fallback stacks now have two implementations that can drift. Since the ticket says to preserve existing v5 font behavior where applicable, the cleaner structure is to extract the shared font policy and have both renderers call it.

**Problematic code snippet**

```js
const isSafeFontName = value => typeof value === 'string' && /^[\w\s-]+$/.test(value.trim());
const isSafeFontSource = value => typeof value === 'string' && /^https:\/\/[^'")\s]+$/i.test(value);

function formatFontFamilyName(value) {
    if (GENERIC_FONT_FAMILIES.has(value)) {
        return value;
    }

    return `'${value}'`;
}
```

**Suggested code snippet change**

```js
// src/server/message/font.js or a shared src/server/fontRules.js
export function buildFontRules({ style, selector, fallbackFamily }) {
    // one shared implementation for:
    // - fontSource filtering
    // - fontFamily formatting
    // - @font-face generation
    // - fallback stack
    // - text.size handling
}
```

```js
// src/server/v2/styles.js
import { buildFontRules } from '../message/font';

const fontRules = buildFontRules({
    style: { layout: 'text', text: { fontFamily, fontSource, size: fontSize } },
    selector: '.pp-message',
    fallbackFamily: DEFAULT_FONT_FAMILY
});
```

tests/unit/spec/server/v2/render.test.js
```diff
@@ -219,4 +232,243 @@ describe('v2 render', () => {
...
+    test.each([['black'], ['white'], ['monochrome'], ['grayscale']])('maps text color: %s', color => {
+        const options = {
+            style: {
+                layout: 'text',
+                logo: { type: 'primary', position: 'left' },
+                text: { color, size: 12, align: 'left' }
+            }
+        };
+        const result = render(options, baseV2Content);
+        expect(result).toContain(`data-pp-style-text-color="${color}"`);
```
Braluna-pp: **Description**

The new matrix tests mostly prove that `data-pp-style-*` attributes are present, not that the renderer applies the actual styling behavior. That leaves room for CSS/rendering gaps to pass tests, including the inline-logo color issue above. These tests should assert rendered classes and generated CSS for representative combinations.

**Problematic code snippet**

```js
test.each([['black'], ['white'], ['monochrome'], ['grayscale']])('maps text color: %s', color => {
    const result = render(options, baseV2Content);
    expect(result).toContain(`data-pp-style-text-color="${color}"`);
});
```

**Suggested code snippet change**

```js
test.each([['black'], ['white'], ['monochrome'], ['grayscale']])('applies text color class and CSS: %s', color => {
    const result = render(
        {
            style: {
                layout: 'text',
                logo: { type: 'primary', position: 'left' },
                text: { color, size: 12, align: 'left' }
            }
        },
        baseV2Content
    );

    expect(result).toMatch(new RegExp(`class="main[^"]*${color}`));
    expect(result).toContain(`.pp-message .main.${color}`);
});
```

Add a focused cross-case too:

```js
test('applies white logo styling when logo type is inline', () => {
    const result = render(
        {
            style: {
                layout: 'text',
                logo: { type: 'inline', position: 'left' },
                text: { color: 'white', size: 12, align: 'left' }
            }
        },
        contentWithLogo
    );

    expect(result).toMatch(/class="logo[^"]*white[^"]*inline/);
    expect(result).toContain('.pp-message .logo.white img');
});
```

Braluna-pp (commented):
Thermo-nuclear code-quality review: leaving three maintainability findings inline.

src/server/v2/utils/buildLogoConfiguration.js
sailaya99: Fixed. buildLogoConfiguration now extracts the IMAGE block from mainItems before the inline branch, so logoBlock is set and mainBlocks excludes the image for all logo types. A renderLogo() helper is extracted in message.jsx and called inside the .main span for inline logos, wrapping the image in .logo.inline.{color} — putting it on the same CSS path as left/right/top and applying the color-filter rules.

src/server/v2/styles.js
sailaya99: Fixed. Added a buildFontRules({ fontSource, fontFamily, fallbackStack, defaultFontFamily, fontNamePrefix }) export to src/server/message/font.js using the shared isSafeFontSource, isSafeFontName, and formatFontFamilyName helpers already in that file. styles.js now imports and calls it, removing the five local declarations (GENERIC_FONT_FAMILIES, isSafeFontName, isSafeFontSource, formatFontFamilyName, and the inline @font-face loop). CSS output is identical.

tests/unit/spec/server/v2/render.test.js
sailaya99: Fixed. The text-color test.each now also asserts the CSS class on .main (class="main[^"]*${color}") and the presence of the generated CSS rule (.pp-message .main.${color}) in the stylesheet. Added a focused inline + white test that checks both the logo white inline class string and .pp-message .logo.white img in the embedded stylesheet. Updated the stale inline logo test to reflect that inline logos are now rendered via .logo.inline rather than left as bare images in the generic main path.

Braluna-pp (approved):
LGTM thank you for addressing the feedback

surekhaw (approved):

src/server/v2/message.jsx
jeremy-herman, line 120: I’m concerned that `renderV2Message` currently renders logo image blocks directly from `item.source_url`. That does not match legacy PMC behavior or the v6 renderer closely enough for first-party logos.

Legacy PMC does use ``, but the `src` comes from local PayPal logo assets generated through `@paypal/sdk-logos` and selected by style mutations like `logo.type` and `text.color`. v6 similarly does not render arbitrary image URLs for brand logos; it maps image content blocks into known local SVG templates via `brandTemplate`/`graphicTemplate`.

For PayPal and PP Credit logos, I think we should resolve the logo locally from the semantic image block plus `logo.type`, `logo.position`, and text color. That keeps rendering deterministic, makes `logo.type` actually control the displayed logo variant, and avoids depending on CPS-provided asset URLs for core brand presentation.

A reasonable implementation path for this PR would be to reuse or adapt the existing legacy PMC logo resolver/data-URI assets rather than porting the full v6 inline SVG implementation. If `source_url` still needs to be supported, I’d treat it as fallback-only and document/validate or allowlist it explicitly.

Braluna-pp: I agree with Jeremy’s concern here. The current implementation makes the v2 `IMAGE` block URL the source of truth for first-party branding:

```jsx
case 'IMAGE':
    return <img src={item.source_url} alt={item.alternative_text || 'PayPal'} />;
```

That means `logo.type`, `logo.position`, and `text.color` mostly affect wrapper classes/CSS, but they do not actually choose the logo asset. For example, `logo.type: primary` and `logo.type: alternative` can still render the same CPS-provided `source_url`.

Legacy PMC handles this differently. The logo asset is renderer-owned, and the style options select the variant:

```js
['text.color:white && logo.type:primary', { logo: Logo.PP_PAYPAL.WHITE }],
['text.color:grayscale && logo.type:primary', { logo: Logo.PP_PAYPAL.GRAYSCALE }],
['text.color:monochrome && logo.type:primary', { logo: Logo.PP_PAYPAL.MONOCHROME }],

['text.color:white && logo.type:alternative', { logo: Logo.PP_PAYPAL.WHITE[0] }],
```

Then the renderer uses the resolved local/logo object:

```jsx
{logos.map(({ src, dimensions: [width, height] }) => (
    <div className="message__logo message__logo--svg">
        <img src={src} alt="" role="presentation" />
        <canvas height={height} width={width} />
    </div>
))}
```

I think the right path is to keep this PR’s simple v2 renderer shape, but add a small first-party logo resolver before rendering the image. Something like:

```js
resolveLogoBlock({
    block: logoBlock,
    logoType,
    logoPosition,
    textColor
});
```

For known first-party blocks like `paypal_logo` and PayPal Credit logo names, that resolver should pick from the existing PMC logo assets based on `logo.type` and `text.color`. For unknown/non-first-party image blocks, `source_url` can remain the fallback path.

So the behavior would be:

```js
if (isKnownPayPalLogo(block.name)) {
    return getLocalPayPalLogo({ logoType, textColor });
}

if (isKnownPayPalCreditLogo(block.name)) {
    return getLocalPayPalCreditLogo({ logoType, logoPosition, textColor });
}

return { src: block.source_url };
```

That keeps the implementation small, avoids bringing back the full v5 mutation system, but still makes the core PayPal brand rendering deterministic and style-driven.

I’d also add tests for:
- known `paypal_logo` does not render the CPS `source_url`
- `logo.type: primary` vs `alternative` resolves different local assets
- `text.color: white/monochrome/grayscale` resolves the matching local variant
- unknown image blocks still fall back to `source_url`

jeremy-herman (approved):

PR #1339, opened 2026-06-26
feat: add flex layout support to renderV2Message

sailaya99:
## Description

Adds flex layout support to `renderV2Message` (DTCRCMERC-5374), following the same static CSS/class-based approach used by the text layout renderer. The flex renderer does not depend on `getMutations`, `applyCascade`, locale/product mutation files, or any v5 mutation cascade logic.

**Changes:**
- `src/server/v2/message.jsx` — adds `FlexMessage` component that extracts logo from `main_items` and renders `pp-flex__logo-container`, `pp-flex__messaging`, `pp-flex__main`, `pp-flex__action`, and `pp-flex__disclaimer` sections directly from CPS v2 content blocks
- `src/server/v2/styles.js` — adds `flexStyles()` covering all 7 color themes (`blue`, `black`, `white`, `white-no-border`, `gray`, `monochrome`, `grayscale`) and all 4 ratio layouts (`1x1`, `1x4`, `8x1`, `20x1`) with responsive media queries; color and ratio are applied via static CSS class selectors, not dynamic injection
- `src/server/v2/getParentStyles.js` — existing `ratioMap` already handles flex ratio outer iframe sizing; no changes needed
- `src/server/message/font.js` — adds shared `buildFontRules` utility used by v2 stylesheets (text and flex)
- `src/server/v2/utils/buildLogoConfiguration.js` — updates to accept `logoType` and correctly handle inline logo extraction
- Tests and snapshots for all new flex behavior

Color/ratio values are normalized before reaching `renderV2Message` by the existing `validateStyle` logic (`grey`→`gray`, `greyscale`→`grayscale`).

## Screenshots / Videos

N/A

## Testing instructions

- `npm test` — 152 unit tests pass covering all color and ratio combinations, content block structure, stylesheet isolation, alias normalization, and snapshots
- Flex rendering can be exercised locally via `node scripts/preview-v2-render.js` with `style.layout: "flex"` options


src/server/v2/styles.js
```diff
@@ -1,15 +1,226 @@
...
+    return buildFontRules({
+        fontSource,
+        fontFamily,
+        fallbackStack: FONT_FALLBACKS,
+        defaultFontFamily: DEFAULT_FONT_FAMILY,
+        fontNamePrefix: 'PP Merchant Font'
+    });
+}
+
+function flexStyles({ fontFamily, fontSource } = {}) {
```
Braluna-pp: This adds a large hand-written flex stylesheet directly into the shared v2 styles module. The file now mixes font setup, text styles, seven color themes, four ratio layouts, and responsive behavior in one template string, which makes future flex changes harder to review because a small theme or ratio change means scanning the whole CSS blob.

Problematic shape:
```js
function flexStyles({ fontFamily, fontSource } = {}) {
    // themes, ratios, media queries, and font setup all live here
}
```

Can we split flex into its own module and make the themes/ratios table-driven? For example, keep `textStyles` and `flexStyles` in separate files, define `FLEX_THEMES` and `FLEX_RATIOS`, then render the selectors from those maps. Same output, but fewer repeated selectors and a clearer owner for flex styling.

src/server/message/font.js
```diff
@@ -90,3 +90,25 @@ export const getFontRules = style => {
     }
     return rules;
 };
+
+// Shared validation and @font-face generation for v2 stylesheet.
+const isSafeFontName = value => typeof value === 'string' && /^[\w\s-]+$/.test(value.trim());
+const isSafeFontSource = value => typeof value === 'string' && /^https:\/\/[^'")\s]+$/i.test(value);
+
+export function buildFontRules({ fontSource, fontFamily, fallbackStack, defaultFontFamily, fontNamePrefix }) {
```
Braluna-pp: This puts a second font-generation path into the legacy message font module, but the existing `getFontRules` path above still uses different naming, fallback, and validation behavior. That leaves one file owning two contracts that can drift.

Problematic shape:
```js
export function buildFontRules(...) { ... }
```

Can we either move this v2-only helper under `src/server/v2` or fully refactor `getFontRules` to use the same shared primitive? Right now the abstraction looks shared, but only v2 uses it, so it adds cross-layer coupling without deleting the old complexity.

tests/unit/spec/server/v2/render.test.js
```diff
@@ -219,4 +236,483 @@ describe('v2 render', () => {
...
+    };
+
+    test.each([
+        ['blue', '8x1'],
+        ['black', '8x1'],
+        ['white', '1x1'],
+        ['gray', '1x4'],
+        ['monochrome', '20x1']
+    ])('full render snapshot for flex color=%s ratio=%s', (color, ratio) => {
+        const options = { style: { layout: 'flex', color, ratio } };
```
Braluna-pp: These full render snapshots add a new 1,235-line snapshot file. That is too much generated output for this change, and it makes the review depend on scanning huge snapshots instead of clear expectations. It also duplicates many string assertions already covered above.

Problematic shape:
```js
expect(render(options, flexContentWithLogo)).toMatchSnapshot();
```

Can we shrink this to one representative HTML snapshot plus semantic assertions for the full color/ratio matrix? The ratio/theme coverage can stay table-driven, but the tests should assert the meaningful contract directly: root classes, data attributes, key selectors, and logo/action/disclaimer structure.

src/server/v2/message.jsx
```diff
@@ -27,17 +27,102 @@ function renderBlock(item) {
...
+function renderLogo(block, className) {
+    if (!block) return null;
+    return (
+        <span role="img" aria-label={block.alternative_text || 'PayPal'} className={className}>
+            {renderBlock(block)}
+        </span>
+    );
+}
+
+function FlexMessage({ style, v2Content }) {
```
Braluna-pp: This adds the whole flex renderer inside `message.jsx`, while the same file still owns the text renderer and shared block rendering. The branch at `style.layout === 'flex'` is clean today, but this file is now the place where every future layout can grow more renderer-specific logic.

Problematic shape:
```jsx
function FlexMessage({ style, v2Content }) { ... }
```

Can we move `FlexMessage` into a dedicated v2 flex module and keep `message.jsx` as a small dispatcher? The shared helpers like `renderBlock`, `renderLogo`, and content extraction can stay reusable, but the layout-specific structure should live behind its own component so this does not become a multi-layout grab bag.

Braluna-pp (commented):
Thermo-nuclear code quality review focused on maintainability, structure, and reviewability. These are not behavior blockers from a test run; they are structural review comments from the PR diff and Jira scope.

src/server/v2/styles.js
sailaya99: Done. flexStyles.js now has FLEX_THEMES, PORTRAIT_RATIOS, and LANDSCAPE_RATIOS data tables. A lscpSel() helper expands LANDSCAPE_RATIOS into grouped selectors everywhere, including inside @media blocks. CSS output is unchanged; all 95 tests pass.

src/server/message/font.js
sailaya99: buildFontRules is in src/server/v2/font.js. The only change to src/server/message/font.js on this branch is a trailing-newline removal — no new exports were added there.

tests/unit/spec/server/v2/render.test.js
sailaya99: The snapshot file is 491 lines. The color/ratio matrix uses semantic toContain assertions; only two flex snapshots exist: one full HTML render (blue/8x1) and one for the stylesheet.

src/server/v2/message.jsx
sailaya99: FlexMessage is in src/server/v2/flex.jsx. message.jsx just imports and dispatches to it.

src/server/v2/flex.jsx
```diff
@@ -0,0 +1,76 @@
...
+/** @jsxFrag Fragment */
+import { h, Fragment } from 'preact';
+
+import { buildContentLabel } from './utils/buildContentLabel';
+import { renderBlock } from './utils/renderBlock';
+import flexStyles from './flexStyles';
+
+export default function FlexMessage({ style, v2Content }) {
+    const color = style.color ?? 'black';
+    const ratio = style.ratio ?? '8x1';
```
Braluna-pp: This creates a correctness edge case: the direct flex renderer defaults do not match the validated flex defaults. `validOptions` defaults flex to `color: "blue"` and `ratio: "1x1"` because those are the first valid values, but this lower-level render path falls back to `black` and `8x1`.

That means `render({ style: { layout: "flex" } })` can produce `class="pp-message pp-flex black r-8x1"`, while the validated path produces `blue r-1x1`. Since `render`, `validateStyle`, and `getParentStyles` are all exported from `src/server/v2/index.js`, callers can hit this mismatch and end up with a rendered message that does not match the wrapper/default style contract.

Can we make these defaults come from the same source? At minimum this should use `blue` and `1x1` here, and ideally the flex defaults should live in one shared constant used by both `validOptions` and `FlexMessage`. Please also add a small test for `render({ style: { layout: "flex" } })` so this stays aligned.

sailaya99: Done. Added FLEX_DEFAULTS = { color: 'blue', ratio: '1x1' } to constants.js and updated flex.jsx to use it instead of the hardcoded black/8x1. Added a test for render({ style: { layout: 'flex' } }) that asserts the output matches validateStyle's defaults — any future drift will break the test.

surrsurus (commented):
Tested this visually with the harness that Jeremy put together - piped it through the local bundle built from this PR instead of the v5 stage bundle. Looks good, tested it with a variety of options. Just a few really minor things I noticed.

I noticed the width of the 1x1 and 1x4 didn't match but this might be due to the size of the zoid in the window - can you confirm if that should be the case?

I also noticed the 20x1 text doesn't look fully aligned with the logo, if this is intentional then feel free to ignore:

Image

## Before (cdn-v5_rewrite bundle)

| 1x1 | 1x4 | 8x1 | 20x1 |
| - | - | - | - |
| Image | Image | Image | Image |

## After (This bundle)

| 1x1 | 1x4 | 8x1 | 20x1 |
| - | - | - | - |
| Image | Image | Image | Image |

surrsurus:
Ignore my comment about the ratios - I'm now seeing where the widths are coming from in the `ratioMap`, these dimensions make more sense to me now:

```js
  const ratioMap = {
    '1x1': [
        {
            ratio: '1x1',
            width: [120, 300]
        }
    ],
    '1x4': [
        {
            ratio: '1x2',
            width: [160, 160]
        },
        {
            ratio: '1x4',
            breakpoint: 768
        }
    ],
```

PR #1367, opened 2026-07-21
feat: add flex layout support to renderV2Message (DTCRCMERC-5374)

surrsurus:
## Description

This PR is an attempt to cherry pick the flex changes from [Laya's PR](https://github.com/paypal/paypal-messaging-components/pull/1339) onto the latest develop. 

### Original Description

Adds flex layout support to `renderV2Message` (DTCRCMERC-5374), following the same static CSS/class-based approach used by the text layout renderer. The flex renderer does not depend on `getMutations`, `applyCascade`, locale/product mutation files, or any v5 mutation cascade logic.

**Changes:**
- `src/server/v2/message.jsx` — adds `FlexMessage` component that extracts logo from `main_items` and renders `pp-flex__logo-container`, `pp-flex__messaging`, `pp-flex__main`, `pp-flex__action`, and `pp-flex__disclaimer` sections directly from CPS v2 content blocks
- `src/server/v2/styles.js` — adds `flexStyles()` covering all 7 color themes (`blue`, `black`, `white`, `white-no-border`, `gray`, `monochrome`, `grayscale`) and all 4 ratio layouts (`1x1`, `1x4`, `8x1`, `20x1`) with responsive media queries; color and ratio are applied via static CSS class selectors, not dynamic injection
- `src/server/v2/getParentStyles.js` — existing `ratioMap` already handles flex ratio outer iframe sizing; no changes needed
- `src/server/message/font.js` — adds shared `buildFontRules` utility used by v2 stylesheets (text and flex)
- `src/server/v2/utils/buildLogoConfiguration.js` — updates to accept `logoType` and correctly handle inline logo extraction
- Tests and snapshots for all new flex behavior

Color/ratio values are normalized before reaching `renderV2Message` by the existing `validateStyle` logic (`grey`→`gray`, `greyscale`→`grayscale`).

### Author's Note

This fork was fairly out of date so this PR captures the additional effort to address rendering issues that popped due to all of the differences. The existing font and logo logic was broken when I put it on the latest develop and I also tried cleaning up the flex styles CSS & ratio logic. The files changed as a result look completely different.

## Screenshots / Videos

I'm using the v5 test harness to render these messages out

| 1x1 | 1x4 | 8x1 | 20x1 |
| - | - | - | - |
| image | image | image | image |

### Stage Tag

Test with tag `v5_rewrite_77046c1`

### v5 Harness

- Start CPS locally
- Point CPNW towards your local CPS:

```
        "@paypalcorp/creditpresentmentserv": {
            "hostname": "localhost.paypal.com",
            "port": 15606,
            "rejectUnauthorized": false
        }
```

- Point CPNW at your local bundle (`https://localhost.paypal.com:8080`) in `getMessagingBundle`
- Start CPNW, ClientSDKNodeWeb, and PMC
- Using the v5 rewrite test, create a bunch of flex messages

### Test page

`node scripts/preview-v2-render.js` apparently also can test this but I will be honest - I have no clue how to get this to work

## Review


src/server/v2/flexStyles.js
```diff
@@ -0,0 +1,484 @@
+import { buildFontRules } from '../message/font';
+
+const DEFAULT_FONT_FAMILY = '"PayPal Pro", Helvetica, Arial, "Liberation Sans", sans-serif';
```
jeremy-herman: I think this needs to preserve v5 flex’s default Helvetica/Arial stack. Could we use the v5 default here while retaining custom-font handling, and add a no-custom-font snapshot?

src/server/v2/utils/renderBlock.js
```diff
@@ -0,0 +1,22 @@
+/** @jsx h */
+import { h } from 'preact';
+
+export function renderBlock(item) {
```
jeremy-herman: Would it be possible to create a shared renderBlock for both flex and text layouts?

jeremy-herman (commented):
Some minor comments from the code review. Will do some testing next!

src/server/v2/utils/renderBlock.js
surrsurus: Yes, it'll need a new argument to accept the different layout options but should be doable. Working on it now

surrsurus: [3b4d5ca](https://github.com/paypal/paypal-messaging-components/pull/1367/commits/3b4d5ca7584e9a796a9a0f8f97f20983a0b88f76) Let me know if you're liking the way this looks

src/server/v2/flexStyles.js
surrsurus: Gotcha, wasn't fully aware of those requirements, should be fixed now in [54b82ce](https://github.com/paypal/paypal-messaging-components/pull/1367/commits/54b82ce7531af4a09d6836b7227644d9d7269270) + added a no custom font snapshot test

src/server/v2/utils/renderBlock.js
jeremy-herman: The click_url shouldn't be mapped to the iframe click and instead the entire flex banner should be a button that launches the smart/modal/ instead of going to a redirect link in the content.

surrsurus: The SDK already wraps the whole message in a `<button>` so this link can just be removed, this was something I overlooked from the original PR

surrsurus: [901d205](https://github.com/paypal/paypal-messaging-components/pull/1367/commits/901d2056eb67bf8eb316b26bbc33640e40979006)

tests/functional/spec/createBannerTest.js
surrsurus, line 1: These tests were pretty flaky so this is my attempt to fix that. Was noticing that the iframe existed before `document.body` was ready, this should help smooth that out in CI

src/server/v2/flexStyles.js
```diff
@@ -0,0 +1,484 @@
...
+
+button {
+    width: 100%;
+    border: none;
+    padding: 0;
+}
+
+html {
+    overflow: hidden;
+}
```
jeremy-herman, line 545: Could we set `font-size: 14px` on `html` here? The legacy renderer establishes a 14px root in `styles/common.css`, while v2 currently falls back to the browser’s 16px root. For example, GB 1x4 renders `1.1rem` as 17.6px instead of legacy’s 15.4px. This should preserve the existing responsive rules while restoring v5 sizing parity.

surrsurus: [cbeb2f9](https://github.com/paypal/paypal-messaging-components/pull/1367/commits/cbeb2f9dfa5b892990d007bfdfd476e9f23eb426) added as part of the flex styles fix for 1x1 and 8x1

jeremy-herman (approved):

src/server/v2/flex.jsx
```diff
@@ -0,0 +1,98 @@
...
+        logoName: logoBlock.name,
+        effectiveLogoType: 'wordmark',
+        effectiveLogoPosition: 'left',
+        textColor
+    });
+
+    if (assets) {
+        return assets.map(({ src, dimensions: [width, height] }, idx) => (
+            // eslint-disable-next-line react/no-array-index-key
+            <span key={idx} className="pp-flex__logo">
```
Braluna-pp: ### Finding Description

**[P1] Size single-piece brand logos as full lockups**

The renderer gives every resolved asset the same `pp-flex__logo` class, but the ratio CSS treats the first child as a small PayPal monogram and the second as the wider wordmark. PayPal Credit and Venmo each resolve to one combined lockup, so that full logo lands in the monogram slot. In 1x1 it is constrained to 29px wide, and in 1x4 to 27px. The PayPal Credit lockup is therefore only about 4px tall.

### Problematic Code Snippet(s)

```jsx
return assets.map(({ src, dimensions: [width, height] }, idx) => (
    <span key={idx} className="pp-flex__logo">
        <img src={src} alt="" role="presentation" width={width} height={height} />
    </span>
));
```

### Suggested Change(s)

- **Implementation:** Add brand or asset-shape classes and size single-piece PayPal Credit and Venmo lockups as whole logos instead of by PayPal asset position.
- **Test:** Cover PayPal Credit, Venmo, and fallback images in portrait and landscape flex ratios, asserting the selected class and rendered dimensions.

### Relevant Links

- [Changed renderer](https://github.com/paypal/paypal-messaging-components/blob/77046c128c1ba990a666c830846c1d9e73bf9f0a/src/server/v2/flex.jsx#L20-L25)
- [Ordinal portrait sizing](https://github.com/paypal/paypal-messaging-components/blob/77046c128c1ba990a666c830846c1d9e73bf9f0a/src/server/v2/flexStyles.js#L98-L118)
- [Single-piece brand asset resolution](https://github.com/paypal/paypal-messaging-components/blob/77046c128c1ba990a666c830846c1d9e73bf9f0a/src/server/v2/logos.js#L62-L86)
- [DTCRCMERC-5374](https://paypal.atlassian.net/browse/DTCRCMERC-5374)
- [Phase 1 HLD](https://paypal.atlassian.net/wiki/spaces/GPL/pages/2873114506/SDK+v5+Messages+Rewrite+Phase+1+HLD)

src/server/v2/utils/renderBlock.js
```diff
@@ -0,0 +1,62 @@
...
+                        {renderLogoImages(item, logoPresentation)}
+                    </span>
+                );
+            }
+            return <img src={item.source_url} alt={item.alternative_text || 'PayPal'} />;
+        case 'LINK':
+            return linkClassName ? <span className={linkClassName}>{item.text}</span> : <span>{item.text}</span>;
+        case 'TEXT':
+            return item.brand ? <strong>{item.text}</strong> : item.text;
+        default:
```
Braluna-pp, line 57: ### Finding Description

**[P1] Render supported TEXT_VARIABLE blocks**

The renderer handles `IMAGE`, `LINK`, and `TEXT`, then returns `item.text` for every other type. The repository's supported `TEXT_VARIABLE` fixture contains `name: 'installment_amount'` and no `text`, so this branch returns `undefined`. Flex sends every non-image block through this helper, causing valid variable content to disappear from both the visible message and its accessible label.

### Problematic Code Snippet(s)

```js
case 'TEXT':
    return item.brand ? <strong>{item.text}</strong> : item.text;
default:
    return item.text;
```

### Suggested Change(s)

- **Implementation:** Implement the CPS `TEXT_VARIABLE` contract at this boundary, passing any required resolution context instead of silently returning an absent `text` field.
- **Test:** Render a resolved variable between two text blocks and assert its visible position and inclusion in the accessible label. Add a defined unresolved-variable failure case.

### Relevant Links

- [Changed block renderer](https://github.com/paypal/paypal-messaging-components/blob/77046c128c1ba990a666c830846c1d9e73bf9f0a/src/server/v2/utils/renderBlock.js#L55-L60)
- [Current variable fixture](https://github.com/paypal/paypal-messaging-components/blob/77046c128c1ba990a666c830846c1d9e73bf9f0a/tests/unit/spec/server/v2/render.test.js#L722-L737)
- [DTCRCMERC-5374](https://paypal.atlassian.net/browse/DTCRCMERC-5374)
- [Messaging API V2 contract](https://paypal.atlassian.net/wiki/spaces/PPCMP/pages/2586232315/Upstream+Messaging+API+V2+Dev+Docs)

src/server/v2/flex.jsx
```diff
@@ -0,0 +1,98 @@
...
+
+export default function FlexMessage({ style, v2Content }) {
+    const color = style.color ?? FLEX_DEFAULTS.color;
+    const ratio = style.ratio ?? FLEX_DEFAULTS.ratio;
+
+    const mainItems = v2Content?.main_items ?? [];
+    const actionItems = v2Content?.action_items ?? [];
+    const disclaimerItems = v2Content?.disclaimer_items ?? [];
+
+    const logoBlock = mainItems.find(item => item.type === 'IMAGE');
```
Braluna-pp, line 49: ### Finding Description

**[P1] Preserve ordered image blocks**

The flex renderer finds one image, filters every image out of `main_items`, and renders the selected image before the messaging container. A valid `TEXT, IMAGE, TEXT` sequence therefore becomes `IMAGE, TEXT, TEXT`, and any second image is removed completely. This breaks the ordered typed-block contract instead of rendering CPS content directly.

### Problematic Code Snippet(s)

```js
const logoBlock = mainItems.find(item => item.type === 'IMAGE');
const mainBlocks = mainItems.filter(item => item.type !== 'IMAGE');
```

### Suggested Change(s)

- **Implementation:** Preserve the ordered block stream. Extract only a contract-defined standalone logo, or render images in place and use layout classes for visual placement.
- **Test:** Assert DOM order for `TEXT, IMAGE, TEXT` and define explicit behavior for multiple images so no valid block is silently discarded.

### Relevant Links

- [Changed extraction logic](https://github.com/paypal/paypal-messaging-components/blob/77046c128c1ba990a666c830846c1d9e73bf9f0a/src/server/v2/flex.jsx#L36-L43)
- [Repository ordered-image payload examples](https://github.com/paypal/paypal-messaging-components/blob/77046c128c1ba990a666c830846c1d9e73bf9f0a/utils/devServerProxy/messages.js#L43-L80)
- [DTCRCMERC-5374](https://paypal.atlassian.net/browse/DTCRCMERC-5374)
- [Messaging API V2 contract](https://paypal.atlassian.net/wiki/spaces/PPCMP/pages/2586232315/Upstream+Messaging+API+V2+Dev+Docs)

src/server/v2/flex.jsx
```diff
@@ -0,0 +1,98 @@
...
+                        fontSource: style.text?.fontSource,
+                        ratio
+                    })
+                }}
+            />
+            {/* eslint-enable react/no-danger */}
+            <div className="pp-flex__background" />
+            <div className="pp-flex__content">
+                {logoBlock ? (
+                    <div className="pp-flex__logo-container" aria-hidden="true">
```
Braluna-pp, line 76: ### Finding Description

**[P1] Keep the logo alternative text accessible**

`mainLabel` is built only from `mainBlocks` after the image has been removed, and the separate logo container is `aria-hidden`. For the normal leading PayPal image, its `alternative_text` is therefore absent from all flex accessible output even though the text renderer exposes the same brand label. A message such as `PayPal Pay Later` is announced only as `Pay Later`.

### Problematic Code Snippet(s)

```jsx
const mainLabel = buildContentLabel(mainBlocks);

<div className="pp-flex__logo-container" aria-hidden="true">
    {renderFlexLogo(logoBlock, color)}
</div>
```

### Suggested Change(s)

- **Implementation:** Expose the logo's `alternative_text` once, either through the main label in source order or through a labeled logo container with decorative child images.
- **Test:** Assert the full accessible name for a standard leading logo and for an unknown image with custom alternative text.

### Relevant Links

- [Changed accessible output](https://github.com/paypal/paypal-messaging-components/blob/77046c128c1ba990a666c830846c1d9e73bf9f0a/src/server/v2/flex.jsx#L40-L75)
- [Text renderer logo labeling](https://github.com/paypal/paypal-messaging-components/blob/77046c128c1ba990a666c830846c1d9e73bf9f0a/src/server/v2/message.jsx#L38-L50)
- [Messaging API V2 contract](https://paypal.atlassian.net/wiki/spaces/PPCMP/pages/2586232315/Upstream+Messaging+API+V2+Dev+Docs)
- [DTCRCMERC-5374](https://paypal.atlassian.net/browse/DTCRCMERC-5374)

Braluna-pp (commented):
Approved intent-aware PR review findings: F1, F2, F3, F4.

src/server/v2/utils/renderBlock.js
```diff
@@ -0,0 +1,62 @@
...
+        case 'IMAGE':
+            if (logoPresentation) {
+                return (
+                    <span
+                        role="img"
+                        aria-label={item.alternative_text || 'PayPal'}
+                        className={`logo inline wordmark ${getLogoBrandClass({
+                            logoName: item.name,
+                            alternativeText: item.alternative_text
+                        })}`.trim()}
```
Braluna-pp: Nit: could we make the optional brand class explicit instead of relying on `.trim()` to remove the trailing space? Pulling it into a variable and joining the truthy classes makes this easier to scan:

```js
const brandClass = getLogoBrandClass({
    logoName: item.name,
    alternativeText: item.alternative_text
});
const className = ['logo', 'inline', 'wordmark', brandClass].filter(Boolean).join(' ');
```

tests/functional/spec/createBannerTest.js
```diff
@@ -48,32 +48,51 @@ const getTestNameParts = (locale, { account, amount, style: { layout, ...style }
...
                                 height: iframeBody.clientHeight,
                                 width: iframeBody.clientWidth
                             }
                         );
                     }
-                    return banner?.clientHeight && { height: banner.clientHeight, width: banner.clientWidth };
+
+                    const banner = iframeBody.querySelector(bannerSelectors.container);
+                    return (
+                        banner?.clientHeight > 0 && {
```
Braluna-pp: Could we make this readiness check include the things that affect the screenshot instead of relying on the fixed 500 ms sleep below? A positive size proves the body exists, but fonts or images can still finish afterward and change the layout. Waiting inside the iframe for `document.fonts.ready`, incomplete images, and stable dimensions across two animation frames would remove the remaining timing guess.

src/server/v2/utils/renderBlock.js
surrsurus: Need to push back on this one a bit - CPS/UCG already merges `TEXT_VARIABLE` → `TEXT` unless the client opts into `TEXT_VARIABLES_*`. CPNW does not request those features, and adaptCpsPresentmentResponse rejects `TEXT_VARIABLE` entirely. The cited renderer switch is PMCs deferred-v6 soft-fallback, not a CPNW contract gap. Can go into more detail if you'd like more clarification on this.

src/server/v2/flex.jsx
surrsurus: Makes sense, addressed in [c66135d](https://github.com/paypal/paypal-messaging-components/pull/1367/commits/c66135df66d992dc2d71aa056c76c5ceea934609)

src/server/v2/flex.jsx
surrsurus: [72e7634](https://github.com/paypal/paypal-messaging-components/pull/1367/commits/72e763459b6c41dc4322513b69cd8b86283cddc9) Fixed here, nice catch

src/server/v2/flex.jsx
surrsurus: The find/filter behavior is intentional flex presentation, not a contract violation. Flex (like v5 and text left/top/right) extracts the brand logo into a dedicated logo container and renders the remaining blocks as messaging. Order-preserving in-stream `IMAGE` rendering is the inline path (buildLogoConfiguration + text layout), which has an explicit `TEXT`, `IMAGE`, `TEXT` test. Flex has no logo.position option, and CPNW does not request `INLINE_LOGO` for flex - so the mid-stream `TEXT`, `IMAGE`, `TEXT` shape the finding assumes is not the flex content contract.

CPS samples for this path use a single logo `IMAGE`. Dropping additional `IMAGE` blocks is theoretical vs real flex traffic, not evidence that flex must render an ordered multi-image stream.

src/server/v2/utils/renderBlock.js
surrsurus: Agreed, this I believe might have been some copypasta - fixed it here [73cac51](https://github.com/paypal/paypal-messaging-components/pull/1367/commits/73cac513356fdb01972355a9d21dab2d55ccee68) but then noticed we use this pattern in multiple locations and addressed it [32d5840](https://github.com/paypal/paypal-messaging-components/pull/1367/commits/32d58403202edef4e0551203e3c4589660dee69d)

tests/functional/spec/createBannerTest.js
surrsurus: [5dfcb44](https://github.com/paypal/paypal-messaging-components/pull/1367/commits/5dfcb44f596eb934f3d0af55aac9e95cebdbfb81) - will probably need to iterate on this but got rid of the timeout

Braluna-pp (approved):
Thanks for addressing the feedback, after ci passes you should be good to go

PR #1375, opened 2026-08-03
fix: fall back to server-resolved account for FPTI on credential-less landers

abhishekpaypal:
## Description

Credential-less lander modals (e.g. Pay Later Hub, and the upcoming TikTok Pay Later Hub) are loaded with a URL that has none of `merchant_id`, `client_id`, or `payer_id` — just an `offer`/`channel` pair. `getAccount()` in `zoid-polyfill.js` resolves the FPTI tracking "account" field exclusively from those three URL params; when all three are absent, the resolved account is empty, and the transport silently skips sending the tracking event entirely (no error, just zero requests). This means these lander flows produce **no FPTI tracking data today**.

This PR adds a fallback: `getAccount()` now also accepts a server-resolved `accountIdOverride` (delivered via the existing `serverData`/`meta` object already passed into the modal) and uses it when none of the URL-derived ids are present. This is purely additive — behavior is unchanged for any existing integration that already sends `merchant_id`/`client_id`/`payer_id`.

Note: this fix is inert on its own until paired with a small server-side change in the internal presentment service that actually populates `meta.accountIdOverride` (that service already resolves a usable fallback account for these channels for other purposes — it just isn't surfaced to the client yet). Landing this PR independently is safe either way since the new fallback parameter defaults to `undefined`, matching current behavior exactly.

## Screenshots / Videos

Working Tiktok PLH
rendered-paylater-hub

## Testing instructions

- Unit tests added in `zoid-polyfill.test.js` cover the new fallback path directly (falls back to `accountIdOverride` when `merchantId`/`clientId`/`payerId` are all absent; leaves `account` undefined when none are present, matching current behavior).
- End-to-end verification requires the paired server-side change to be deployed alongside this: load a credential-less lander URL with no `merchant_id`/`payer_id`/`client_id`, and confirm a tracking event now fires (previously zero events fired in this case).

### DIFF
--- a/src/components/modal/v2/lib/zoid-polyfill.js
+++ b/src/components/modal/v2/lib/zoid-polyfill.js
@@ -51,7 +51,7 @@ export function handleBrowserEvents(clientOrigin, propListeners, event) {
     sendEvent(createPostMessengerEvent('ack', id), clientOrigin);
 }
 
-const getAccount = (merchantId, clientId, payerId) => {
+const getAccount = (merchantId, clientId, payerId, accountIdOverride) => {
     if (merchantId) {
         return merchantId;
     }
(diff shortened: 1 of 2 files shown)


smoyalpaypal (approved):

perco12 (approved):

perco12:
pleaes remove the screenshots that show the internal urls, no internal links as PMC is a public repo

PR #1388, opened 2026-08-20
fix: replace modal Long term offer reversal with sorting

mandevi:
## Description

1. Replaced the .reverse() hack for ES/IT/CA with an explicit per-country sort direction (asc/desc) based on total_payments.
2. Changed the APR disclaimer lookup from array-index-based to keyed by total_payments, so it stays correct regardless of sort order.
3. Updated the few places reading that disclaimer object as an array to read it as a keyed object instead.

## Screenshots / Videos

US:
US

ES:
ES

IT:
IT

DE:
DE
DE2

AT:
at
AT2

FR:
FR

## Testing instructions

Can use the below URLs with different params.

V5 sdk:
https://young-anchorage-87182.herokuapp.com/pilot/stage/?account=FMN9H9D82WEF2&amount=1000&buyerCountry=IT&countryCode=IT&currency=EUR&language=it-IT&testEnv=te-remove-lt-offers-reversal&stageTag=stage_lt_offer_sorting

V6 sdk:
https://www.te-remove-lt-offers-reversal.qa.paypal.com/credit-presentment/lander/modal?amount=1000.00&buyer_country=IT&integration_type=API&language=it-IT&offer=PAY_LATER_LONG_TERM&payer_id=FMN9H9D82WEF2&stageTag=stage_lt_offer_sorting

V1 api:
https://www.te-remove-lt-offers-reversal.qa.paypal.com/credit-presentment/lander/modal?client_id=B_AreilU_5H-xdW8ESI3eQjJAamc_WvO4bFJEGMGWNO3JhCjgS3zmRFtrSCTI0G4PhctXB-SokHp9_SQzc&amount=1400&currency=EUR&buyer_country=IT&language=it-IT&offer=PAY_LATER_LONG_TERM&integrationType=API&channel=UPSTREAM&disableSetCookie=true&stageTag=stage_lt_offer_sorting

V2 api:
https://www.te-remove-lt-offers-reversal.qa.paypal.com/credit-presentment/lander/modal?amount=1200.00&buyer_country=IT&integration_type=API&language=it-IT&offer=PAY_LATER_LONG_TERM&payer_id=FMN9H9D82WEF2&stageTag=stage_lt_offer_sorting

Pay later hub native IOS:
https://www.te-remove-lt-offers-reversal.qa.paypal.com/credit-presentment/lander/modal?client_id=B_A8S1coUwMulelFvFJBQaYbqTiKiP9FxajK2nCJ9OzCFAGwaIE8aMfI84bqGvFTC_AzsCEsA0yBrGGIBo&merchant_id=W9WK2RH9RWWQC&amount=1500.00&buyer_country=AT&channel=PAY_LATER_HUB_NATIVE_IOS&stageTag=stage_lt_offer_sorting

Pay later hub native Android:
https://www.te-remove-lt-offers-reversal.qa.paypal.com/credit-presentment/lander/modal?client_id=B_A8S1coUwMulelFvFJBQaYbqTiKiP9FxajK2nCJ9OzCFAGwaIE8aMfI84bqGvFTC_AzsCEsA0yBrGGIBo&merchant_id=W9WK2RH9RWWQC&amount=1500.00&buyer_country=AT&channel=PAY_LATER_HUB_NATIVE_ANDROID&stageTag=stage_lt_offer_sorting


lyerabati16 (approved):
LGTM

papallavi (approved):

src/components/modal/v2/parts/TermsTable.jsx
```diff
@@ -36,19 +36,23 @@ const TermsTable = ({
         );
     }
 
-    // ES, IT, and CA countries offer accordion should display in ascending order (6, 12, 24 months)
-    const processedOffers =
-        offerCountry === 'ES' || offerCountry === 'IT' || offerCountry === 'CA' ? [...offers].reverse() : offers;
+    // US, ES, IT, and CA display offers ascending (6, 12, 24 months); AT, DE, and FR display descending.
+    const offerSortDirection = { US: 'asc', ES: 'asc', IT: 'asc', CA: 'asc', AT: 'desc', DE: 'desc', FR: 'desc' };
```
srinibuilds: from the earlier comment it says - "ES, IT, and CA  are in ascending order" and why are we marking US in ascending order? we should retain the current behaviour, to be inline with what is there on Production.

src/components/modal/v2/parts/TermsTable.jsx
```diff
@@ -36,19 +36,23 @@ const TermsTable = ({
...
+    const direction = offerSortDirection[offerCountry];
+    const processedOffers = direction
+        ? [...offers].sort((a, b) =>
+              direction === 'asc'
```
srinibuilds: have an explicit 'desc' check and try having a warning on unrecognised values.

src/components/modal/v2/parts/TermsTable.jsx
mandevi: The intended order for US is ascending. Also, since MORS sends US long term offers in ascending, they have not performed any reversal earlier.

src/components/modal/v2/parts/TermsTable.jsx
mandevi: I have added an explicit desc check and have also added a warning on unrecognised value/direction.

src/components/modal/v2/parts/TermsTable.jsx
```diff
@@ -36,19 +36,23 @@ const TermsTable = ({
...
+                  ? a.meta.total_payments - b.meta.total_payments
```
srinibuilds: non-numeric/malformed input produces NaN. should be validated.

mandevi: Added NaN-safe validation for total_payments before sorting offers. Non-numeric values now warn once and are left unsorted instead of silently corrupting the sort order.

src/components/modal/v2/parts/TermsTable.jsx
```diff
@@ -36,19 +36,23 @@ const TermsTable = ({
...
     const qualifyingOffers = processedOffers
         .filter(offer => offer.meta.qualifying === 'true')
         .map((offer, idx) => {
             // DE, ES, and IT use the accordion style for presentation of offers in the modal.
             if (offerAccordionCountries.includes(offerCountry)) {
-                const disclaimer =
-                    aprDisclaimer.length < processedOffers.length
-                        ? aprDisclaimer[aprDisclaimer.length - 1].aprDisclaimer
-                        : aprDisclaimer[idx].aprDisclaimer;
+                const disclaimer = (aprDisclaimer[offer.meta.total_payments] ?? aprDisclaimer.default).aprDisclaimer;
```
srinibuilds: aprDisclaimer.default is valid for zero offers case only right, so for any reason, will it resolve to undefined & may not display disclaimer text?

https://github.com/paypal/paypal-messaging-components/pull/1388/changes/da83f735f788372b6b5efb21fa7ba6f7e3589595#diff-6b9d0581d35cedac10276773567c46dc9919a7131224ad4e2631bd2826c52bcdR38

abhishekpaypal:
Can we have the branch name in semantic fashion - `bug/replace-LT-offer-reversal-with-sorting`

src/components/modal/v2/parts/TermsTable.jsx
```diff
@@ -36,19 +39,39 @@ const TermsTable = ({
...
+              .sort((a, b) =>
+                  Number.isNaN(a.totalPayments) || Number.isNaN(b.totalPayments)
+                      ? 0
+                      : sortSign * (a.totalPayments - b.totalPayments)
+              )
+              .map(({ offer }) => offer)
+        : offers;
 
     const qualifyingOffers = processedOffers
         .filter(offer => offer.meta.qualifying === 'true')
```
abhishekpaypal: Any specific reason we are sorting first and then filtering out the offers which are missing on `meta.total_payments` ?
Ideally, an optimized approach would be to filtering and then sort, to rule out unnecessary entries as a pre-check.

src/components/modal/v2/parts/views/LongTerm/Content.jsx
```diff
@@ -29,44 +29,46 @@ const getAPRDetails = ({ offers, genericDisclaimer, disclaimer: { zeroAPR, mixed
...
             }
-        ];
+        };
     }
 
     // TODO: Clean up backwards compatible code after release and content updates.
-    return qualifyingOffers.map(({ content: { disclaimer } }) => {
+    // Keyed by each offer's own term (total_payments) so the disclaimer stays paired with its offer
+    // regardless of what order `offers` is rendered/sorted in downstream.
+    return qualifyingOffers.reduce((acc, { meta, content: { disclaimer } }) => {
```
abhishekpaypal: Can we add explicit UT to cover the explicit usecase of mapping right APRDetails to its offer - **given offers in scrambled order, each rendered offer still gets its own correct disclaimer**

abhishekpaypal (changes requested):

src/components/modal/v2/parts/TermsTable.jsx
```diff
@@ -36,19 +39,39 @@ const TermsTable = ({
...
         .map((offer, idx) => {
             // DE, ES, and IT use the accordion style for presentation of offers in the modal.
             if (offerAccordionCountries.includes(offerCountry)) {
                 const disclaimer =
```
jeremy-herman: Could we add one direct regression assertion in a test that expands reordered ES/IT terms and verifies each term’s APR/TIN disclaimer? The current snapshots are green, but the existing disclaimer helper only asserts the offer headline. Since the live defect is specifically the disclaimer-to-term association, a semantic assertion would protect that mapping more precisely.

tests/functional/v2/snapshots/modal/api/AT/DEV_AT_LONG_TERM/mobile/99-Offer_accordion_show_correct_payment_breakdown_information_for_amount-mobile-snap.png
jeremy-herman, line 1: It looks like many of these snapshots were incorrect before and they didn't reflect the same ordering we saw in production for these offers? It looks like the deterministic sorting fixed these. Thank you for updating them!

src/components/modal/v2/parts/TermsTable.jsx
```diff
@@ -5,6 +5,9 @@ import OfferCard from './OfferCard';
 import OfferAccordion from './OfferAccordion';
 import LoadingShimmer from './LoadingShimmer';
 
+// US, ES, IT, and CA display offers ascending (6, 12, 24 months); AT, DE, and FR display descending.
+const offerSortDirection = { US: 'asc', ES: 'asc', IT: 'asc', CA: 'asc', AT: 'desc', DE: 'desc', FR: 'desc' };
```
jeremy-herman: I like the self-documenting nature of the `asc` / `desc` strings, but maybe they should be in constants?

I'd even go a step further and suggest we don't need string here. Would something like this work? An incline comment could document the intent well and this would likely simplify the need to derive the sort direction below. 
```javascript
const OFFER_SORT_MULTIPLIER_BY_COUNTRY = {
    US: 1,
    ES: 1,
    IT: 1,
    CA: 1,
    AT: -1,
    DE: -1,
    FR: -1
};

const multiplier = OFFER_SORT_MULTIPLIER_BY_COUNTRY[offerCountry];

const processedOffers = multiplier
    ? [...offers].sort(
          (a, b) => multiplier * (Number(a.meta.total_payments) - Number(b.meta.total_payments))
      )
    : offers;
```

jeremy-herman (commented):
Some minor comments, but overall looks good!

src/components/modal/v2/parts/TermsTable.jsx
jeremy-herman: If we wanted the code to self-document a constant like this would work too! Offers the most clarity and a simplified algorithm. 
```javascript
const SORT_DIRECTION = {
    ASCENDING: 1,
    DESCENDING: -1,
};

const OFFER_SORT_DIRECTION_BY_COUNTRY = {
    US: SORT_DIRECTION.ASCENDING,
    ES: SORT_DIRECTION.ASCENDING,
    IT: SORT_DIRECTION.ASCENDING,
    CA: SORT_DIRECTION.ASCENDING,
    AT: SORT_DIRECTION.DESCENDING,
    DE: SORT_DIRECTION.DESCENDING,
    FR: SORT_DIRECTION.DESCENDING
};
```

src/components/modal/v2/parts/TermsTable.jsx
jeremy-herman: Looks like you addressed this here while I was reviewing: [379cffd](https://github.com/paypal/paypal-messaging-components/pull/1388/commits/379cffd14cbda19d53a39823183825b29ba987d3) 👍

src/components/modal/v2/parts/TermsTable.jsx
mandevi: fixed by filtering to qualifying offers first, then sorting only that subset.

src/components/modal/v2/parts/views/LongTerm/Content.jsx
mandevi: Added an UT.

src/components/modal/v2/parts/TermsTable.jsx
mandevi: Thanks Jeremy!!!

I switched to numeric SORT_DIRECTION constants for more clarity. 
Also kept a validation guard: if a country's mapped value isn't exactly one of the two known constants , it warns and falls back to unsorted rather than silently producing NaN comparisons.

jeremy-herman (approved):

srinibuilds (approved):

PR #1390, opened 2026-08-24
fix: improve calculator field accessibility

ryanmagoon:
## Description

This keeps the purchase amount field's accessible name aligned with the text shown on screen. Empty US calculators use "Enter amount"; populated calculators use "Purchase amount." Invalid amounts point to one error message. Service failures stay announced without incorrectly marking a valid amount as invalid.

## Screenshots / Videos

N/A — semantics-only change with no intended visual impact.

## Testing instructions

Regression coverage verifies empty, populated, zero, cleared, and invalid amounts, plus the service-failure path.

- `npm run lint`
- `npm test -- --runInBand` — 41 suites, 511 tests passed
- `npm run build:production`
- Targeted Chromium modal/axe path — 1 test passed

## Review

Please start with the accessible-name and error-state assertions in `Calculator.test.js`.


src/components/modal/v2/parts/Calculator.jsx
```diff
@@ -102,6 +103,9 @@ const Calculator = ({
 
     // Pass view, isLoading state, and calculator props into getError to get the appropriate error, if any. Could return as 'null'.
     const error = getError(view, isLoading, calculator, delocalize(displayValue ?? '0', country, language), country);
+    const hasInputError = Boolean(
+        error && error !== calculator.genericError && (hasInitialAmount || hasUsedInputField)
+    );
```
jeremy-herman, line 108: I think this conflicts with the PR's “semantics-only change with no intended visual impact” scope. The expected behavior is that zero or empty values may show error text, but only a non-zero invalid amount gets the red field styling and warning icon.

Could we separate the semantic and visual predicates? For example:

```suggestion
const hasInputError = Boolean(
    error && error !== calculator.genericError && (hasInitialAmount || hasUsedInputField)
);
const showVisualInputError = Boolean(
    error && error !== calculator.genericError && hasEnteredAmount
);
```

Then `hasInputError` could control `aria-invalid`, `aria-describedby`, and the error-message ID, while `showVisualInputError` controls the warning icon and `input__wrapper--error` class. This should add the intended accessibility semantics without changing the existing zero/empty visual state.

src/components/modal/v2/parts/Calculator.jsx
```diff
@@ -215,17 +219,20 @@ const Calculator = ({
...
-                    } ${country || ''} ${error && hasEnteredAmount ? 'input__wrapper--error' : ''}`}
+                    } ${country || ''} ${hasInputError ? 'input__wrapper--error' : ''}`}
                 >
                     <label htmlFor="purchase-amount" className={`input__label ${country}`}>
                         {renderInputLabelOnEmptyField(country)}
                     </label>
                     {inputCurrencySymbol && <div className="input__currency-symbol">{inputCurrencySymbol}</div>}
                     <input
                         id="purchase-amount"
+                        aria-label={country === 'US' && displayValue === '' ? formattedInputPlaceholder : undefined}
```
ryanmagoon, line 231: The US empty state hides the native label and only shows ‘Enter amount.’ This gives the field the same accessible name. Once an amount is entered, the native ‘Purchase amount’ label takes over.

src/components/modal/v2/parts/Calculator.jsx
ryanmagoon: Good catch. I split the accessibility state from the visual error state, so zero and cleared values no longer get the warning icon or red styling. I added regression coverage for both cases in `9f1e20b6e`.

jeremy-herman (approved):
LGTM. 

Tested the Pay Monthly modal while running locally as compared to the existing. 

Error message values are now associated with the field validation:

before | after
-- | --
Image | Image

yeul (approved):
LGTM! Feel free to squash and merge

PR #1393, opened 2026-08-24
fix: restore PayPal logo text alternative

ryanmagoon:
## Description

Restores the PayPal text alternative on legacy message logos. Composite logos expose the brand once and keep any additional image fragment decorative, so the containing button gets one native accessible name from the logo and its visible offer copy.

V2 rendering is unchanged.

## Screenshots / Videos

N/A - semantics-only change with no visual impact. The server-rendered logo test asserts `alt="PayPal"`, and the browser tests assert the resulting button names before and after the responsive copy breakpoint.

## Testing instructions

Regression coverage includes single-image and composite logos, the No Interest message above and below its rendered 141.552px breakpoint, the Long Term message, and the existing axe scan.

- `npm run lint`
- `npm test -- --runInBand`
- `npx playwright test tests/playwright/tests/sdk/messages/multi.spec.js`
- `npm run build:production`

## Review

Please start with `Logo.jsx`, the new focused logo test, and the two browser accessible-name assertions.


src/server/message/parts/Logo.jsx
```diff
@@ -7,10 +7,11 @@ const Logos = ({ mutations }) => {
...
 
     return (
-        <div className="message__logo-container" aria-hidden="true">
-            {logos.map(({ src, dimensions: [width, height] }) => (
+        <div className="message__logo-container">
+            {/* Multiple image fragments render one logo, so only one contributes its text alternative. */}
+            {logos.map(({ src, dimensions: [width, height] }, index) => (
                 <div className="message__logo message__logo--svg">
-                    <img src={src} alt="" role="presentation" />
+                    <img src={src} alt={index === 0 ? 'PayPal' : ''} />
```
jeremy-herman: I think we can maintain that enhancement from the other solution where we conditionally show `PayPal Credit` for the screen reader. Below are some suggestions, let me know what you think.

The renderer derives messageType from markup metadata at `src/server/message/index.jsx:47` and uses it to select logo mutations at `src/server/message/index.jsx:60`. We can derive the semantic brand at that point, with "PayPal" as default.

```jsx
// src/server/message/index.jsx:4
import {
    objectMerge,
    objectFlattenToArray,
    curry,
    getStandardProductOffer,
    OFFER
} from '../../utils/server';

// Replace src/server/message/index.jsx:87-88
const logoType = style.logo?.type;
const standardOffer = getStandardProductOffer(messageType);
const brandName =
    standardOffer === OFFER.PAYPAL_CREDIT_NO_INTEREST ? "PayPal Credit" : "PayPal";

const logoEl = (
    <>
        <Logo mutations={mutationRules.logo} />
        <span className="sr-only">{brandName}</span>
    </>
);
```

Then, at `src/server/message/parts/Logo.jsx:10-17`, could we remove aria-hidden from the logo container and keep only the visual asset wrappers hidden?

```jsx
return (
    <div className="message__logo-container">
        {logos.map(({ src, dimensions: [width, height] }) => (
            <div className="message__logo message__logo--svg" aria-hidden="true">
                <img src={src} alt="" role="presentation" />
                <canvas height={height} width={width} />
            </div>
        ))}
    </div>
);
```

jeremy-herman (commented):
LGTM, but would like to explore the feasibility of using the offer type to get the branding label. 

Image

src/server/message/parts/Logo.jsx
ryanmagoon: Yep, that makes sense. I moved the brand name up to the renderer so PayPal Credit messages announce “PayPal Credit” and everything else defaults to “PayPal.” The logo images are decorative again, with unit and browser coverage for both paths.

jeremy-herman (approved):
Very nice!

PPC | PLLT
-- | --
Image | Image

yeul (approved):
LGTM, thank you for taking care of this! Feel free to squash and merge

PR #1394, opened 2026-08-24
fix: make modal loading states accessible

ryanmagoon:
## Description

This makes the calculator loading state accessible across every current variant. It adds one stable status announcement, keeps completed financing results in a polite live region while updates are busy, hides visual-only placeholders from assistive technology, and raises the shimmer colors to at least 3:1 contrast. The component falls back to "Loading financing options" if content omits a label, while localized content can override it. A content contract still requires every current calculator locale and variant to provide its own loading announcement.

## Screenshots / Videos

Focused captures of the shimmer colors. The modal's separate spinner and blur are intentionally omitted so the contrast change is visible.

### Standard

| Before | After |
| --- | --- |
| Standard loading state before the contrast change | Standard loading state after the contrast change |

### Checkout

| Before | After |
| --- | --- |
| Checkout loading state before the contrast change | Checkout loading state after the contrast change |

## Testing instructions

Regression coverage verifies the English fallback and localized override, one status region, the combined live and busy results state, card and accordion placeholders, every current calculator content file, and each rendered shimmer color against its composited background. Browser coverage samples standard, checkout, German accordion, and dark-mode layouts.

- `npm run lint`
- `npm test -- --runInBand`
- `npx playwright test tests/playwright/tests/sdk/modals/multi.spec.js`
- `npm run build:production`

## Review

Please start with the loading-state boundary in `Calculator.jsx`, the shared content contract, and the rendered contrast measurements.

### DIFF
--- a/src/components/modal/v2/parts/Calculator.jsx
+++ b/src/components/modal/v2/parts/Calculator.jsx
@@ -254,7 +254,7 @@ const Calculator = ({
                 {isLoading ? loadingLabel : ''}
             </div>
             {(hasInitialAmount || hasUsedInputField) && !error ? (
-                <div className="content-column" aria-busy={isLoading ? 'true' : undefined}>
+                <div className="content-column" aria-live="polite" aria-busy={isLoading ? 'true' : undefined}>
                     <TermsTable
                         view={view}
                         isLoading={isLoading}
--- a/src/components/modal/v2/parts/Calculator.jsx
+++ b/src/components/modal/v2/parts/Calculator.jsx
@@ -84,7 +84,7 @@ const Calculator = ({
     const { amount } = useXProps();
     const { country, views } = useServerData();
     const { language } = views[0].meta;
-    const { title, genericTitle, inputLabel, inputPlaceholder, inputCurrencySymbol } = calculator;
+    const { title, genericTitle, inputLabel, inputPlaceholder, inputCurrencySymbol, loadingLabel } = calculator;
     const formattedInputPlaceholder = currencyFormat(inputPlaceholder).replace(/(\s?€)/g, '');
 
     // Set hasUsedInputField to true if someone has typed in the input field at any point.
@@ -243,8 +243,11 @@ const Calculator = ({
                 </div>
                 <div aria-live="polite">{renderError(error || emptyState || isLoading)}</div>
             </form>
+            <div role="status" className="sr-only">
+                {isLoading ? loadingLabel : ''}
+            </div>
             {(hasInitialAmount || hasUsedInputField) && !error ? (
-                <div aria-live="polite" className="content-column">
+                <div className="content-column" aria-busy={isLoading ? 'true' : undefined}>
                     <TermsTable
                         view={view}
                         isLoading={isLoading}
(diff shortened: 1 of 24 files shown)


jeremy-herman (approved):
LGTM. Contrast is much improved. The loading labels will be tested as part of a separate, follow-up content change.

src/components/modal/v2/parts/Calculator.jsx
```diff
@@ -235,8 +235,11 @@ const Calculator = ({
                 </div>
                 <div aria-live="polite">{renderError(error || emptyState || isLoading)}</div>
             </form>
+            <div role="status" className="sr-only">
+                {isLoading ? loadingLabel : ''}
```
jeremy-herman, line 254: I think we should have an english default message when the localized content isn't available for screen reader text. Can we add a default loading label here: `Loading financing options`?

jeremy-herman (changes requested):

src/components/modal/v2/parts/Calculator.jsx
```diff
@@ -235,8 +235,11 @@ const Calculator = ({
...
+            </div>
             {(hasInitialAmount || hasUsedInputField) && !error ? (
-                <div aria-live="polite" className="content-column">
+                <div className="content-column" aria-busy={isLoading ? 'true' : undefined}>
```
tsmyrniospp: This removes aria-live="polite" and adds only aria-busy. aria-busy does not make an element a live region. TermsTable, OfferCard, and OfferAccordion do not provide another live role, so when loading finishes and offers replace the shimmers, screen readers no longer announce the updated financing results. Retain the existing live-region behavior or add an equivalent completion announcement.

src/components/modal/v2/parts/Calculator.jsx
```diff
@@ -235,8 +242,11 @@ const Calculator = ({
...
```
tsmyrniospp: This removes `aria-live="polite"` and adds only `aria-busy`. Though `aria-busy` does not make an element a live region, and TermsTable, OfferCard, and OfferAccordion do not provide another live role, so when loading finishes and offers replace the shimmers, screen readers no longer announce the updated financing results. Consider either retaining the existing live-region behavior or adding an equivalent completion announcement.

tsmyrniospp (changes requested):
Apple Wallet (content/modals/US/PL2GO/apple_wallet_long_term.json) needs the loadingLabel specified.

```bash
npm test -- --runInBand tests/unit/spec/content/modalAccessibility.test.js
```

 Result:

```text
FAIL tests/unit/spec/content/modalAccessibility.test.js

Test Suites: 1 failed, 1 total
Tests:       1 failed, 17 passed, 18 total
 ```

The failing case is:

 ```text
content/modals/US/PL2GO/apple_wallet_long_term.json provides a loading announcement
 ```

 Failure:

 ```text
Expected: Any<String>
Received: undefined

ryanmagoon:
Fixed both. Apple Wallet now includes the loading label, and the results region keeps `aria-live="polite"` with `aria-busy` so the updated options are announced when loading finishes. All checks are green.

tsmyrniospp (approved):

jeremy-herman (approved):

